Issue #41, opened 2017-03-20
Error handling: Panics vs Result


boxtown:
Currently the responsibility for guarding against exceptional cases (e.g. input not in valid domain, mathematically invalid operations etc) is passed to the user. We `panic` when an operation does not make mathematical sense (e.g. calculating the cumulative distribution function for discrete distributions at a negative input) which forces users to double check to make sure their inputs are valid. While this results in technically correct and predictable behavior from the API, I'm not sure if it's ergonomic or idiomatic and have been mulling over possibly introducing a `Result` based API either replacing or in addition to the stricter `panic` based API. This however warrants some discussion and I would love feedback from the community

cbreeden:
I think consistency is huge here.  I did a cursory glance at other crates and noticed the following:

## Quick Survey
### STD
- `f64`/`f32` operations: No result or panics. `NaN`, `Inf`, and `-Inf` are used appropriately.
- Arithmetic operations: `panic!` in debug on overflow.
- `Result` is used for parsing.
- `Option` is used for `TryFrom` conversion of numeric types (which can fail, otherwise `From`).

### Rand crate
- Distributions `panic!` for invalid input parameters (negative std, min > max, etc).

### Num/Num-Traits
- Ratios with zero denominator `panic!`s
- `Option` is used for fallible type conversions (ie, `Float` -> `BigInt` if `Float` is not finite)
- `FromPrimitive` and `ToPrimitive` traits use `Option`

It seems to me that a takeaway could be:
- `Option` should be used for fallible conversions.
- `Result` should be used for parsing from binary or utf8.
- `panic!` for invalid inputs for mathematical operations, including with the initialization of such objects (like `Ratio` or distributions).
- Use `NaN`, `Inf` and `-Inf` for floats as appropriate.
- Provide "checked" variations when possible.  These methods use an `Option`.  (I suppose under the assumption that only one thing could go wrong; or it's usually assumed that the exact reason for failure isn't that important).

boxtown:
[...]

PR #56, opened 2017-06-02
Keep calm and carry on

Migi:
See issue #41.

Don't panic when calculating a pdf, cdf or pmf whenever the input is a value that the distribution cannot attain. The pdf, pmf and cdf are still defined at those inputs and we should return the correct values.

This commit also adds a few tests to each distribution to test that pdf(-∞)=0, pdf(∞)=0, cdf(-∞)=0 and cdf(∞)=1. It also uses simple numerical integration to test for continuous distributions that the integral of the pdf is approximately equal to the cdf, and for discrete distributions that the sum of the pmf is equal to the cdf.


src/distribution/binomial.rs
```diff
@@ -293,18 +289,15 @@ impl Discrete<u64, f64> for Binomial {
...
-    ///
     /// # Formula
     ///
     /// ```ignore
     /// (n choose k) * p^k * (1 - p)^(n - k)
     /// ```
     fn pmf(&self, x: u64) -> f64 {
-        assert!(x <= self.n,
-                format!("{}", StatsError::ArgLte("x", 1.0)));
+		if x > self.n {
```
boxtown: can we put this into the `match` statement? same with the `ln_pmf` method below

src/distribution/chi.rs
```diff
@@ -306,36 +297,30 @@ impl Continuous<f64, f64> for Chi {
     ///
     /// where `k` is the degrees of freedom and `Γ` is the gamma function
     fn pdf(&self, x: f64) -> f64 {
-        assert!(x >= 0.0, format!("{}", StatsError::ArgNotNegative("x")));
-        if self.freedom == f64::INFINITY || x == f64::INFINITY || x == 0.0 {
+        if self.freedom == f64::INFINITY || x == f64::INFINITY || x <= 0.0 {
             0.0
         } else if self.freedom > 160.0 {
-            self.ln_pdf(x)
+            self.ln_pdf(x).exp()
```
boxtown: good catch, could we possibly add a test case for this?

src/distribution/gamma.rs
```diff
@@ -566,14 +564,14 @@ mod test {
         test_almost(10.0, 10.0, 1.0251532120868705806216092933926141802686541811003037e-30, 1e-44, |x| x.pdf(10.0));
         test_almost(10.0, 1.0, 0.0000010137771196302974029859010421116095333052555418644397, 1e-20, |x| x.pdf(1.0));
         test_almost(10.0, 1.0, 0.12511003572113329898476497894772544708420990097708601, 1e-15, |x| x.pdf(10.0));
-        test_is_nan(10.0, f64::INFINITY, |x| x.pdf(1.0));
-        test_is_nan(10.0, f64::INFINITY, |x| x.pdf(f64::INFINITY));
+        test_is_nan(10.0, f64::INFINITY, |x| x.pdf(1.0)); // is this really the behavior we want?
```
boxtown: Honestly not sure what the desired behavior is in this case. We can leave it for now and open another issue for discussion. See #57

src/distribution/binomial.rs
Migi: No, the opposite should happen: the match should go away. Matching on floating-point values [is deprecated](https://github.com/rust-lang/rust/issues/41255). It's currently only a warning but it will become a hard error in the future.

src/distribution/chi.rs
Migi: Done.

Migi:
> adding tests for the upper bounds (i.e. x == f64::INFINITY) would be greatly appreciated

These tests are already done for each distribution in `internal::test::check_continuous_distribution` and `internal::test::check_discrete_distribution`.

I also updated the changelog.

Rustfmt clearly hasn't been run in a while, because when I ran it, a lot of things were changed that I haven't touched. And some of the things rustfmt did (especially in src/function/factorial.rs and src/function/stable.rs) look very strange, so I undid them. It seems rustfmt with default arguments does some weird things within {} brackets. Perhaps it would be good to use some non-default rustfmt parameters and add those to a rustfmt.toml file to fix this.

boxtown:
Cool, looks good. I'll merge this in and open up issues for removing matches on floats and for creating a rustfmt.toml

PR #60, opened 2017-06-22
Fix a bug in the uniform continuous distribution

mp4096:
Hi and thanks for the library!

### Problem

I think there's a bug in the implementation of the continuous uniform distribution. According to the documentation, it should return values in the range `[min, max]`, i.e. `min` ≤ random value ≤ `max`.

Unfortunately, the current implementation generates values in the range `[min, max + 1)`, i.e. `min` ≤ random value < `max + 1`.

I think it might be a copy-paste bug 'inherited' from the discrete uniform distribution, where you really need to add 1 to the upper bound.

### Solution

This PR is a __partial__ fix for this bug: It changes the range of the continuous uniform distribution to `[min, max)`, i.e. `min` ≤ random value < `max`. I don't have a quick fix to include the upper bound, but I think it's important to at least fix the `+ 1.0` issue. As of `0.7.0`, you cannot really sample values between 0 and 1 without resorting to wild workarounds.

### References
* [MathNET implementation](https://github.com/mathnet/mathnet-numerics/blob/36a46bd7fcd3dc273fbd147bf6f27dabfdd0d586/src/Numerics/Distributions/ContinuousUniform.cs#L284-L287)
* [discrete uniform distribution in `statrs`](https://github.com/boxtown/statrs/blob/36a99a98880d16cbbab2c626a1ce18a4d13e84c1/src/distribution/discrete_uniform.rs#L94-L96)
* [`rand::Rng` trait](https://doc.rust-lang.org/rand/rand/trait.Rng.html#method.gen_range), which says that `gen_range()` does not include the upper bound

### DIFF
--- a/src/distribution/uniform.rs
+++ b/src/distribution/uniform.rs
@@ -466,7 +466,7 @@ mod test {
     }
 
     #[test]
-    fn test_in_range() {
+    fn test_samples_in_range() {
         use super::Distribution;
 
         use rand::{StdRng, SeedableRng};
--- a/src/distribution/uniform.rs
+++ b/src/distribution/uniform.rs
@@ -473,9 +473,9 @@ mod test {
         let seed: &[_] = &[1, 2, 3, 4, 5];
         let mut r: StdRng = SeedableRng::from_seed(seed);
 
-        let min = -4.2;
-        let max = 0.42;
-        let num_trials = 1000;
+        let min = -0.5;
+        let max = 0.5;
+        let num_trials = 10_000;
         let n = try_create(min, max);
 
         assert!((0..num_trials)
--- a/src/distribution/uniform.rs
+++ b/src/distribution/uniform.rs
@@ -464,4 +464,22 @@ mod test {
         test::check_continuous_distribution(&try_create(0.0, 10.0), 0.0, 10.0);
         test::check_continuous_distribution(&try_create(-2.0, 15.0), -2.0, 15.0);
     }
+
+    #[test]
+    fn test_in_range() {
+        use super::Distribution;
+
+        use rand::StdRng;
+        let mut r = StdRng::new().unwrap();
+
+        let min = -4.2;
+        let max = 0.42;
+        let num_trials = 1000;
+        let n = try_create(min, max);
+
+        assert!((0..num_trials)
+            .map(|_| n.sample::<StdRng>(&mut r))
+            .all(|v| (min <= v) && (v < max))
+        );
+    }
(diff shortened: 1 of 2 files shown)


vks:
Could you please add a test verifying it works now?

mp4096:
Sure, but I'm not sure how can you test this in a deterministic way. However, I can just sample a lot of values and see whether they are all in the range `[min, max)`.

vks:
Yes, I was thinking of something like this:

```diff
diff --git a/src/distribution/uniform.rs b/src/distribution/uniform.rs
index 11b03da..b839849 100644
--- a/src/distribution/uniform.rs
+++ b/src/distribution/uniform.rs
@@ -464,4 +464,17 @@ mod test {
         test::check_continuous_distribution(&try_create(0.0, 10.0), 0.0, 10.0);
         test::check_continuous_distribution(&try_create(-2.0, 15.0), -2.0, 15.0);
     }
+
+    #[test]
+    fn test_sample() {
+        use super::Sample;
+
+        let mut u = Uniform::new(0., 1.).unwrap();
+        let mut r = ::rand::StdRng::new().unwrap();
+        for _ in 0..100000 {
+            let x = u.sample(&mut r);
+            assert!(0. <= x);
+            assert!(x < 1.);
+        }
+    }
 }
```

vks:
If you want it to be deterministic, you can seed the RNG.

mp4096:
Wow, we've written the tests within a minute :hand: :sunglasses:

mp4096:
Seeding the PRNG is surely a good idea! I'll take a look at it tomorrow.

src/distribution/uniform.rs
```diff
@@ -464,4 +464,23 @@ mod test {
         test::check_continuous_distribution(&try_create(0.0, 10.0), 0.0, 10.0);
         test::check_continuous_distribution(&try_create(-2.0, 15.0), -2.0, 15.0);
     }
+
+    #[test]
+    fn test_in_range() {
```
boxtown: If we could re-name this to something like `test_sample_in_range` to be a little more explicit that would be awesome

boxtown:
This is an excellent catch. If you could address the comment I'll merge asap.

src/distribution/uniform.rs
mp4096: Yep, fixed it!

mp4096:
I struggled a bit with the wording for changelog, but I hope one can understand what I mean.

Issue #58, opened 2017-06-02
Create rustfmt.toml


boxtown:
Need a rustfmt.toml to enforce consistent style guidelines

PR #67, opened 2017-09-24
Create rustfmt.toml #58

sminez:
This addresses #58,  adding a `rustfmt.toml` file and applying the config to the current code-base.

The default `rustfmt` config was erroring for me when I ran it due to line length an and other couple of issues so the config supplied is the one I have been using. I've included a note in the commit that this commit will need to be ignored when going through git blame in order to find the source of changes prior to the reformatting.


rustfmt.toml
```diff
@@ -0,0 +1,32 @@
+# Run using `rustfmt --write-mode overwrite *.rs` in the
+# root of the src directory. You may still get some
+# formatting errors (whitespace etc) which should be
+# fixed manually before committing.
+
+# Max line length that will be parsed by rustfmt
+max_width = 160
```
boxtown: Can we re-run with this set to 120? I think 160 is a little much

boxtown:
This looks great, thanks for doing this. Can you address the comment about max line length and then I can merge?

rustfmt.toml
sminez: Not sure if the comment is confusing here...160 is the max line length that will be parsed by the `rustfmt` tool _not_ the max line length allowed in the code base. (I hope that makes sense!)

So, under current settings: any line up to 160 characters will be parsed and formatted, anything longer will be an error. 160 is overkill now but rustfmt was unable to parse some of the codebase without going this high. I'll bring it down to 120 and then from now on, when you run rustfmt it will complain at you if you go over 120 characters (which you will need to manually fix each time).

sminez:
Of course! I'm just getting my head round the codebase but I'll see if there is anywhere more productive that I can help out :)

rustfmt.toml
sminez: looks like its mainly test code that `#[cfg_attr(rustfmt, rustfmt_skip)]` doesn't seem to be working for...

boxtown: ah I think I misunderstood the purpose of max_width. It should be fine to set to 160 in that case. Sorry about the mix up

PR #68, opened 2017-10-06
Implement Pareto distribution

### LINKED ISSUE #64: Implement Pareto distribution
References:

https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics/Distributions/Pareto.cs
https://en.wikipedia.org/wiki/Pareto_distribution

mp4096:
Closes #64 

---

Hi! I implemented Pareto distribution as in the Math.NET library with a few modifications:

* Sampling is done by inverse cdf transform from uniformly distributed numbers in `(0, 1]`, see [Wikipedia](https://en.wikipedia.org/wiki/Pareto_distribution#Random_sample_generation)
* Cdf returns 0 for values less than scale
* Mean returns infinity for shape <= 1
* Entropy uses difference of logs instead of log of a quotient
* Skewness panics if shape <= 3


src/distribution/pareto.rs
```diff
@@ -0,0 +1,539 @@
...
+    ///
+    /// # Formula
+    ///
+    /// ```ignore
+    /// x_m*2^(1/α)
+    /// ```
+    ///
+    /// where `x_m` is the scale and `α` is the shape
+    fn median(&self) -> f64 {
+        self.scale * (2.0_f64.powf(1.0 / self.shape))
```
boxtown: I think you'd be fine with just `2f64` here

boxtown:
This is an excellent PR. I appreciate the hard work! I just had a single comment, if you could address that I'll happily merge

src/distribution/pareto.rs
mp4096: Sure, but I kind of prefer to be extra-explicit about this being a float literal. It's also consistent then, e.g. when you delete `_f64`, you don't need to add `.0`.

Anyway it's a subjective code style issue, like trailing commas in lists. If you insist on using `2f64`, I'll be glad to change this line as well as ones down below (lines 506-515).

mp4096:
Thank you too for rapid feedback!

I would also appreciate if you squash on merge, I think I made too many dirty-wip-tests implementing-commits.

src/distribution/pareto.rs
boxtown: While I agree with you on the virtues of explicitness, I believe one would be hard pressed to mistake `f64` for anything other than a float. I honestly don't really feel *that* strongly one way or the other but I would like it to be consistent with the rest of the code base which I believe uses `{n}f64`

mp4096: Sure, will do!

mp4096: Done in 25d501245121570efa4e3f01ea9723b3de5bb044

PR #69, opened 2017-10-09
Remove own exp(x) - 1 implementation

mp4096:
Hello again! :grinning:

This PR removes the in-house implementation of the exponent minus one function and uses the Rust [std version](https://doc.rust-lang.org/std/primitive.f64.html#method.exp_m1) instead.

I tested the std implementation first in ba4c859d3089107b15436abf336868503599a58a and it had shown only one very minor [numerical regression](https://github.com/boxtown/statrs/commit/ba4c859d3089107b15436abf336868503599a58a#diff-8b00a0cead809d74f3b4906cadb831bfR27).

---

Pros:
* less code to maintain
* fewer modules (and tests) ⇒ hopefully faster compilation time

Cons:
* nothing I'm aware of


boxtown:
Excellent work again! Really appreciate the help 👍

PR #70, opened 2017-10-10
Add entropy of a categorical distribution

### LINKED ISSUE #49: Entropy for categorical distribution
Math.NET has an implementation but no unit tests, and I couldn't find a source for the calculation from a cursory search. If anyone can provide a source or derivation for either a closed form or numerical solution (or prove the calculation for https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics/Distributions/Categorical.cs#L274) then we can move forward with implementation

mp4096:
Hello again! :grinning: 

So, about #49: The entropy formula for the categorical distribution is actually the _definition_ of the information entropy, see e.g. [Wikipedia](https://en.wikipedia.org/wiki/Entropy_(information_theory)#Definition).
If all categories have the same probability, the formula collapses to `ln(n)`, which is exactly the entropy of a discrete uniform distribution.

The only gotcha is to handle the cases when a category has zero probability. The naive implementation (used by Math.NET) would throw a `NaN` when multiplying `0.0` by `Inf`. So I added a `filter()` which maps `p * ln(p)` only on non-zero values. This case is also added to the unit test.

---

Close #49

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ HEAD
  - implemented infinite sequence generator for sawtooth sequence
  - deprecate old non-infinite iterators in favor of new infinite iterators with `take`
  - Implemented `Pareto` distribution
+ - Implemented `Entropy` trait for the `Categorical` distribution
 
 v0.8.0
  - `cdf(x)`, `pdf(x)` and `pmf(x)` now return the correct value instead of panicking when `x` is outside the range of values that the distribution can attain.
--- a/src/distribution/categorical.rs
+++ b/src/distribution/categorical.rs
@@ -267,9 +267,10 @@ impl Entropy<f64> for Categorical {
     /// # Formula
     ///
     /// ```ignore
-    /// - Σ_j ( p_j * ln(p_j) )
+    /// - sum(p_j * ln(p_j)) for j in 0..k-1
     /// ```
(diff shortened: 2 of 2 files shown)


mp4096:
Uhm, I've noticed one more things: The summation ranges in the docstrings were not correct, since [Rust's range notation](https://doc.rust-lang.org/std/ops/struct.Range.html) `a..b` means from `a` inclusive to `b` _exclusive_. So basically the summation was done till `k - 2`.

I've changed the notation to the strictly matematical one; however, maybe we should go for [the one used in the Dirichlet distribution](https://github.com/boxtown/statrs/blob/2a42bbb60bf45ea874e48ab6323467e726d5cea8/src/distribution/dirichlet.rs#L231)?

boxtown:
So the notation within the Formula blocks in the doc strings are to represent Mathematical formulas rather than actual Rust code, but I do agree that standardizing with the notation found in the `Dirichlet` file would be good. If you could make that change I'd happily approve this PR. Again, I appreciate the hard work!

mp4096:
Sure, I changed the docstring formula style to match `dirichlet.rs`.

I also appreciate your rapid feedback (and merging :smiley:)! It's always a great pleasure to contribute to a well-maintained project.

PR #76, opened 2017-10-17
Fix panic description for the mode of the Chi distribution

mp4096:
* Fix a typo in the panic description (compare it with the formula)
* Add a unit test for this panic

### DIFF
--- a/src/distribution/chi.rs
+++ b/src/distribution/chi.rs
@@ -447,7 +447,7 @@ mod test {
 
     #[test]
     #[should_panic]
-    fn test_mode_panic() {
+    fn test_mode_freedom_lt_one() {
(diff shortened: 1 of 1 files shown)


src/distribution/chi.rs
```diff
@@ -446,6 +446,12 @@ mod test {
     }
 
     #[test]
+    #[should_panic]
+    fn test_mode_panic() {
```
boxtown: The convention I've been following is to name these panic cases for why they're panicking, e.g. `test_mode_freedom_lt_one`. If we could follow that, that would be great

mp4096: Good to know this, thanks!

mp4096: Done!

PR #77, opened 2017-10-19
Refactor chi-squared distribution

mp4096:
Ok, this is a larger one:

### Mode

Fix formula for the mode. Math.NET implementation has an error for `freedom < 2` and faulty unit tests as well.

### Pdf and log density

When investigating this unit test:

```rust
// TODO: figure out why this test fails:
test::check_continuous_distribution(&try_create(1.0), 0.0, 10.0);
```

I realised the problem two-fold: First, the old pdf implementation returned `Inf` at `0` instead of `0`. Second, for `freedom == 1.0` the pdf is very steep around `0` and cannot be integrated with the hard-coded step size. So I implemented a branching in the pdf and log density (also see Wikipedia) and changed `freedom` to `1.5`.

Actually, Math.NET has a little bit more branching in the chi-squared pdf, but the bulk of it is covered in `statrs`'s `gamma.rs` implementation. I'm not sure if we want to cover the case of `freedom = Inf`. @boxtown What's your opinion on this? If yay, I'll add this branching to `pdf()` and `ln_pdf()` and uncomment the unit tests. If nay, I'll just delete the commented unit tests.

### Misc

And of course added more unit tests, but this is kind of self-explanatory.

### DIFF
--- a/src/distribution/chi_squared.rs
+++ b/src/distribution/chi_squared.rs
@@ -329,7 +329,7 @@ impl Continuous<f64, f64> for ChiSquared {
         if x > 0.0 {
             self.g.ln_pdf(x)
         } else {
-            -f64::INFINITY
+            f64::NEG_INFINITY
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/distribution/chi_squared.rs
```diff
@@ -322,16 +326,20 @@ impl Continuous<f64, f64> for ChiSquared {
     /// ln(1 / (2^(k / 2) * Γ(k / 2)) * x^((k / 2) - 1) * e^(-x / 2))
     /// ```
     fn ln_pdf(&self, x: f64) -> f64 {
-        self.g.ln_pdf(x)
+        if x > 0.0 {
+            self.g.ln_pdf(x)
+        } else {
+            -f64::INFINITY
```
boxtown: conventionally we've been using `f64::NEG_INFINITY`,  I'm not actually 100% sure `-f64::INFINITY == f64::NEG_INFINITY`

src/distribution/chi_squared.rs
```diff
@@ -310,7 +310,11 @@ impl Continuous<f64, f64> for ChiSquared {
     ///
     /// where `k` is the degrees of freedom and `Γ` is the gamma function
     fn pdf(&self, x: f64) -> f64 {
-        self.g.pdf(x)
+        if x > 0.0 {
```
boxtown: I think we can be more specific, `x == 0.0` is defined (at least according to Wikipedia) as long as `freedom != 1.0`, so we can probably do something like `if x > 0.0 || (self.freedom != 1.0 && x == 0.0)`. May want to consider `prec::almost_eq` instead of `==` or `!=` for floats but I'm just spitballing

src/distribution/chi_squared.rs
```diff
@@ -366,9 +415,112 @@ mod test {
...
+        test_almost(2.0, 0.031963930603353786351, 1e-15, |x| x.pdf(5.5));
+        test_almost(2.0, 6.1810004550085248492e-25, 1e-37, |x| x.pdf(110.1));
+        test_case(2.0, 0.0, |x| x.pdf(f64::INFINITY));
+        test_case(2.5, 0.0, |x| x.pdf(0.0));
+        test_almost(2.5, 0.24812852712543073541, 1e-15, |x| x.pdf(0.1));
+        test_almost(2.5, 0.28134822576318228131, 1e-15, |x| x.pdf(1.0));
+        test_almost(2.5, 0.045412171451573920401, 1e-15, |x| x.pdf(5.5));
+        test_almost(2.5, 1.8574923023527248767e-24, 1e-36, |x| x.pdf(110.1));
+        test_case(2.5, 0.0, |x| x.pdf(f64::INFINITY));
+        // test_case(f64::INFINITY, 0.0, |x| x.pdf(0.0));
```
boxtown: What does `pdf` return for `freedom == f64::INFINITY` normally? I'd like to have this case defined but am not necessarily sure we need an explicit branch in the code

boxtown (changes requested):

src/distribution/chi_squared.rs
mp4096: It returns a `NaN`. I guess as a result of `0 * Inf` when computing the pdf.

TBH I don't like the idea of returning `0` instead of `NaN` because it doesn't make sense for me... Shouldn't a pdf integrate to `1` over support? How can it be upheld when `pdf()` returns `0` everywhere?

src/distribution/chi_squared.rs
mp4096: Sorry, I completely missed this one.

BTW, I didn't know that these constants are defined like this :grin: :

```rust
pub const INFINITY: f64 = 1.0f64 / 0.0f64
pub const NEG_INFINITY: f64 = -1.0f64 / 0.0f64
```

src/distribution/chi_squared.rs
boxtown: That's a good point. I think maybe being explicit about the `NaN` would be good, like adding a specific branch and note it in the docstring

src/distribution/chi_squared.rs
mp4096: Sorry, I screwed up here. Should've done better research.

The thing is a little bit more complicated. Wikipedia silently assumes only integer degrees of freedom. If we take `freedom` as a positive real, then we have 3 cases:
1. `0 < freedom < 2`: `pdf(0) = +∞`, hence `0` is excluded from the support
2. `freedom == 2`: `pdf(0) = 0.5`
3. `2 < freedom`: `pdf(0) = 0`

This is obvious from the pdf formula (see Wikipedia) and the `x^(k/2 - 1)` term. All other terms are non-zero for any `freedom`.

---

Anyway, I've tried out `scipy`'s implementation of chi-squared. Its behaviour is exactly as above (i.e. `+∞`, `0.5` and `0`).

---

So the question is, what to do for the case when `0 < freedom < 2`: Should we let the pdf function return `+∞` (as `scipy` does)? Or should we handle it as a separate case?

My personal preference right now is to return `+∞`, but document this behaviour in the documentation.

boxtown: Yeah that's fine by me as long as it's clear in the documentation

mp4096: Ok! Just to make sure: You prefer to return `+∞` when `0 < freedom < 2` and pdf is evaluated at `0`.

Then I'll rewrite the unit tests and update the docs.

boxtown:
Yup, I think being consistent with scipy is probably a good thing

boxtown:
Hey @mp4096 what's the status on this PR? Do you need any help bringing it across the finish line?

day01:
@YeungOnion it is outdated currently. solved and explained in #397

YeungOnion:
Aha, thanks for the ping on the parent

PR #86, opened 2018-04-20
Bug fix: Gamma distribution: scale should be multiplied

rohitjoshi:
Fixes issue: https://github.com/boxtown/statrs/issues/85  where scale is multiplied rather than divide.
With multiplication, I am able to reproduce same output as` rand` crate. 
Lines: 415 and 418

Other lines are code format.


boxtown:
I believe there is a misunderstanding since `statrs` uses the `rate` parameter rather than the `scale` parameter for Gamma distributions where `rate = 1/scale`.

vks:
I believe this can be closed. Using the scale instead of the rate would have the advantage of avoiding divisions when sampling and calculating some statistics, but it would introduce divisions when calculating the PDF and similar.

PR #87, opened 2018-06-03
statistics: fix example for statrs::statistics::Statistics::std_dev

phunehehe:


vks:
The example works as is, but I agree that it makes more sense to use the trait that is being documented.

phunehehe:
Thanks! I didn't notice that one can reach std_dev via Variance too.

PR #89, opened 2018-07-31
Update to rand 0.5

vks:
This required implementing `Distribution` in terms of Rand's
`Distribution`, because it is more general.

cc @dhardy


vks:
@bheklilr There a few cases where the math was changed:
- `Bernoulli` and `Uniform` now use the Rand implementation for sampling.
- I replaced `1. - rng.gen()` code with `rng.sample(OpenClosed01)`.
However, this should not have large effects. For all other distributions the `statrs` implementation is used even if there is a `rand` implementation.

@boxtown Yes, I think this makes sense. This is however a breaking change. The current PR does not make any breaking changes. If you don't plan another 0.9 release, I can just remove `Distribution` and `WeakRngDistribution` from `statrs`.

bheklilr:
@vks Yeah, I had noticed those changes, but based on the names I was able to guess that it was essentially the same thing.  I read over those changes pretty carefully to make sure no typos snuck in and didn't see any problems, but on some of those I just had to trust that the math itself was correct.  For most of them I was able to see where the math was moved from one impl to the new impl.

boxtown:
@vks The next planned release is 0.10.0 which I'll probably cut (hopefully sometime this week after this PR is merged in). You should be safe to remove `Distribution` and `WeakRngDistribution`.

bheklilr:
Removal of `Distribution` and `WeakRngDistribution` looks good.

boxtown:
Can we preserve the comments from the implementations of `Distribution`? The doctests are good sanity check and some of them have useful information on the algorithm used. Also, we probably don't need to alias `rand::distributions::Distribution` as `RandDistribution` anymore.

vks:
I removed them, because it is uncommon document implementations of traits.
(None of the distributions in Rand do that.) Also, the doctests are almost
identical and cause a lot of churn when they have to be updated. The
information on the algorithm could be moved elsewhere.

I'll remove the aliases.

boxtown:
Sounds good, I'm alright with merging this PR once the aliases are removed

vks:
I restored the information on the sampling implementations and removed the aliases.

boxtown:
Thanks @vks , hopefully I have the time to cut a release this week that includes your changes!

PR #93, opened 2019-01-04
Update dependency on rand

### LINKED ISSUE #92: Update the dependency on rand
There's a compatibility issue between the current version of `statrs`, which depends on `rand` 0.5 (so, up to 0.5.5), and the latest version of `rand` (0.6).

**Minimal example**

Using the `Cargo.toml` dependencies section
```toml
[dependencies]
statrs = "0.10.0"
rand = "0.6.3"
```
and trying to compile the `main.rs` file
```rust
use statrs::distribution::Normal;

use rand::Rng;
use rand::distributions::Distribution;

fn main() {
    let n = Normal::new(0., 0.).unwrap();
    let ref mut rng = rand::thread_rng();
    let v = n.sample(rng);
    println!("Value: {}", v);
}
```
will not work, and produce the trait-related error message
```
error[E0599]: no method named `sample` found for type `statrs::distribution::normal::Normal` in the current scope
 --> src\main.rs:9:15
  |
9 |     let v = n.sample(rng);
  |               ^^^^^^
  |
  = note: the method `sample` exists but the following trait bounds were not satisfied:
          `statrs::distribution::normal::Normal : rand::Rng`
  = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope, perhaps add a `use` for it:
  |
1 | use rand::distributions::Distribution;
  |
```

I forked `statrs` and did a couple of commits to update the dependency. Patching the `Cargo.toml` in the above example with my fork seems to work. I'll make a pull request

ManifoldFR:
This pull requests updates `statrs`' dependency on the `rand` random number generation crate. I re-ran the tests and benchmarks to check everything works, fixed up a couple deprecation warnings (that didn't appear when just doing `cargo build`).

Resolves #92

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 
 #editor specific
 /.vscode/
+.idea/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,4 +17,4 @@ name = "statrs"
 path = "src/lib.rs"
 
 [dependencies]
-rand = "0.5"
+rand = "0.6.3"
--- a/src/distribution/multinomial.rs
+++ b/src/distribution/multinomial.rs
@@ -249,11 +249,12 @@ impl<'a> CheckedDiscrete<&'a [u64], f64> for Multinomial {
             return Err(StatsError::ContainerExpectedSumVar("x", "n"));
         }
         let coeff = factorial::multinomial(self.n, x);
-        let val = coeff * self
-            .p
-            .iter()
-            .zip(x.iter())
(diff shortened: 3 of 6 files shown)


dhardy:
I wonder if this is the correct solution, or if we should put out a compatibility bridge between 0.5 and 0.6.

Depending on rand 0.6.0 should be good enough here.

ManifoldFR:
The command `cargo fmt --all -- --check` returned 1 as an exit code in the CI, because of some formatting issue.

Issue #74, opened 2017-10-17
Investigate usage of git pre-commit hooks


boxtown:
It would be nice to have pre-commit hooks that run rustfmt, clippy, and tests. Need to do research into a smooth workflow

paulkernfeld:
Is this still of interest to you? I have had success defining a build process using [cargo-make](https://sagiegurari.github.io/cargo-make/). The build steps are defined in a file called `Makefile.toml` ([example](https://github.com/paulkernfeld/ndarray-csv/blob/master/Makefile.toml)), and you would set up a git pre-commit hook that looked something like `cargo make all`.

If `cargo-make` doesn't seem like a good solution, this could also be done using a Bash script, but that wouldn't be as portable to platforms like Windows.

boxtown:
Yeah looks interesting. If you'd be interested in drawing something up I'd definitely take a look at it

dhardy:
Sorry to interject now, but I question the wisdom of using pre-commit hooks. If you wish all submissions to be `rustfmt`- and `clippy`-clean, then you may state that in your contributing guide (you do already for the first) and even enable CI testing if you wish (I have found this a nuisance in the past but it's viable).

Running these automatically during pre-commit could impact workflow negatively given that `rustfmt` actively adjusts code (I have had it prevent code from compiling before). Using a `pre-receive` hook may make more sense (see https://githooks.com/).

I'm not sure that @paulkernfeld's solution will help much given that many users are used to typing `cargo test` and `cargo fmt`? It is simply another way of achieving the same thing.

boxtown:
It's interesting that `rustfmt` can break your code, that doesn't seem like something that should happen. I mostly opened this issue because it's hard to tell in a PR whether or not `rustfmt` has been run, particularly with white-spacing. This then becomes a slight nuisance later on when I or someone else edits a file that sneaked in without formatting, and our PRs become bloated with formatting changes.

I'm going to play around with @paulkernfeld branch and cargo make for a little but I'm also open to other suggestions to resolve the above issue

dhardy:
> It's interesting that rustfmt can break your code, that doesn't seem like something that should happen.

They've probably fixed it by now; it was a corner-case involving a range (`..`) and other op.

I tried running `rustfmt` on the Rand project this week; nothing broke but I decided the result was in many cases worse than the current formatting (in my opinion).

boxtown:
[...]

PR #94, opened 2019-02-12
Add cargo-make to script builds

paulkernfeld:
Useful for #74

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -1,0 +1,6 @@
+[tasks.all]
+dependencies = [
+    "test",
+    "check-format",
+    "clippy",
+]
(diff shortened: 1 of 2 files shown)


boxtown:
The more I think about this one the more I'm inclined to agree with @dhardy. Although I personally feel it would be useful to have an automated process for running format and clippy, I'm not sure introducing a dependency on `cargo-make` is the solution at the moment. Thoughts?

dhardy:
This isn't really a dependency on an external tool, just a suggestion to use one. You will likely still receive PRs that are not `rustfmt` / `clippy` clean since not everyone reads "contributing" guides. So I don't think this breaks *or* fixes anything :man_shrugging:

paulkernfeld:
@dhardy yeah, I agree that there will always be some janky PRs. I do think that build automation can be useful since it makes it easier for conscientious contributors to check their work.

@boxtown since it's your project, I think the decision is up to you! `bash`, a `Makefile`, and CI are all viable alternatives to this PR.

PR #95, opened 2019-03-16
Add laplace distribution

sirfrancisbeaufort:


src/distribution/laplace.rs
```diff
@@ -0,0 +1,386 @@
+use distribution::{Continuous, Univariate};
+use rand::distributions::Distribution;
+use statistics::*;
+use rand::Rng;
+use rand::distributions::OpenClosed01;
+use std::f64;
+use {Result, StatsError};
+
+/// <summary>
```
boxtown: I'm not sure rustdoc can handle doc comments like this, but it's good to be consistent with the rest of the code base. Please refer to [the rustdoc docs](https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html) or [another distribution](https://github.com/boxtown/statrs/blob/master/src/distribution/binomial.rs#L9) for examples

sirfrancisbeaufort: Fixed it up!

src/distribution/laplace.rs
```diff
@@ -0,0 +1,363 @@
...
+use rand::Rng;
+use statistics::*;
+use std::f64;
+use {Result, StatsError};
+
+/// Continuous Univariate Laplace distribution.
+/// The Laplace distribution is a distribution over the real numbers parameterized by a mean and
+/// scale parameter. The PDF is:
+///     p(x) = \frac{1}{2 * scale} \exp{- |x - mean| / scale}.
+/// <a href="http://en.wikipedia.org/wiki/Laplace_distribution">Wikipedia - Laplace distribution</a>.
```
boxtown: You can use markdown to generate your links which I think is preferable from a consistency standpoint. See [this example documentation from the binomial distribution](https://github.com/boxtown/statrs/blob/master/src/distribution/binomial.rs#L10).

It would also be nice to include an `# Examples` section (see documentation referenced above).

src/distribution/laplace.rs
```diff
@@ -0,0 +1,363 @@
...
+pub struct Laplace {
+    location: f64,
+    scale: f64,
+}
+
+impl Laplace {
+    fn sample_unchecked<R: Rng + ?Sized>(r: &mut R, location: f64, scale: f64) -> f64 {
```
boxtown: I don't believe there's a need to split this function out from the `impl Distribution` trait as that `sample` method simply calls this one

src/distribution/laplace.rs
```diff
@@ -0,0 +1,363 @@
...
+        let r: f64 = r.gen();
+        let u = r - 0.5;
+        location - (scale * f64::signum(u) * f64::ln(1.0 - (2.0 * f64::abs(u))))
+    }
+
+    fn valid_parameter_set(location: f64, scale: f64) -> bool {
```
boxtown: Since the conditional is so simple I think (although it might sound counter-intuitive) that it would almost be more readable just checking it from within `new` rather than pulling it out into it's own function. See [this implementation of new for the normal distribution](https://github.com/boxtown/statrs/blob/master/src/distribution/normal.rs#L48)

src/distribution/laplace.rs
```diff
@@ -0,0 +1,363 @@
...
+        if Laplace::valid_parameter_set(location, scale) {
+            Ok(Laplace { location, scale })
+        } else {
+            Err(StatsError::BadParams)
+        }
+    }
+}
+
+impl Mean<f64> for Laplace {
+    /// Gets the mean of the distribution.
```
boxtown: There's a pretty standard template for documentation we've been using of

```
/// Returns the {attribute} of the X distribution
///
/// # Formula
///
/// ```ignore
/// {formula_for_attribute}
///  ```
```

Feel free to reference one of the other distributions

src/distribution/laplace.rs
```diff
@@ -0,0 +1,363 @@
...
+        test_almost(1.0, 7.0, 98.0, 1E-12, |l| l.variance());
+        test_almost(5.0, 10.0, 200.0, 1E-12, |l| l.variance());
+        test_almost(f64::INFINITY, f64::INFINITY, f64::INFINITY, 1E-12, |l| {
+            l.variance()
+        });
+    }
+
+    #[test]
+    fn test_stddev() {
+        test_almost(
```
boxtown: We've been using `#[cgf_attr(rustfmt, rustfmt_skip)]` to prevent the formatter from mangling test code. See [here](https://github.com/boxtown/statrs/blob/master/src/distribution/normal.rs#L307)

src/distribution/laplace.rs
```diff
@@ -0,0 +1,363 @@
...
+        try_create(f64::NEG_INFINITY, 0.1);
+        try_create(-5.0 - 1.0, 1.0);
+        try_create(0.0, 5.0);
+        try_create(1.0, 7.0);
+        try_create(5.0, 10.0);
+        try_create(f64::INFINITY, f64::INFINITY);
+    }
+
+    #[test]
+    fn test_create_fail() {
```
boxtown: Pedantic but for consistency we've been calling this `test_bad_create`

troublescooter:
In #125 I merged in some changes inspired due to this PR. After updating this branch to the new state of the repo the now empty empty changes was merged to credit the contibution.

PR #96, opened 2019-04-01
Add inverse CDF implementation for normal distribution

Palladinium:


bheklilr:
The code looks good, but I was unable to find a reference with the equation used here.  Do you have a source for where that equation came from?  That'd be good to add to a link in the documentation in case anyone wants to understand the math further.

Palladinium:
I copied the equation (and test cases) from Math.NET [here](https://github.com/mathnet/mathnet-numerics/blob/2361efa6543b6ff57fe5ac8870e75102eda4f240/src/Numerics/Distributions/Normal.cs#L307).

You can also obtain it from the formula on Wikipedia by applying  `erf_inv(x) = erfc_inv(1 - x)` and the fact that `erfc_inv` is odd (so `erfc_inv(-x) = -erfc_inv(x)`):
```
μ + sqrt(2) * σ * erf_inv(2x - 1)
μ + sqrt(2) * σ * erfc_inv(1 - (2x - 1))
μ + sqrt(2) * σ * erfc_inv(-2x)
μ - sqrt(2) * σ * erfc_inv(2x)
```

I'm not sure if either representation is better here, so I just went with Mat.NET's one.

PR #100, opened 2020-01-02
Implement Multivariate Normal Distribution with nalgebra

sphqxe:
This uses another crate "nalgebra-mvn" as a dependency, which is on an older version of nalgebra that depends on an older version of rand. I've opened a PR to update it to nalgebra v0.19, but for now Cargo.toml reflects that I'm compiling with a local version of nalgebra-mvn.

My first PR, please comment! :D


sphqxe:
Ok, if you prefer that I'll work on implementing the rest of the functionality within statrs and remove nalgebra-mvn as a dependency.

sphqxe:
1. Sure I'll add some tests with reference to the Math.NET lib.

2. That's true, and sampling from this distribution actually only supports f64 since it draws a vector of normals from the univariate distribution within this crate, I only realized that when I actually implemented sampling though. I'll make the changes so it only supports f64 for the rest of the APIs. I suppose having one less type parameter also reduces the complexity for users.

sphqxe:
I'm just about done with this, let me know if you think I should add more test cases.

Cargo.toml
```diff
@@ -19,3 +19,5 @@ path = "src/lib.rs"
 
 [dependencies]
 rand = "0.7"
+nalgebra = "0.19"
+num-traits = "0.2.10"
```
boxtown: do we still need the explicit dependency on `num-traits`?

sphqxe: Nope, removed it.

sphqxe:
Should be all done but I don't think I have permission to merge? Not quite sure how this works... XD

Issue #102, opened 2020-04-17
Remove degenerate cases


michaendr:
In some places, like the Gamma function, some effort is made to deal with degenerate
parameters like infinity. I think this is too much effort for a minute use case, as
I don't think users will insert these values to query the library to see what
degenerate distributions look like.

Having to deal with the special cases reduces clarity and complicates the API
considerably, as it forces to contend with mathematical details difficult to convey
in a numerical library. As the limit α -> inf is approached, the pointwise
limit of the pdf and the cdf is zero, so these aren't pdfs and cdfs anymore. As the
limit β -> inf is approached, the continuous distribution degenerates to a discrete
one: P(X = 0) = 1.

I propose to remove all special cases and document why they aren't handled.

boxtown:
I think we'd have to evaluate this on a case-by-case basis for distributions but I agree I think at least for the Gamma distribution it does not necessarily make sense to allow INF for `α` or `β`. I'll create an issue specifically for the Gamma distribution

dhardy:
FYI, we've recently had someone interested in supporting a particular [degenerate case of Exp](https://github.com/rust-random/rand/issues/969).

michaendr:
The use case referenced to justify degenerate parameter values is likely erroneous, an absorbing state can be modelled by having all outgoing paths loop back to the same state. In that case any distribution would work, unless one is interested in statistics about how many times a transition, including transitions to the same state, occurred. 

I disagree with the proposed solution and agree with the Wikipedia page that the case is degenerate for a reason.
To this argument, mathematical infinity is more often than not not an actual element of the probability space but a statement of what happens when a limit is taken. If a distribution is dependent on a parameter α then the function can be seen as depending on two variables p(α,x). If any IEEE infinities or other degenerate cases are to be allowed, then one would have to either
1. treat all possible permutations of taking the limit and find a way to query the library for each of them, as they're unequal in general
2. agree to take the limit in one specific order
3. some mixture of 1. and 2.

Additionally to these complications, as stated above, either cdf can exist but pdf be a non-classical function as is the case with a discrete probability or both limits exist but aren't cdf's and pdf's anymore. If degenerate cases are allowed where not all the functions exist in a classical sense there's an argument to be made that returning errors is more helpful than approximating a delta function with p(x0) = infinity and for all other x   p(x) = 0

michaendr:
[...]

PR #104, opened 2020-04-19
Removes gamma special cases

michaendr:
Removed also the tests which instantiated gamma with infinity as a parameter.
Tests are still failing, but don't know what the motivation is behind those
numbers.
If it doesn't matter too much I'll just update the testvalues.

### DIFF
--- a/src/distribution/gamma.rs
+++ b/src/distribution/gamma.rs
@@ -33,7 +33,11 @@ impl Gamma {
     /// # Errors
     ///
     /// Returns an error if `shape` or `rate` are `NaN`.
-    /// Also returns an error if `shape <= 0.0` or `rate <= 0.0`
+    /// Also returns an error if `shape <= 0.0` or `rate <= 0.0` or either is
+    /// `INFINITY`.
+    /// Those cases degenerate in some form or another: Either the limit isn't
+    /// a continuous distribution, the limit isn't a distribution or the limit
+    /// doesn't exist at all.
     ///
     /// # Examples
     ///
@@ -47,14 +51,15 @@ impl Gamma {
     /// assert!(result.is_err());
     /// ```
     pub fn new(shape: f64, rate: f64) -> Result<Gamma> {
-        let is_nan = shape.is_nan() || rate.is_nan();
-        match (shape, rate, is_nan) {
-            (_, _, true) => Err(StatsError::BadParams),
-            (_, _, false) if shape <= 0.0 || rate <= 0.0 => Err(StatsError::BadParams),
-            (_, _, false) => Ok(Gamma {
+        if shape <= 0.0 || shape == f64::INFINITY || shape.is_nan() {
+            Err(StatsError::BadParams)
+        } else if rate <= 0.0 || rate == f64::INFINITY || rate.is_nan() {
+            Err(StatsError::BadParams)
+        } else {
+            Ok(Gamma {
                 shape: shape,
                 rate: rate,
-            }),
+            })
         }
     }
 
@@ -109,10 +114,6 @@ impl Univariate<f64, f64> for Gamma {
     fn cdf(&self, x: f64) -> f64 {
         if x <= 0.0 {
             0.0
-        } else if x == self.shape && self.rate == f64::INFINITY {
-            1.0
-        } else if self.rate == f64::INFINITY {
-            0.0
         } else if x == f64::INFINITY {
             1.0
         } else {
@@ -154,11 +155,6 @@ impl Max<f64> for Gamma {
 impl Mean<f64> for Gamma {
     /// Returns the mean of the gamma distribution
     ///
-    /// # Remarks
-    ///
-    /// Returns `shape` if `rate == f64::INFINITY`. This behavior
-    /// is borrowed from the Math.NET implementation
-    ///
     /// # Formula
     ///
     /// ```ignore
@@ -167,11 +163,7 @@ impl Mean<f64> for Gamma {
     ///
     /// where `α` is the shape and `β` is the rate
     fn mean(&self) -> f64 {
-        if self.rate == f64::INFINITY {
-            self.shape
-        } else {
-            self.shape / self.rate
-        }
+        self.shape / self.rate
     }
 }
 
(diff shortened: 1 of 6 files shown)


boxtown:
This looks good to me, if you're done I can go ahead and merge it in lmk

michaendr:
I disallowed special parameter cases for other distributions too, the reasons being the same discussed with gamma.
It's good with me.

michaendr:
Looking back at the original issue I filed I should have probably first initiated the discussion with separate issues about those functions. Forgot about that, so I hope you don't mind.

src/distribution/gamma.rs
```diff
@@ -506,26 +460,22 @@ mod test {
...
-        test_case(10.0, f64::INFINITY, f64::INFINITY, |x| x.max());
     }
 
     #[test]
     fn test_pdf() {
-        test_case(1.0, 0.1, 0.090483741803595961836995913651194571475319347018875963, |x| x.pdf(1.0));
-        test_case(1.0, 0.1, 0.036787944117144234201693506390001264039984687455876246, |x| x.pdf(10.0));
-        test_case(1.0, 1.0, 0.36787944117144232159552377016146086744581113103176804, |x| x.pdf(1.0));
-        test_case(1.0, 1.0, 0.000045399929762484851535591515560550610237918088866564953, |x| x.pdf(10.0));
+        test_almost(1.0, 0.1, 0.090483741803595961836995913651194571475319347018875963, 1e-10, |x| x.pdf(1.0));
```
boxtown, line 471: I'm curious as to why these changed and is `1e-10` the strictest bound on them?

src/distribution/weibull.rs
```diff
@@ -308,14 +309,10 @@ impl Continuous<f64, f64> for Weibull {
     fn ln_pdf(&self, x: f64) -> f64 {
         if x < 0.0 {
             f64::NEG_INFINITY
-        } else if x == 0.0 && self.shape == 1.0 {
```
boxtown, line 311: can we add these two paths back in, the `0.0 - ` is probably unnecessary but the way they were written is to prevent unnecessary work or divisions

boxtown:
Erlang, Inverse-Gamma, and Weibull are probably fine I'll need to think a little bit on Beta and @dhardy brought up a usage for degenerate distributions that I need to look over in #102

src/distribution/gamma.rs
michaendr: I can't explain why those are removed, must be an error on my part.

I'm not sure what a good bound is or how the values in the tests were obtained. Is there a rule to follow on this?

src/distribution/weibull.rs
michaendr: I've also removed other specialisations, do you want me to put those back in too?
I'll have to profile it. It gets rid of divisions but more logarithms have to be evaluated. An additional division could be left out by writing `x.ln() - self.scale.ln()`.

src/distribution/gamma.rs
michaendr: Sorry for the confusion, I only saw the four lines outlined above and assumed I'd mistakenly removed tests.
In general testing exact values in floating point is a bad idea. I opened an issue #108 explaining my rationale.

boxtown: A lot of these were either ported directly from Math.NET or generated by comparing to other gamma distribution calculators if there were no comparable unit tests in the Math.NET library.

I definitely agree with #108 and as part of that issue it might be worth it to determine what a good bound should be. Anecdotally `1e-14` is probably as much precision as most people will need although there are definitely unit tests that don't meet that criteria (I know there's a couple that are only `1e-12`).

These are probably fine as is for now I was just asking out of curiousity

src/distribution/weibull.rs
boxtown: I don't see any other ones that were removed outside the ones that are no longer relevant due to removals of `f64::INFINITY` but good point about `ln`. Do you mind running a quick benchmark? Should be pretty simple using https://github.com/bluss/bencher/, here's a toy example https://github.com/boxtown/counter/blob/master/benches/bench.rs from another repo

michaendr: In the formulas of `pdf` and `ln_pdf` I removed the specialisation `shape = 1`.

My preference is towards predictability. Trying to manage special cases leads to uglier code with hard to predict control flow. As such tricks accumulate it gets increasingly hard to understand what's going on, with the interactions with caches and branch prediction. On some machines it may become quicker, on some slower.

michaendr:
```
running 2 tests
test weibull_new ... bench:         110 ns/iter (+/- 1)
test weibull_old ... bench:         116 ns/iter (+/- 1)
```

michaendr:
```
running 4 tests
test weibull_new_x_equals_0 ... bench:           4 ns/iter (+/- 0)
test weibull_new_x_equals_2 ... bench:         110 ns/iter (+/- 4)
test weibull_old_x_equals_0 ... bench:          25 ns/iter (+/- 0)
test weibull_old_x_equals_2 ... bench:         117 ns/iter (+/- 8)
```
Here's the bench if you want to run it.
```
#[macro_use]
extern crate bencher;

extern crate statrs;

use bencher::Bencher;
use std::f64::*;
use statrs::distribution::{Weibull,Continuous};

struct WeibullNew(Weibull);

impl WeibullNew {
    fn ln_pdf(&self, x: f64) -> f64 {
        if x < 0.0 {
            NEG_INFINITY
        } else if x == INFINITY {
            NEG_INFINITY
        } else {
            self.0.pdf(x).ln()
        }
    }
}

fn weibull_old_x_equals_0(bench: &mut Bencher) {
    let w = Weibull::new(1.0, 0.5).unwrap();
    bench.iter(|| {
        w.ln_pdf(0.0);
    })

}

fn weibull_new_x_equals_0(bench: &mut Bencher) {
    let w = WeibullNew(Weibull::new(1.0, 0.5).unwrap());
    bench.iter(|| {
        w.ln_pdf(0.0);
    })
}

fn weibull_old_x_equals_2(bench: &mut Bencher) {
    let w = Weibull::new(1.0, 0.5).unwrap();
    bench.iter(|| {
        w.ln_pdf(2.0);
    })

}

fn weibull_new_x_equals_2(bench: &mut Bencher) {
    let w = WeibullNew(Weibull::new(1.0, 0.5).unwrap());
    bench.iter(|| {
        w.ln_pdf(2.0);
    })
}

benchmark_group!(first, weibull_old_x_equals_0, weibull_new_x_equals_0);
benchmark_group!(second, weibull_old_x_equals_2, weibull_new_x_equals_2);
benchmark_main!(first,second);
```

michaendr:
Splitting into multiple pr, one for each function.

PR #106, opened 2020-04-21
[RFC] Student's T inverse CDF

asayers:
This branch is not fit for merging (see below), but I wanted to gauge interest in having this functionality in statrs.

Issues with the implementation in this branch:

* [ ] Ignores location and scale parameters (they're assumed to be 0 and 1 respectively)
* [x] Pulls in "special" and "approx" as deps
* [x] No `CheckedInverseCDF` impl
* [ ] 400 lines of unit tests is a bit much
* [ ] Docs don't describe the formula


boxtown:
I'd be happy to accept an implementation of the inverse CDF! Preferably we'd also provide an implementation of the inverse incomplete beta function to try to keep dependencies as lean as possible and provide a single source of truth for functionality :)

asayers:
> an implementation of the inverse incomplete beta function

The licenses of "special" and "statrs" are compatible so there's the option of just stealing their code.

boxtown:
We can definitely use it as a reference although in order to be consistent with the rest of the library (both for the API and coding style) there's probably a couple places that will need to be cleaned up. I can also try to find the reference I used for the incompete beta function, it's been a couple years but I think it was some reference C implementation

troublescooter:
Is there an interest in getting this PR merged?

`approx` is now a dependency and `CheckedInverseCDF` was removed.

At a glace I can't see whether `special` is maintained, it would improve discoverability if the usual candidates (Gamma, Beta, Hypergeometric, ...) were split into their own crate. I think for the moment including their code makes sense.

asayers:
I'd still be interested in seeing this merged.  I can inline the code from special into `inv_beta_inc()`.

asayers:
* Rebased on upstream master
* Renamed `inv_beta_inc()` to `inv_beta_reg()`
* Inlined the relevant code from the "special" crate.

asayers:
I _think_ including "special"'s license in the function comments is sufficient, but maybe you want to put it it LICENSE.md as well?

PR #111, opened 2020-05-18
Added Dirac Distribution [Feature Request]

JulianKnodt:
Hi, I wanted to use this library in a small side-project that I was thinking about, and I noticed that it was missing the Dirac distribution, https://en.wikipedia.org/wiki/Dirac_delta_function#As_a_distribution, which would've been useful for my use case.

Please let me know if there's anything I can do to fix this PR up to meet your standards, or anything at all!

Thanks for all your work on the library!

I was also curious if you've ever explored using the quickcheck crate for property based testing, as it seems that some of the tests for distributions are testing some basic property and could be reduced?


JulianKnodt:
I just started off with the normal distribution as a base, so I kept the format/comments there while I modified the functions. I didn't know how to implement the inverse_cdf/pdf functions, but I hope that's fine.

I noticed that I forgot to change some of the names in the documentation from normal, so I updated that. It should be good to go otherwise

Issue #82, opened 2018-03-21
Remove functions/structs deprecated in 0.9.0


boxtown:
`Periodic`
`Sinusoidal`
`Square`
`Triangle`
`Sawtooth`

troublescooter:
Fixed by #116

PR #116, opened 2020-11-11
Fix clippy warnings

troublescooter:
This pr is a mixed bag of stuff. Its main target is to fix the enormous amount of clippy lints, but also to fix some wrong special case formulas in the Gamma function, as well as introduce the crate `approx` to get a coherent handling of floating-point comparisons. Lastly it bumps dependencies.

In short, there are three possible ways to compare floating-point numbers: By absolute difference, by relative difference and by ulps, which are explained in the `approx` doc links. Absolute difference is almost never what one would want, unless some degenerate case occurs like comparing with 0. This pr introduces comparisons by ulps to be the default for inputs, while relative precision would make more sense for outputs when adapting the tests to the new changes and picking a sensible precision to target. This would fix most of the remaining 237 clippy warnings.

Two clippy warnings are silenced by default: Excessive precision and too many single-character variable names, because I didn't have access to the papers describing the algorithms and thus couldn't check whether they could have been named more descriptively.


troublescooter:
I also removed the deprecated structs that were once scheduled for removal in #82 .

PR #119, opened 2020-11-18
Ideal and robust soliton distribution

refugeesus:
Hello all. 

I have added in an implementation for the ideal and robust soliton distributions which are required in ECC and coding method implementations I am working on. To my knowledge these are correctly implemented as the libraries I utilize them in function as expected in their tests. 

There are likely methods which should be implemented for the purposes of this library which I have not completed, and if so, I would be happy to complete them so this can be merged appropriately. I hope this has been added in appropriately, as it is a bit of an odd implementation of discrete uniform distributions. If the structure should change it would be helpful to have some guidance for what is best.


src/distribution/mod.rs
```diff
@@ -269,3 +273,24 @@ pub trait CheckedDiscrete<T, K> {
     /// ```
     fn checked_ln_pmf(&self, x: T) -> Result<K>;
 }
+
+/// The 'IdealSoliton' trait provides an interface for interacting
+/// with discrete statistical distributions from integers 1..N with
+/// N as the single parameter for the distribution
+pub trait Soliton<T, K> {
```
troublescooter, line 280: Traits are often, but not always, a contract on types one does not yet know about. In this case the trait is superfluous, as there only seem to be two soliton distributions, both of which are provided.

src/distribution/robust_soliton.rs
```diff
@@ -0,0 +1,345 @@
...
+    min: i64,
+    max: i64,
+    spike: Option<i64>,
+    cumulative_probability_table: Vec<f64>,
+    ripple: Ripple,
+    fail_probability: f64,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct RobustSolitonDistribution {
```
troublescooter, line 31: This struct seems forgotten code after a refactor? It's not used anywhere and has no constructor, but private fields.

src/distribution/robust_soliton.rs
```diff
@@ -0,0 +1,345 @@
...
+    /// assert!(result.is_ok());
+    ///
+    /// result = RobustSoliton::new(-1, true, 0.1, 0.1);
+    /// assert!(result.is_err());
+    /// ```
+    pub fn new(max: i64, heuristic: bool, ripple: f64, fail_probability: f64) -> Result<RobustSoliton> {
+        if max < 1 {
+            Err(StatsError::BadParams)
+        } else {
+            let pmf_table: Vec<f64> = Vec::with_capacity(max as usize);
```
troublescooter, line 58: This type is inferred and can be elided.

src/distribution/robust_soliton.rs
```diff
@@ -0,0 +1,345 @@
...
+            &Ripple::Heuristic(n) => {
+                n * (max as f64 / fail_probability).ln() * (max as f64).sqrt()
+            }
+        }
+    }
+}
+
+impl Distribution<f64> for RobustSoliton {
+    fn sample<R: Rng + ?Sized>(&self, r: &mut R) -> f64 {
+        r.gen_range(0, 1) as f64
```
troublescooter, line 117: This seems off.

src/distribution/ideal_soliton.rs
```diff
@@ -0,0 +1,248 @@
...
+/// ```
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct IdealSoliton {
+    min: i64,
+    max: i64,
+}
+
+impl IdealSoliton {
+    /// Constructs a new discrete uniform distribution with a minimum value
```
troublescooter, line 33: There are seven doc strings referencing the discrete uniform distribution instead of the (ideal|robust) soliton distribution.

src/distribution/robust_soliton.rs
```diff
@@ -0,0 +1,345 @@
...
+    ///
+    /// # Formula
+    ///
+    /// ```ignore
+    /// sum(Xn)/n
+    /// ```
+    fn mean(&self) -> f64 {
+        let sum: f64 = Iterator::sum(self.cumulative_probability_table.iter());
+        let mean = sum / self.cumulative_probability_table.len() as f64;
+        mean
```
troublescooter, line 156: This would be clearer as `.iter().sum()`

Can you explain how this calculation works? As far as I know, there are two equivalent formulas for calculating the mean: Sum(n*p(n)) or when using cumulative probability Sum(P(X>n)) = Sum(1-P(X<=n)) = Sum(1-CPF(n))
I do not see either of them being used here.

src/distribution/robust_soliton.rs
refugeesus: Yep, missed this. Will remove.

src/distribution/robust_soliton.rs
refugeesus: :+1:

src/distribution/ideal_soliton.rs
refugeesus: Forgot to update this as had been done in robust soliton. :+1:

Issue #117, opened 2020-11-13
API discussion


troublescooter:
From my point of view, there are some issues with the current API that make the implementation of new distributions less ideal, make it less suited to handle extensions of functionality and make using this library more difficult than necessary. This issue serves to initiate the discussion on said problems and suggest changes.

### Traits

Traits generally serve to increase the upfront investment into an API to reduce the running costs of implementation. The API of a library becomes more complex the more public traits it exposes. This complexity usually pays off by reducing code duplication in generic programming, but in order to do that the implementations of composed structures must be in some way derivable from the implementations of the substructures.
1. The global properties of distributions `Mean`, `Variance`, `Entropy` and `Skewness` are not modular. There are very few ways to compose distributions that allow deriving of mean, variance, ... from the means and variances of the distributions that make up the composition. Even if desired, expressing this composition in types is a challenge.
2. The fragmentation of traits increases the overhead of implementing a new distribution. Implementors must pick the right traits and assemble them. The compiler cannot help determine whether anything is missing.
3. It's not obvious to me that there is any gain to a user of the library whether a given property (a) doesn't exist in closed form, where the implementation of the trait means to have a closed form according to the current documentation, (b) is known to not exist as a limit or (c) there exists some closed-form solution that may rely on special functions which simply have not been implemented yet. It's debatable of what use it is to incorporate any definition of closed-form in a trait whose method outputs `T`. The user will probably care most about whether a reasonable numerical estimate can be given for the distribution in question, but also defining what is closed-form is difficult. The current definition also precludes the implementation of the trait for empiric distributions or completely arbitrarily constructed distributions from a closure, which seems like a reasonable extension to support. In any case, it's likely there will be an interest in numerically determining the property of the distribution, whether there is a nice algebraic formula representing the number or not.
Does this library provide an output for this distribution that can be reasonably calculated with? If no, the user should be prevented from accidentally performing calculations with the output, but there's no complex control flow involved, so `Option<T>` should suffice. Being able to use `?` to short-circuit a calculation of the variance depending on the mean is one small improvement that is gained. `Result<T>` seems of limited additional use compared to `Option<T>`, so I do not see `Checked` versions as necessary.
4. The types in the output of the traits are too general, disabling boilerplate savings. A default implementation of `std_dev` could be provided if `T` were restricted to `T: num_traits::float::Float`.

Putting the above together, my suggestions are to
[...]

Issue #112, opened 2020-05-27
PPF functions in particular for Student's T


willir:
Hi, are there any plans for implementing the Percent Point Function (a.k.a. quantile)?

willir:
Oh, you have `InverseCDF`, it just isn't implemented for all distributions. For example it isn't implemented for Student's T. So the function is planned.

If I find time to implement it for Student's T, would it be a welcome merge request?

troublescooter:
Since there's no nice form of the inverse in general yet one would want quantiles for all implemented distributions, it would perhaps make sense to merge `InverseCDF` into `Univariate`.
A default implementation could then be given that uses binary search on the domain to approximate `inf x : F(x) >= a`.

willir:
Well, having a default implementation might be a good idea for distributions that do not have a nice form of the inverse.

CDF is a non-decreasing function, so the binary search should work. Though I'm not that good at math and I'm not aware of all corner cases that might prevent binary search implementation.

We, though, would have to change the declaration of `Univariate` trait, currently, it only requires `Min`, `Max`. We would need a comparison. And since we to stop at some epsilon, the comparison must be made with regard to some epsilon.

On the other hand, if some distributions don't "support" inverse, it might be reasonable to leave `cdf` and `ppf` (`inverse_cdf`) in separate traits, so it is more clear where it is really implemented.

In the meantime, I think I will try to implement `inverse_cdf` for `StudentsT` based on the [C++ boost implementation](https://github.com/boostorg/math/blob/develop/include/boost/math/distributions/students_t.hpp#L207).

troublescooter:
A default implementation of the inverse cdf has been provided with #120 .

Issue #110, opened 2020-05-08
Should the Statistics trait exist?


FelixBenning:
There seems to be a list of traits in the `trait.rs file`, which are grouped at least to some degree in the Statistics trait. This causes some strange things like: The Statistics trait is implemented for IntoIterator, and thus implements a mean, but Mean is not implemented for IntoIterator. If both were implemented this might actually cause problems, when calling the methods. Since the compiler won't know which one to call. 

I think having individual traits for all these functions is probably the right idea, since that allows things to implement some but not all of them and they are not interdependent. If you really wanted to group those traits you could always create a supertrait which inherits from all the other traits. But I currently can not think of a reason why this would be necessary.

FelixBenning:
Another issue: Since `Min` is implemented for `[f64]` someone will eventually try to use it on `[i32]` etc. But since `std::cmp::Ord` is implemented for `[T]` where `T:Ord`, this would cause a name collision with the min from `Ord`, if Min was implemented for `[i32]`. If Min is not implemented for `[i32]` it instead causes unexpected behaviour, since min from `Ord` lexographically compares two vectors, so it would expect another vector as a parameter. Should the method maybe be called `get_min` or `minimum` instead?

boxtown:
I think splitting up the traits with a super trait sounds pretty reasonable. I'm not too worried about the collision though, it seems like you could just do `cmp::min(a, b)` or `min(a, b)` which seems clearer than trying to do `a.min(b)`?

CGMossa:
Is there any update on this? I had an iterator and I wanted to calculate the mean of it, and just wanted to use this crate for that.. but it seems that this crate is being actively rewritten?

vks:
> I had an iterator and I wanted to calculate the mean of it, and just wanted to use this crate for that..

If that's all you want, you could have a look at my crate [average](https://github.com/vks/average) that specializes in calculating statistics on iterators.

Issue #107, opened 2020-04-25
empirical-cdf support


ronniec95:
Math.NET.Statistics.* has support for sampling data and calculating e-cdf, quantiles and so on. Is there a chance that you'll extend your feature matching to those

boxtown:
I've been pretty busy with work so haven't really planned out any new features but always open to feature requests. If you have a couple high priority requests I'd be happy to take a look at porting some of that functionality over in my spare time

troublescooter:
A bare-bones empirical distribution was implemented in #120 , and quantiles should also work, although it could do with more testing.

PR #120, opened 2020-11-26
API refactoring, Empirical distribution

troublescooter:
This pr implements the changes described in #117 , implements the empirical distribution requested in #107 , splits `Univariate` into the two traits `ContinuousCDF` and `DiscreteCDF`, implements inverse CDFs for all distributions by moving `InverseCDF` into `(Continuous|Discrete)CDF` and providing a default implementation described in #112 and removes the `CheckedInverseCDF` trait. Additionally this rearchitecturing paves the way for supporting more numerical types #17 , as after doing exploration on how the traits can be split and providing default implementations the API feels to me more elucidated that way.

As next steps, the `Statistics` trait will be removed (#110) and replaced with associated methods, as there do not seem to be many types around that would profit from such a trait.
Currently unclear is the split of the traits `Continuous` and `Discrete` to me. Sure, calling a probability mass function a pmf and not a pdf is more correct, but given the necessary split of `Univariate` into `ContinuousCDF` and `DiscreteCDF`, these traits now already signal whether the distribution is continuous or discrete, so there is some duplication going on which does not appear to be worth it. A distribution being a continuous or discrete is a worthwhile property to encode in some api contract, so there is no doubt for me that this separation has to be done somewhere.


troublescooter:
This also homogenises the multidimensional distribution types, and makes the vector types dynamic instead of static. As detailed in #118 there are trade-offs to this.

src/consts.rs
```diff
@@ -23,3 +23,6 @@ pub const TWO_SQRT_E_OVER_PI: f64 = 1.860382734205265717336249247266663112059421
 /// { 1/k - ln(n) } }`
 pub const EULER_MASCHERONI: f64 =
     0.5772156649015328606065120900824024310421593359399235988057672348849;
+
+/// Targeted accuracy instantiated over `f64`
+pub const ACC: f64 = 10e-11;
```
WarrenWeckesser, line 28: `10e-11` is equal to `1e-10`.  Was that intentional?

PR #126, opened 2020-12-31
Prevent Chi mean and variance producing NaN for dof > 340

millardjn:
The issue is mostly explained in the commit messages. Happy to answer any questions though.

### DIFF
--- a/src/distribution/chi.rs
+++ b/src/distribution/chi.rs
@@ -146,10 +146,23 @@ impl Distribution<f64> for Chi {
     fn mean(&self) -> Option<f64> {
         if self.freedom.is_infinite() {
             return None;
+        } else if self.freedom > 300.0 {
+            // Large n approximation based on the Stirling series approximation to the Gamma function
+            // This avoids call the Gamma function with large arguments and returning NaN
+            //
+            // Relative accuracy follows O(1/n^4) and at 300 d.o.f. is better than 1e-12
+            // For a f32 impl the threshold should be changed to 150
+            Some(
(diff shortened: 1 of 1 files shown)


src/distribution/chi.rs
```diff
@@ -146,10 +146,23 @@ impl Distribution<f64> for Chi {
...
         if self.freedom.is_infinite() {
             return None;
+        } else if self.freedom > 300.0 {
+            // Large n approximation based on the Stirling series approximation to the Gamma function
+            // This avoids call the Gamma function with large arguments and returning NaN
+            //
+            // Relative accuracy follows O(1/n^4) and at 300 d.o.f. is better than 1e-12
+            // For a f32 impl the threshold should be changed to 150
+            Some(
+                self.freedom.sqrt()
```
boxtown, line 156: Can we just add the stirling approximation to `functions::gamma`? I'm assuming this issue will present itself to any users of the gamma function anyways

millardjn: I don't think so. The issue with the gamma function is that the return value exceeds the maximum non-infinite f64 value at an input of ~170. This can't really be extended as it is mostly a limitation of the f64 type. Because the Chi distribution mean is the division of one gamma evaluation by another that are growing at slightly different rates, the Stirling approximation allows us to cancel out most of the extreme growth.

I think this workaround can only be used where one gamma evaluation is being divided by another with similar growth in the input value.

millardjn: Hi, just following up. Was there anything else you were after?

PR #128, opened 2021-02-11
cargo: nalgebra 0.23.0 -> 0.24.1

lovesegfault:
Was having trouble using this due to the outdated nalgebra dep.

Any chance a release can be cut following the merge of this PR?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ path = "src/lib.rs"
 
 [dependencies]
 rand = "0.7.3"
-nalgebra = "0.23.0"
+nalgebra = "0.24.1"
 approx = "0.4.0"
 num-traits = "0.2.14"
 


boxtown:
Thanks @lovesegfault I'll take a look later tonight

troublescooter:
A release right now would be suboptimal, will explain tomorrow. Going to bed.

You can use `cargo vendor` and/or the [patching mechanism](https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html) if you have a pressing need to use `statrs` as a dependency.

troublescooter:
#125 was blocked after realizing I had commented out the `slice_statistics` module without ever getting around to replacing it. I've added some changes so that the module is at least not deleted, although it might be questionable whether the module is not better suited by being replaced with `EmpiricalDistribution` since they cover similar use cases.

With these fixes a new release could be made without much further breakage, now that the new API is in full effect.

PR #132, opened 2021-04-12
fix typo: statrs::distribution::Gamma sample_unchecked

doraneko94:
I might find a mistake of sign in the random value generation function in `statrs::distribution::Gamma`.
Please refer to another implementation of "Marsaglia and Tsang’s Method", for example

https://gist.github.com/fasiha/f289c1d4b9c244a42abee54d1d6206ed

or

https://www.hongliangjie.com/2012/12/19/how-to-generate-gamma-random-variables/


vks:
For what it's worth, the sign proposed in this PR is the one that the Rand implementation used until we switched to a different algorithm.

@boxtown Does it make sense for `statrs` to use sampling algorithms in `rand_distr`?

vks (approved):

PR #134, opened 2021-04-13
Fixes and updates

### LINKED ISSUE #133: Update to rand >= 0.8
Currently, `statrs` relies on `rand` 0.7 and `nalgebra` 0.23 (which itself relies on `rand` 0.7).  The newer releases of `nalgebra` update their dependency to the latest `rand`, which is currently 0.8.3.  One of the minor, but frustrating, differences between 0.7 and 0.8 is a change in the syntax of the `rand_range` function, from taking 2 arguments to taking a single range argument.

It would be great if `statrs` could be updated to rely on a newer `nalgebra` and a newer `statrs`.  Currently, if one is depending on `rand` >= 0.8 (or is depending on any package that depends on this), then multiple different versions of `rand` are pulled down.  Not only does this bloat the build, but it can run the risk of confusing the compiler about definitions that appear in both versions of the package.

The actual changes to conform with the new API interface of `rnd_range` are small, but I think there are some other changes that would need to be made, since the version of `nalgebra` should be bumped (but perhaps not to the very latest (0.26.1), unless other changes are made because they have deprecated some interfaces that are currently used in `statrs`).

vks:
The tests and benchmarks on master are currently broken. They are fixed in this PR. Additionally, some warnings are fixed and the dependencies are updated.

Fixes #128.
Fixes #133.


rob-p:
Is there any chance this will be merged soon?  I'm looking to tag a release of something that depends on statrs, and I'd prefer to pull from the official crate if possible.  However, if there are reasons that this won't be merged soon (or potentially) at all, then it is worth it to pull from the fork for the time being to avoid the rand 0.7 dependency creeping in.  Thanks!

troublescooter:
I currently don't have the permissions to make a release, if that's what you need, but I'll try and get in touch with the owner.
If a merge is what you need, I'll take care of it today after work.

troublescooter:
Thanks!

Issue #149, opened 2021-08-21
Broken link to "Exp distribution" Wikipedia article in documentation.


vi:
[Documentation](https://docs.rs/statrs/0.15.0/statrs/distribution/struct.Exp.html) links https://en.wikipedia.org/wiki/Exp_distribution, while probably should link to https://en.wikipedia.org/wiki/Exponential_distribution.

Issue #150, opened 2021-08-22
Is statrs::distribution::Laplace correct?


vi:
Assuming it is [this Laplace distribution](https://en.wikipedia.org/wiki/Laplace_distribution), generated samples look strangely distributed:

![q](https://user-images.githubusercontent.com/173219/130338534-89b31e57-e862-4555-b7ca-1cf0d8e134d8.png)

Wikipedia shows symmetric density function, but all values generated by `statrs::distribution::Laplace` are positibe (assuming zero `location` parameter).

[Formula](https://github.com/statrs-dev/statrs/commit/fd3a447cd02ef031a58cf086fe9a703357026644#diff-fe4427f3325a5e2ecbd8bfefe46c9a542af8404013fdca6377ce2e83ed7345deR85) used in the code reuses uniform float value both as sign and as exponential distribution generator, but actually has no way to ever generate negative values.

boxtown:
Looks like there's a bug [here](https://github.com/statrs-dev/statrs/blob/master/src/distribution/laplace.rs#L85)

Should be

```Rust
self.location - self.scale * x.signum() * (1.0 - 2.0 * x.abs()).ln()
```

vi:
@boxtown Are you going to submit a pull request for this?

PR #151, opened 2021-08-22
fix laplace distribution sampling

boxtown:
Sampling has a bug where negative values are not being generated due to missing `abs()` for random variable. See https://github.com/statrs-dev/statrs/issues/150

This PR fixes that issue and adds a sanity check test for sampling to ensure balanced draws.
Also fixes https://github.com/statrs-dev/statrs/issues/149

### DIFF
--- a/src/distribution/laplace.rs
+++ b/src/distribution/laplace.rs
@@ -486,7 +486,7 @@ mod tests {
             });
             assert!(
                 result > -tolerance && result < tolerance,
-                "Balanace is {} for seed {}",
+                "Balance is {} for seed {}",
                 result,
                 seed
             );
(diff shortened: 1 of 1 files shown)


vi:
Maybe the test can be extended to all symmetric distributions, not just the Laplace one?

src/distribution/laplace.rs
```diff
@@ -458,4 +458,38 @@ mod tests {
...
+                    sum + 1
+                } else if val < 0.0 {
+                    sum - 1
+                } else {
+                    0
+                }
+            });
+            assert!(
+                result > -tolerance && result < tolerance,
+                "Balanace is {} for seed {}",
```
vks, line 489: "Balanace" -> "Balance"

vks, line 489:
```suggestion
                "Balance is {} for seed {}",
```

vks (approved):
Looks good!

troublescooter:
LGTM

vi:
When to expect the fix to propagate to crates.io?

vi:
Why this and other accepted fixes are being withheld from crates.io?

PR #153, opened 2021-08-28
Add logistic distribution

dsaxton:
This PR adds the logistic distribution to `statrs`, following the goal of trying to port over the distributions from https://github.com/mathnet/mathnet-numerics/tree/master/src/Numerics/Distributions. It's deviating a bit from some of the other implementations in `statrs` (e.g., Laplace) in that infinite parameters are not permitted, since I think it's incorrect to allow this. The pointwise test cases were calculated using the Python `scipy` library.

https://en.wikipedia.org/wiki/Logistic_distribution


src/distribution/logistic.rs
```diff
@@ -0,0 +1,454 @@
...
+    ///
+    /// ```ignore
+    /// INF
+    /// ```
+    fn max(&self) -> f64 {
+        f64::INFINITY
+    }
+}
+
+impl Distribution<f64> for Logistic {
```
dsaxton, line 151: It seems we're using dedicated traits for some things like min and max, whereas others are nested inside this more general Distribution trait. Could it make sense to only do one or the other?

troublescooter: The reason these have not been coalesced is because it was not clear to me how useful these other properties outside of `Distribution` were at the time I wrote the traits into their current form. Max and min for example have no generalization to higher dimensions. `Mode` might be a (n-1)D subset if distribution is nD.
I can't think of a reason why I didn't include `Median` into `Distribution` though, so I could revisit that. I'm generally speaking for fewer traits, because I don't see traits here pulling their weight. How to properly integrate higher dimensional distributions is also unfinished.

dsaxton: That makes sense, thanks for the explanation @troublescooter

src/distribution/logistic.rs
```diff
@@ -0,0 +1,454 @@
...
+    /// ```ignore
+    /// -(x - μ) / σ - 2 * ln(1 + exp(-(x - μ) / σ)) - ln(σ)
+    /// ```
+    ///
+    /// where `μ` is the location and `b` is the scale
+    fn ln_pdf(&self, x: f64) -> f64 {
+        -(x - self.location) / self.scale
+            - self.scale.ln()
+            - 2.0 * (1.0 + (-(x - self.location) / self.scale).exp()).ln()
+    }
```
troublescooter, line 260: Unless there's an example of a range of values which can more accurately be represented by expanding the logarithm, I prefer not to do that. Multiplications are numerically more stable than additions and are simpler to read. In a range where overflow occurs it might make sense, but dealing with all those cases is unpleasant. If the scale is small, both expressions will run into issues.

dsaxton: Updated, let me know if this way will work

PR #155, opened 2021-11-05
Support sampling integers from discrete distributions

SabrinaJewson:


henryjac:
Very reasonable PR, adding to milestone 0.17 and we'll probably get this merged! Preliminary checks looks good to me.

SabrinaJewson:
For most the distributions, their implementations effectively use integers anyway and this PR just exposes that functionality.

The only exception is the geometric distribution:

```rust
x.log(1.0 - self.p).ceil() as u64
```

The largest finite value this expression can be is if `x = 1.4e-45` and `1.0 - self.p = 0.9999999999999999`, in which case we get `930262250532780300`, which when casted to a `u64` is `930262250532780288`. Importantly it does stay within the range of a `u64`, so I would say the lossiness is acceptable here.

src/distribution/geometric.rs
```diff
@@ -68,17 +68,23 @@ impl Geometric {
     }
 }
```
YeungOnion, line 70: Would you be open to a remark in the docstring mentioning how this uses f64 internally but that accurate values would be within u64? Perhaps something similar for hypergeometric if that's a feasible analysis.

SabrinaJewson: I’ve added a comment to explain it. I don’t think it applies to hypergeometric because the implementation of that is just adding 1 to an integer a number of times, which will never overflow a `u64` anyway.

YeungOnion:
Thanks for getting back on this despite how long this was left sitting!


Two requests,

1. Would you reword to use a conventional commit message so I can rely on automatic change log generation tools? Just a `feat: ` prefix on the first line of the commit would be great. 

2. Are you able to write some regression tests against the pre-existing float output?

YeungOnion:
Hey, would you want me to resolve the conflicts? I imagine most of them are for the "rand" feature and you've already done the first merge conflict, and I think this is a good feature.

SabrinaJewson:
I’ve rebased on master and used conventional commits. I’m not sure how to do the regression tests: random number generation is hard to test because, well, it’s random – we could fix a seed but I’m sceptical of adding that to the tests, because it’s not tesitng public API (meaning we could make a non-SemVer-breaking change and have to change the tests, which isn’t considered good testing practice).

That said, I’ve refactored some of the implementations to share as much code as possible, so it should be obviously the case that the correctness of the new algorithm is the same as the correctness of the old algorithm.

YeungOnion:
Thanks! It all looks good, but I suppose it won't hurt to run CI checks again.

---
An aside regarding regression tests.
Initially I thought perhaps to just sample an integer and cast to float, then compare it to float, but as you mentioned at the start of this PR, most rely on integer internally, so that's just checking float representation accuracy.

YeungOnion:
Thanks again!

PR #157, opened 2021-12-29
Improve geometric distribution CDF so it handles small p correctly.

WarrenWeckesser:
The subtraction `1 - p` in the formula `1 - (1 - p) ^ x` loses precision when `p` is small.  The mathematically equivalent expression  `-expm1(log1p(-p)*x)` avoids the loss of precision.


troublescooter:
Thanks!

Issue #156, opened 2021-11-25
StudentsT inverse_cdf is broken


scullionw:
It is currently only valid for ranges .25-0.5 and .75-1, which is probably why the tests don't catch it.

```rust
use statrs::distribution::{ContinuousCDF, StudentsT};

fn t_ppf(q: f64, freedom: f64, location: f64, scale: f64) -> f64 {
    StudentsT::new(location, scale, freedom)
        .unwrap()
        .inverse_cdf(q)
}

fn main() {
    let mut x = 0.01;

    while x < 1.0 {
        let y = t_ppf(x, 2.74, 0.0, 1.0);
        debug_plotter::plot!((x, y));
        x += 0.01
    }
}
```

![src-main rs:14:9](https://user-images.githubusercontent.com/11668344/143498378-8bac5f1b-35a0-4f0f-9e8b-4c9b675baf6c.png)

it should look like this:

<img width="502" alt="Screen Shot 2021-11-25 at 3 48 12 PM" src="https://user-images.githubusercontent.com/11668344/143498597-3e98c805-3bdd-4646-a23c-d49753629832.png">

I currently fix it in my code with

```rust
fn t_ppf(q: f64, freedom: f64, location: f64, scale: f64) -> f64 {
    let candidate = StudentsT::new(location, scale, freedom)
        .unwrap()
        .inverse_cdf(q);

    if q < 0.25 {
        candidate * -1.0
    } else if q < 0.5 {
        candidate
    } else if q < 0.75 {
        candidate * -1.0
    } else {
        candidate
    }
}
```

WarrenWeckesser:
I create a pull request with a fix: https://github.com/statrs-dev/statrs/pull/159.

PR #159, opened 2022-01-03
Fix inverse_cdf for StudentsT.

WarrenWeckesser:
Closes gh-156.


troublescooter:
Thanks for the fix. I'll review it this weekend.

WarrenWeckesser:
@troublescooter, I noticed a problem with the PR, and since I don't think anyone has taken a close look at this yet, I force-pushed an update.

troublescooter:
I'll use this opportunity to move over the tests to the new testing macro seen in [internal.rs](https://github.com/statrs-dev/statrs/blob/master/src/distribution/internal.rs) which has not yet been applied to StudentsT. This takes care of the targeted relative accuracy. This will take a day or two, otherwise LGTM.

troublescooter:
Thanks!

Issue #160, opened 2022-01-16
BUG: `laplace.inverse_cdf()` is not correct.


WarrenWeckesser:
The implementation of the `inverse_cdf()` method for the Laplace distribution is missing calls to `ln()`.
The following
https://github.com/statrs-dev/statrs/blob/435739b180913fa0e815403836eb74cd79041d60/src/distribution/laplace.rs#L127-L131

should be
```rust
        if p <= 0.5 {
            self.location + self.scale * (2. * p).ln()
        } else {
            self.location - self.scale * (2. - 2. * p).ln()
        }
```
(See, for example, the entry for *Quantile* in the table on the right in the [Laplace distribution wikipedia page](https://en.wikipedia.org/wiki/Laplace_distribution).)

I'll submit a pull request with a fix.

PR #161, opened 2022-01-16
fix: Fix laplace.inverse_cdf().

WarrenWeckesser:
The expressions in the implementation of `laplace.inverse_cdf()` were
missing calls to `ln()`.

Tests for the `cdf()` and `inverse_cdf()` methods are included.

Closes gh-160.

### DIFF
--- a/src/distribution/laplace.rs
+++ b/src/distribution/laplace.rs
@@ -1,5 +1,5 @@
 use crate::distribution::{Continuous, ContinuousCDF};
-use crate::statistics::*;
+use crate::statistics::{Distribution, Max, Median, Min, Mode};
 use crate::{Result, StatsError};
 use rand::Rng;
 use std::f64;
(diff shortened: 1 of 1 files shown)


WarrenWeckesser:
I added the helper function `test_rel_close` to emulate the existing patterns used in the test module.  To be honest, though, I'd be happy with something like this
```
    #[test]
    fn test_cdf() {
        let loc = 0.0f64;
        let scale = 1.0f64;
        let laplace = Laplace::new(loc, scale).unwrap();
        let reltol = 1e-15f64;

        // Expected value from Wolfram Alpha: CDF[LaplaceDistribution[0, 1], 1/2].
        let expected = 0.69673467014368328819810023250440977f64;
        assert_relative_eq!(laplace.cdf(0.5), expected, epsilon = 0.0, max_relative = reltol);

        // Wolfram Alpha: CDF[LaplaceDistribution[0, 1], -1/2]
        let expected = 0.30326532985631671180189976749559023f64;
        assert_relative_eq!(laplace.cdf(-0.5), expected, epsilon = 0.0, max_relative = reltol);

        // Wolfram Alpha: CDF[LaplaceDistribution[0, 1], -100]
        let expected = 1.8600379880104179814798479019315592e-44f64;
        assert_relative_eq!(laplace.cdf(-100.0), expected, epsilon = 0.0, max_relative = reltol);
    }
```
which has a little less abstraction and seems clearer to me.

(`rustfmt` doesn't like those long lines, so the calls of `assert_relative_eq` would actually end up with each parameter on a separate line.)

PR #162, opened 2022-01-28
Update nalgebra to 0.30

Palladinium:


YeungOnion:
Updated to 0.32 with #187

Issue #115, opened 2020-07-06
Characterization of negative binomial extended to the reals


boxtown:
See https://github.com/boxtown/statrs/commit/6b2c726aa924b4187565a381690f7d345c16ca4c#diff-8f4094da35079696cbcdd4a269333ad0R101 for a little more context.

We're currently allowing the negative binomial distribution to accept positive real `r` values as it seems to be a common use case. However, this causes the CDF to behave continuously rather than like in other discrete distributions which breaks our internal `check_discrete_distribution` test.

I'm not familiar enough with this particular distribution to provide much guidance but there are a couple options:

1. simply treat the distribution as continuous, allowing real value inputs for the "probability mass functions"
2. Create a special case test for negative binomial (not sure how often this case will arise in practice)
3. Create a special case distribution type

Option 1 seems a vastly superior choice but possibly semantically wrong (although also possibly to the point of pedantry)

WarrenWeckesser:
`statrs` is still version 0.x, so wouldn't it be acceptable to change the type of `r` to an integer?  I agree with @dsaxton; every reference that I've seen refers to the negative binomial distribution as a discrete distribution, so allowing that parameter to be continuous is unusual. 

If it turns out later that there is sufficient demand for a continuous version, then perhaps a new distribution could be added then, with a name such as `ContinuousNegativeBinomial`, that makes it explicit that it is not the traditional discrete distribution.  This would avoid the need to shoehorn an inherently discrete distribution into a continuous API, or vice versa.

WarrenWeckesser:
On the other hand, `r` is just a parameter of the distribution; whether that parameter is float or integer shouldn't change the "discreteness" of the distribution.

> However, this causes the CDF to behave continuously rather than like in other discrete distributions which breaks our internal `check_discrete_distribution test`.

Could you explain what goes wrong?   How does `r` being a float make the CDF "behave continuously"? The two tests at the end of `negative_binomial.rs` that use `check_discrete_distribution` (currently commented out) pass when I uncomment them. (I'll experiment further, but some details on the nature of the problem might be helpful.)

PR #164, opened 2022-02-02
Some updates for NegativeBinomial

WarrenWeckesser:
* Uncomment the check_discrete_distribution tests.
* Expand the documentation--explain the meaning of `r` and `p`.
* Note that `r` can be real.

Closes gh-115.


boxtown:
I've long forgotten the context to #115 but this seems reasonable and if the tests pass then I think it's fine

PR #167, opened 2022-05-06
Add fisher exact

Maxyme:
Adding fisher's exact test (inspiration directly from SciPy). I also created a new folder to host statistical tests.


src/stats_tests/fisher.rs
```diff
@@ -0,0 +1,358 @@
+use crate::distribution::{Discrete, DiscreteCDF, Hypergeometric};
+use crate::StatsError;
+
+#[derive(Debug, Copy, Clone)]
```
YeungOnion, line 4: A brief docstring on this enum would be nice for our docs.rs, but I don't think one is needed for each of the variants.

Maxyme: I approved your PR into this branch with your changes, thank you!

YeungOnion:
Thanks for your contribution and for introducing statistical tests!

Issue #171, opened 2022-08-08
Negative binomial cdf floating point precision not within range of f64 minimum positive.


noamteyssier:
Hey all, I noticed a small discrepancy between floating point precision of the negative binomial CDF when compared to scipy. For very unlikely observations it seems like statrs is rounding down to zero even though the precision should be within range of f64 precision. I suspect there's some rounding going on somewhere in the regularized incomplete beta function calculation when under some set precision but haven't explored too deeply yet. 

Here's a minimal reproducible example of the rust code and the python code for comparison of results.
```rust
use statrs::{distribution::{NegativeBinomial, DiscreteCDF}, function::beta::beta_reg};

fn main() {
    let (x, r, p) = (2321., 269., 0.0405);
    let cdf = NegativeBinomial::new(r, p).unwrap().cdf(x as u64);
    println!("{:?}", cdf);
}
```
which results in `0.0`

and the scipy example which under the hood pipes to boost ibeta
```python3
from scipy.stats import nbinom

x, r, p = 2321, 269, 0.0405
cdf = nbinom.cdf(x, r, p)
print(cdf)
```
which results in `2.803588112796817e-43`

For visualization this is what the parameterized distribution and associated observation looks like - you can see this is incredibly unlikely so ultimately the value is incredibly close to zero, but since the f64 minimum positive is `2.2250738585072014E-308f64` it should still be within range of precision. 
![image](https://user-images.githubusercontent.com/22600644/183465255-551e669b-c603-41ce-93e5-a1259532d74b.png)

WarrenWeckesser:
The value `2.80358811[...]e-43` can be independently verified with [Wolfram Alpha](https://www.wolframalpha.com/input?i=CDF%5BNegativeBinomialDistribution%5B269%2C+405%2F10000%5D%2C+2321%5D).

The problem is here:
https://github.com/statrs-dev/statrs/blob/0c8fb1a18d1ddf95cf40b35ae3968d8745dc4cd8/src/distribution/negative_binomial.rs#L114
`beta_reg` is the regularized lower incomplete beta function.  For the given parameters, the 64 bit floating point result of that `beta_reg` call is 1.0, so when it is subtracted from 1.0 the result is 0; the difference that we wanted is lost.

By using the symmetry of the function (see, for example, [Incomplete beta function - Properties](https://en.wikipedia.org/wiki/Beta_function#Properties_2)), that expression can be changed to

[...]

PR #172, opened 2022-08-18
Survival functions

noamteyssier:
# Survival Functions

## Reasoning
This is an attempt to create a survival function for each of the distributions to avoid as much precision loss as possible (#171).

## Background
The survival function is the complement function of the cumulative density function`sf( x ) => 1 - cdf( x )` - but because there is always precision loss when subtracting a very small float from 1. it is always better to calculate the sf directly to avoid any precision loss. 

This is also the approach taken by boost and scipy - and was implemented for the same reason as mentioned above. 

## Methodology
I've put the reasoning behind my implementations here if anyone is interested in reading through them. Many of these were easily solved by identity properties on the functions being used. For example - the regularized incomplete beta 
```
1 - beta_reg(a, b, x) = beta_reg(b, a, 1-x)
```` 
Or alternatively if the cdf is calculated with the **lower** incomplete regularized gamma function `gamma_lr` then the complement can easily be calculated using the **upper** incomplete regularized gamma function `gamma_ur`.  

## Testing
I've also added equivalent testing for all the survival functions for each of the distributions. In these cases I either calculated the 1-cdf(x) of the test case after checking wolfram alpha or just inverted the case if it was obvious (i.e. if it was expecting a 0.25 for the cdf it should expect a 0.75 for the sf).  

--- 

# Discrete Distributions
The discrete distributions were extended to calculate the survival function by adding a new function `sf` to the `DiscreteCDF` trait. 

## Bernoulli Distribution

Bernoulli is a wrapper over an internal binomial distribution where a single trial is conducted.
As a result it is a fixed-case binomial distribution and returns the cdf/sf of that fixed cased binomial. 
```rust
/// Where b is a Binomial(p, 1)
fn cdf(&self, x: u64) -> f64 {
    self.b.cdf(x)
}
```
and so the SF can be calculated the same way by calling the internal binomial SF
```rust
/// Where b is a Binomial(p, 1)
fn sf(&self, x: u64) -> f64 {
    self.b.sf(x)
}
```

## Binomial Distribution

The binomial distribution uses a regularized incomplete beta function to calculate it's CDF.
Luckily this function is symmetrical (as in the case of the negative binomial) and so the survival function could calculated by inverting the arguments to the incomplete beta. 

Here is the current implementation of the binomial CDF
```rust
fn cdf(&self, x: u64) -> f64 {
    if x >= self.n {
        1.0
    } else {
        let k = x;
        beta::beta_reg((self.n - k) as f64, k as f64 + 1.0, 1.0 - self.p)
    }
}
```

and the new implementation of its SF: note the swapped ordering of the beta_reg arguments and the complement `self.p`
```rust
fn sf(&self, x: u64) -> f64 {
    if x >= self.n {
        0.0
    } else {
        let k = x;
        beta::beta_reg(k as f64 + 1.0, (self.n - k) as f64, self.p)
    }
}
```

## Categorical Distribution
The categorical cumulative distribution is currently calculated via the following method
```rust
pub struct Categorical {
    norm_pmf: Vec<f64>,
    cdf: Vec<f64>,
}
fn cdf(&self, x: u64) -> f64 {
    if x >= self.cdf.len() as u64 {
        1.0
    } else {
        self.cdf.get(x as usize).unwrap() / self.cdf_max()
    }
}
```
Essentially a cumulative distribution is calculated by performing a cumulative sum of the un-normalized probability mass.
So a survival function would need to be the mirror of this process and subtract each value in the cdf by the total sum
For example: 
```
pmf = [1, 1, 1, 1]
cdf => [1, 2, 3, 4]
sf => [3, 2, 1, 0]
-----
pmf = [4, 2.5, 2.5, 1]
cdf => [4, 6.5, 9, 10]
sf => [6, 3.5, 1, 0]
```

Here is the current implementation of the CDF
```rust
fn cdf(&self, x: u64) -> f64 {
    if x >= self.cdf.len() as u64 {
        1.0
    } else {
        self.cdf.get(x as usize).unwrap() / self.cdf_max()
    }
}
```
and the new implementation of the SF
```rust
fn sf(&self, x: u64) -> f64 {
    if x >= self.sf.len() as u64 {
        0.0
    } else {
        self.sf.get(x as usize).unwrap() / self.cdf_max()
    }
}
```

## Discrete Uniform

The discrete uniform cdf is calculated via the following formula:
``` 
(floor(x) - min + 1) / (max - min + 1) 
```
```rust
fn cdf(&self, x: i64) -> f64 {
    if x < self.min {
        0.0
    } else if x >= self.max {
        1.0
    } else {
        let lower = self.min as f64;
        let upper = self.max as f64;
        let ans = (x as f64 - lower + 1.0) / (upper - lower + 1.0);
        if ans > 1.0 {
            1.0
        } else {
            ans
        }
    }
}
```

This should be easily mirrored by instead subtracting value from the max and removing the +1 
```
(max - floor(x)) / (max - min + 1)
```
```rust
fn sf(&self, x: i64) -> f64 {
    if x < self.min {
        1.0
    } else if x >= self.max {
        0.0
    } else {
        let lower = self.min as f64;
        let upper = self.max as f64;
        let ans = (upper - x as f64) / (upper - lower + 1.0);
        if ans > 1.0 {
            1.0
        } else {
            ans
        }
    }
}
```

## Geometric

The geometric cdf is calculated via the following formula:
```
cdf => 1 - (1-p)**x
```
So the survival function can actually using the second term explicitly
```
sf => (1-p)**x
```
However the actual implementation can be done without any direct (1-p) using built in ln_1p
```
(1-p) ^ x = exp(log(1-p) * x)
               = exp(log1p(-p) * x)
```
Here is the current implementation of the CDF
```rust
fn cdf(&self, x: u64) -> f64 {
    if x == 0 {
        0.0
    } else {
        // 1 - (1 - p) ^ x = 1 - exp(log(1 - p)*x)
        //                 = -expm1(log1p(-p)*x))
        //                 = -((-p).ln_1p() * x).exp_m1()
        -((-self.p).ln_1p() * (x as f64)).exp_m1()
    }
}
```

and the new implementation of the SF
```rust
fn sf(&self, x: u64) -> f64 {
    // (1-p) ^ x = exp(log(1-p)*x)
    //           = exp(log1p(-p) * x)
    if x == 0 {
        1.0
    } else {
        ((-self.p).ln_1p() * (x as f64)).exp()
    }
}
```

## Hypergeometric

The hypergeometric cdf is defined via the product of binomial coefficients using the provided parameters and the generalized hypergeometric function.
However, since it is discrete, it can be calculated directly without using the generalized hypergeometric function by summing the probability mass function over the required range.
The current implementation calculates the cdf via a forward integration, i.e. summing from 0 (could be made the bound min) to k.
The survival function could then be considered the inverse, and can be calculated by summing from k+1 to the bound max. 

This implementation works - but I did notice that there tends to be precision loss for large population sizes - looking at the [boost hypergeometric](https://www.boost.org/doc/libs/1_52_0/libs/math/doc/sf_and_dist/html/math_toolkit/dist/dist_ref/dists/hypergeometric_dist.html) it seems as though the log-gamma calculation (here I'm thinking of the log pmf calculation) introduces precision loss as a function of larger population sizes. 

In their implementation they also calculate the cdf/sf depending on where the provided k falls in the distribution itself then performs the integration:
```
The CDF and its complement is calculated by directly summing the PDF's. We start by deciding whether the CDF, or its complement, is likely to be the smaller of the two and then calculate the PDF at k (or k+1 if we're calculating the complement) and calculate successive PDF values via the recurrence relations:
```

I think this section of the code in statrs could be improved upon to calculate the cdf in the same way - but will leave it outside the scope of generating survival functions more broadly. 

Here is the current implementation of the cdf - keeping in mind it is just an evaluation of the pmf over the internval range (0..k)
```rust
fn cdf(&self, x: u64) -> f64 {
    if x < self.min() {
        0.0
    } else if x >= self.max() {
        1.0
    } else {
        let k = x;
        let ln_denom = factorial::ln_binomial(self.population, self.draws);
        (0..k + 1).fold(0.0, |acc, i| {
            acc + (factorial::ln_binomial(self.successes, i)
                + factorial::ln_binomial(self.population - self.successes, self.draws - i)
                - ln_denom)
                .exp()
        })
    }
}
```
and the newly calculated sf - which is the same operation but evaluated over the interval range (k..max)
```rust
fn sf(&self, x: u64) -> f64 {
    if x < self.min() {
        1.0
    } else if x >= self.max() {
        0.0
    } else {
        let k = x;
        let ln_denom = factorial::ln_binomial(self.population, self.draws);
        (k + 1 .. self.max() + 1).fold(0.0, |acc, i| {
            acc + (factorial::ln_binomial(self.successes, i)
                + factorial::ln_binomial(self.population - self.successes, self.draws - i)
                - ln_denom)
                .exp()
        })
    }
}
```

## Negative Binomial

The negative binomial cdf is calculated using a regularized incomplete beta function and with the internal parameters r and p. 
The cdf is calculated as `I_p(r,  x+1)` and so the survival function will be inverted as `I_(1-p)(x+1, r)`

This solution was shown to me by @WarrenWeckesser in #171 and was the starting point for this whole escapade. 

I replaced the current implementation of the negative binomial with their solution as well as create the survival function. 
Here is the new cdf which passes all the tests with increased precision
```rust
fn cdf(&self, x: u64) -> f64 {
    beta::beta_reg(self.r, x as f64 + 1.0, self.p)
}
```
and the survival function which takes its complement without the 1 minus operation
```rust
fn sf(&self, x: u64) -> f64 {
    beta::beta_reg(x as f64 + 1.0, self.r, 1. - self.p)
}
```

## Poisson

The poisson cdf is calculated using the the formula `1 - gamma_lr` where `gamma_lr` is the lower incomplete regularized gamma function. 

```rust
fn cdf(&self, x: u64) -> f64 {
    gamma::gamma_ur(x as f64 + 1.0, self.lambda)
}
```
However, this could be simplified by calculating instead the upper incomplete regularized gamma function and not introducing any 1m precision loss. 
This makes it easy to implement the survival function as it will then be the gamma_lr without the 1m preceding it.

```rust
fn sf(&self, x: u64) -> f64 {
    gamma::gamma_lr(x as f64 + 1.0, self.lambda)
}
```

# Continuous Distributions
The continuous distributions were extended by adding a function `sf` to the `ContinuousCDF` trait. 

## Cauchy

This distribution's cdf is calculated using the arctan function: 
```
(1 / pi) * arctan( x - x_0 / y) + 0.5
```
Using a negative argument with arctan flips its sign, so this survival function can derived by just flipping the sign of the internal argument to arctan
```
(1 / pi) * arctan(-( x - x_0 / y )) + 0.5
```
Here is the current implementation of the Cauchy CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    (1.0 / f64::consts::PI) * ((x - self.location) / self.scale).atan() + 0.5
}
```
and the new implementation of its SF
```rust
fn sf(&self, x: f64) -> f64 {
    (1.0 / f64::consts::PI) * (-((x - self.location) / self.scale)).atan() + 0.5
}
```

## Chi
This distribution calculates its cdf using the lower incomplete regularized gamma function - so the complement is easily calculated by using the upper incomplete regularized gamma function. 

Here is the current implementation of the Chi CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    if self.freedom == f64::INFINITY || x == f64::INFINITY {
        1.0
    } else if x <= 0.0 {
        0.0
    } else {
        gamma::gamma_lr(self.freedom / 2.0, x * x / 2.0)
    }
}
```
and the new implementation of its SF
```rust
fn sf(&self, x: f64) -> f64 {
    if self.freedom == f64::INFINITY || x == f64::INFINITY {
        0.0
    } else if x <= 0.0 {
        1.0
    } else {
        gamma::gamma_ur(self.freedom / 2.0, x * x / 2.0)
    }
}
```

## Chi-Squared

This distribution actually relies on an internal gamma distribution - and so its survival function is actually the survival function of its gamma. 
As a result both its CDF and SF just pass to the relevant gamma.

```rust
fn cdf(&self, x: f64) -> f64 {
    self.g.cdf(x)
}

fn sf(&self, x: f64) -> f64 {
    self.g.sf(x)
}
``` 

## Dirac

This distribution is zero everywhere except zero. the cdf is easily calculated by just flipping the logic of currently calculated method:
```
cdf => if x < self.0 { 0.0 } else { 1.0 }
sf  => if x < self.0 { 1.0 } else { 0.0 }
```
```rust
fn cdf(&self, x: f64) -> f64 {
    if x < self.0 {
        0.0
    } else {
        1.0
    }
}

fn sf(&self, x: f64) -> f64 {
    if x < self.0 {
        1.0
    } else {
        0.0
    }
}
```

## Empirical

This swap was done by just reverse the search on the internal binary tree for the empirical distribution. 
Overall I think this could be calculated more cleanly/functionally - but for the time being I just kept the same style as the cdf calculation. 

Here is how the current CDF is calculated:
```rust
fn cdf(&self, x: f64) -> f64 {
    let mut sum = 0;
    for (keys, values) in &self.data {
        if keys.0 > x {
            return sum as f64 / self.sum;
        }
        sum += values;
    }
    sum as f64 / self.sum
}
```

and how the new SF is calculated. Note the reversed iterator and flipped comparator
```rust
fn sf(&self, x: f64) -> f64 {
    let mut sum = 0;
    for (keys, values) in self.data.iter().rev() {
        if keys.0 <= x {
            return sum as f64 / self.sum;
        }
        sum += values;
    }
    sum as f64 / self.sum
}
```

## Erlang

The Erlang distribution is a less generalized form of the gamma distribution where the k parameter is limited to integers as opposed to the entire real number space. as a result the cdf and survival functions are calculated with the internal gamma distribution.

```rust
// where self.g is the internal gamma distribution

fn cdf(&self, x: f64) -> f64 {
    self.g.cdf(x)
}

fn sf(&self, x: f64) -> f64 {
    self.g.sf(x)
}
```

## Exponential

The exponential distribution calculates the cdf as follows:
```
1 - e^(-λ * x)
where `λ` is the rate
```
and so the survival function is easily calculated using the following method:
```
e^(-λ * x)
where `λ` is the rate
```

Here is the current implementation of the Exponential CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    if x < 0.0 {
        0.0
    } else {
        1.0 - (-self.rate * x).exp()
    }
}
```
and the newly implemented SF
```rust
fn sf(&self, x: f64) -> f64 {
    if x < 0.0 {
        1.0
    } else {
        (-self.rate * x).exp()
    }
}
```

## Fisher-Snedecor

This distribution's cdf is calculated using the lower incomplete regularized beta function.
```
I_((d1*x)/(d1*x+d2)) (d1/2, d2/2)
```
 This means that the complement can be calculated using the mirror property of the beta reg function. 
```
I_(1 - (d1*x/d1*x+d2) (d2/2, d1/2)
```

Here is the current implementation of the Fisher-Snedecor CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    if x < 0.0 {
        0.0
    } else if x.is_infinite() {
        1.0
    } else {
        beta::beta_reg(
            self.freedom_1 / 2.0,
            self.freedom_2 / 2.0,
            self.freedom_1 * x / (self.freedom_1 * x + self.freedom_2),
        )
    }
}
```
and its newly implemented SF. Note the swapped order of arguments in the `beta_reg` call and the 1-p. 
```rust
fn sf(&self, x: f64) -> f64 {
    if x < 0.0 {
        1.0
    } else if x.is_infinite() {
        0.0
    } else {
        beta::beta_reg(
            self.freedom_2 / 2.0,
            self.freedom_1 / 2.0, 
            1. - ((self.freedom_1 * x) / (self.freedom_1 * x + self.freedom_2))
        )
    }
}
```

## Gamma

The gamma distribution uses the lower incomplete gamma function to calculate its cdf.
This is an easy complement as we just need to swap in the upper incomplete gamma function to calculate its survival function. 

Here is the current implementation of the gamma CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    if x <= 0.0 {
        0.0
    } else if ulps_eq!(x, self.shape) && self.rate.is_infinite() {
        1.0
    } else if self.rate.is_infinite() {
        0.0
    } else if x.is_infinite() {
        1.0
    } else {
        gamma::gamma_lr(self.shape, x * self.rate)
    }
}
```
and the new implementation of its SF
```rust
fn sf(&self, x: f64) -> f64 {
    if x <= 0.0 {
        1.0
    }
    else if ulps_eq!(x, self.shape) && self.rate.is_infinite() {
        0.0
    }
    else if self.rate.is_infinite() {
        1.0
    }
    else if x.is_infinite() {
        0.0
    }
    else {
        gamma::gamma_ur(self.shape, x * self.rate)
    }
}
```

## Inverse Gamma

The inverse gamma distribution is the reciprocal of the gamma distribution
It calculates its cdf using the upper incomplete gamma function - which means its survival function can be calculated using the lower incomplete gamma function. 

Here is the current implementation of the Inverse Gamma CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    if x <= 0.0 {
        0.0
    } else if x.is_infinite() {
        1.0
    } else {
        gamma::gamma_ur(self.shape, self.rate / x)
    }
}
```

and its newly implemented SF
```rust
fn sf(&self, x: f64) -> f64 {
    if x <= 0.0 {
        1.0
    } else if x.is_infinite() {
        0.0
    } else {
        gamma::gamma_lr(self.shape, self.rate / x)
    }
}
```

## Laplace

As this cdf is currently calculated with an if-statement all I did here was reverse the logic so the arms calculate the complement.

Here is the current implementation of the Laplace CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    let y = (-(x - self.location).abs() / self.scale).exp() / 2.;
    if x >= self.location {
        1. - y
    } else {
        y
    }
}
```

and the newly implemented SF. Note the reversal in the arm returns. 
```rust
fn sf(&self, x: f64) -> f64 {
    let y = (-(x - self.location).abs() / self.scale).exp() / 2.;
    if x >= self.location {
        y
    } else {
        1. - y
    }
}
```

## Log Normal

The log normal distribution uses a similar cdf to the normal distribution. 
it calculates its cdf using the normal cdf function Φ[ln(x) - μ / σ sqrt(2)] 
However, this function is easily complemented due to the fact that it is symmetric around zero. 
As a result the following property is true: 
```
Φ(-x) + Φ(x) = 1
Φ(-x)        = 1 - Φ(x)
```
And so the complement can be calculated by just inverting the sign of the argument. 

Here is the current implementation of the Log Normal CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    if x <= 0.0 {
        0.0
    } else if x.is_infinite() {
        1.0
    } else {
        0.5 * erf::erfc((self.location - x.ln()) / (self.scale * f64::consts::SQRT_2))
    }
}
```

and the new implementation of the SF. Note the sign change in the erfc
```rust
fn sf(&self, x: f64) -> f64 {
    if x <= 0.0 {
        1.0
    } else if x.is_infinite() {
        0.0
    } else {
        0.5 * erf::erfc((x.ln() - self.location) / (self.scale * f64::consts::SQRT_2))
    }
}
```

## Normal

The previous example with the log-normal showed a clean way of calculating the complement cdf - I used the same methodology here when calculating the normal distribution survival function. 
These functions were implemented with an unchecked call - I believe this is the case because they are also called by the Student's T distribution in certain cases.  

Here is the current implementation of the Normal CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    cdf_unchecked(x, self.mean, self.std_dev)
}
pub fn cdf_unchecked(x: f64, mean: f64, std_dev: f64) -> f64 {
    0.5 * erf::erfc((mean - x) / (std_dev * f64::consts::SQRT_2))
}
```
and the new implementation of the SF. Note the sign change in the erfc
```rust
fn sf(&self, x: f64) -> f64 {
    sf_unchecked(x, self.mean, self.std_dev)
}
pub fn sf_unchecked(x: f64, mean: f64, std_dev: f64) -> f64 {
    0.5 * erf::erfc((x - mean) / (std_dev * f64::consts::SQRT_2))
}
```

## Pareto

The Pareto distribution calculates its cdf using the formula:
```
cdf = 1 - (beta / x) ^ alpha
```
and so the sf is easily calculated by subtracting this from one
```
sf = (beta/x) ^ alpha
```
Looks like boost actually calculates this with a trick to avoid the 1m operation in the cdf. 
I'll circle back on this one when I get a chance to try to increase precision here.

Here is the current implementation of the Pareto CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    if x < self.scale {
        0.0
    } else {
        1.0 - (self.scale / x).powf(self.shape)
    }
}
```

and the new implementation of the SF
```rust
fn sf(&self, x: f64) -> f64 {
    if x < self.scale {
        1.0
    } else {
        (self.scale / x).powf(self.shape)
    }
}
```

## Student's T

This cdf was already optimized to return the correct half of the lower incomplete regularized beta depending on where cdf was expected to be calculated. as a result all I needed to do was flip the logic of the if statement arms and return the opposite calculated value. 

Here is the current implementation of the Student's T CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    if self.freedom.is_infinite() {
        super::normal::cdf_unchecked(x, self.location, self.scale)
    } else {
        let k = (x - self.location) / self.scale;
        let h = self.freedom / (self.freedom + k * k);
        let ib = 0.5 * beta::beta_reg(self.freedom / 2.0, 0.5, h);
        if x <= self.location {
            ib
        } else {
            1.0 - ib
        }
    }
}
```

and the new implementation of the SF. note the if/statement arm change
```rust
fn sf(&self, x: f64) -> f64 {
    if self.freedom.is_infinite() {
        super::normal::sf_unchecked(x, self.location, self.scale)
    } else {
        let k = (x - self.location) / self.scale;
        let h = self.freedom / (self.freedom + k * k);
        let ib = 0.5 * beta::beta_reg(self.freedom / 2.0, 0.5, h);
        if x <= self.location {
            1.0 - ib
        } else {
            ib
        }
    }
}
```

## Triangular

The triangular distribution CDF is calculated with the piecewise function:
```
if x == min {
    0
} if min < x <= mode {
    (x - min)^2 / ((max - min) * (mode - min))
} else if mode < x < max {
    1 - (max - min)^2 / ((max - min) * (max - mode))
} else {
    1
}
```
and so it can easily be complemented by doing a complement on each of the arms
```
if x == min {
    1
} if min < x <= mode {
    1 - (x - min)^2 / ((max - min) * (mode - min))
} else if mode < x < max {
    (max - min)^2 / ((max - min) * (max - mode))
} else {
    0
}
```

Here is the current implementation of the Triangular CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    let a = self.min;
    let b = self.max;
    let c = self.mode;
    if x <= a {
        0.0
    } else if x <= c {
        (x - a) * (x - a) / ((b - a) * (c - a))
    } else if x < b {
        1.0 - (b - x) * (b - x) / ((b - a) * (b - c))
    } else {
        1.0
    }
}
```

and the newly implemented SF
```rust
fn sf(&self, x: f64) -> f64 {
    let a = self.min;
    let b = self.max;
    let c = self.mode;
    if x <= a {
        1.0
    } else if x <= c {
        1.0 - ((x - a) * (x - a) / ((b - a) * (c - a)))
    } else if x < b {
        (b - x) * (b - x) / ((b - a) * (b - c))
    } else {
        0.0
    }
}
```

## Uniform

This one was simple - used the same methodology as the discrete uniform
```
cdf = (x - min) / (max - min)
sf = (max - x) / (max - min)
```
I also added some bounds checking on the combinations of infinite x and infinite max to avoid NaN generation with divides.

Here is the current implementation of the Uniform CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    if x <= self.min {
        0.0
    } else if x >= self.max {
        1.0
    } else {
        (x - self.min) / (self.max - self.min)
    }
}
```

and the newly implemented SF
```rust
fn sf(&self, x: f64) -> f64 {
    if x <= self.min {
        1.0
    } else if x >= self.max {
        0.0
    } else if x.is_infinite() && self.max.is_infinite() {
        0.0
    } else if self.max.is_infinite() {
        1.0
    } else {
        (self.max - x) / (self.max - self.min)
    }
}
``` 

## Weibull

The cdf for this is calculated via the following equation - which is easy to complement as it just requires removing the one minus operation
```
cdf => 1 - e^-((x/λ)^k)
sf  => e^-((x/λ)^k)
```

Here is the current implementation of the Weibull CDF
```rust
fn cdf(&self, x: f64) -> f64 {
    if x < 0.0 {
        0.0
    } else {
        -(-x.powf(self.shape) * self.scale_pow_shape_inv).exp_m1()
    }
}
```

and the newly implemented SF. Note the missing negative on the left and the direct `exp` instead of the `exp_m1`
```rust
fn sf(&self, x: f64) -> f64 {
    if x < 0.0 {
        1.0
    } else {
        (-x.powf(self.shape) * self.scale_pow_shape_inv).exp()
    }
}
```


boxtown:
Hey thanks for opening this PR! Definitely seems like a useful addition and the PR description is probably one of the most thorough I've seen 👍 . It's been a while since I've actively maintained this project but give me a couple days to review this and we can hopefully get this merged.

Maybe @troublescooter and @WarrenWeckesser can give it a once over too

src/distribution/mod.rs
```diff
@@ -88,6 +88,21 @@ pub trait ContinuousCDF<K: Float, T: Float>: Min<K> + Max<K> {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use statrs::distribution::{ContinuousCDF, Uniform};
+    ///
+    /// let n = Uniform::new(0.0, 1.0).unwrap();
+    /// assert_eq!(0.5, n.sf(0.5));
+    /// ```
+    fn sf(&self, x: K) -> T;
```
boxtown, line 104: I wonder if it might be useful to provide a degenerate default implementation of `1 - self.cdf(x)`

src/distribution/cauchy.rs
```diff
@@ -99,6 +99,22 @@ impl ContinuousCDF<f64, f64> for Cauchy {
...
+    ///
+    /// ```ignore
+    /// (1 / π) * arctan(-(x - x_0) / γ) + 0.5
+    /// ```
+    ///
+    /// where `x_0` is the location and `γ` is the scale.
+    /// note that this is identical to the cdf except for
+    /// the negative argument to the arctan function
+    fn sf(&self, x: f64) -> f64 {
+        (1.0 / f64::consts::PI) * (-((x - self.location) / self.scale)).atan() + 0.5
```
boxtown, line 116: Would it be cleaner to write `* (self.location - x) / self.scale).atan()` to reduce the noise?

boxtown (approved):
Looks good to me for the most part. Excellent work! Give it maybe a couple of days to see if others want to chime in. If not I think we can probably just merge this

src/distribution/cauchy.rs
noamteyssier: yeah great catch! I made this change and pushed it to branch - definitely cleans it up

src/distribution/mod.rs
noamteyssier: Yeah I had wondered the same thing, but I ran into an issue where since the cdf is returning a generic `T` there was no implementation for `{float} - T` when I wrote up the `1 - self.cdf(x)`

I don't have much experience with generics, but I think it could be an easy addition for someone who did!

noamteyssier:
Looking at the snippet above I realized that the existing docstrings for `cdf` and `sf` in `DiscreteCDF` showed continuous distribution examples. 

I updated them to be discrete with this latest commit.

PR #174, opened 2022-10-24
Mention quantile function in docstring

rikhuijzer:
First, of all, thanks for making and maintaining this project. I'm trying to make a wasm project and I would have probably given up without `statrs`.

Anyway, I'm always confused by the many synonyms used in statistics. This PR suggests to add one synonym to the `inverse_cdf` to the docstring of the normal distribution. This should make it at least visible when searching the docs?

Adding the note to all `inverse_cdf` implementations seems a bit redundant so that's why I kept it to the Normal for now.

### DIFF
--- a/src/distribution/normal.rs
+++ b/src/distribution/normal.rs
@@ -100,7 +100,8 @@ impl ContinuousCDF<f64, f64> for Normal {
     }
 
     /// Calculates the inverse cumulative distribution function for the
-    /// normal distribution at `x`
+    /// normal distribution at `x`.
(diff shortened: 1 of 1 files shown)


YeungOnion:
I'd added the quantile function to inverse_cdf as a doc alias in another PR.

Still open to including the verbiage mentioning that this name is used in other tools like R, could you also mention it on the trait method? I think that normal is a good place to have it since it's such a prolific distribution and likely seen early on reading the docs.

YeungOnion:
I added it to the docs for the trait, merging this to have it in `Normal`, where it's also likely to be seen.

PR #176, opened 2022-12-23
Multivariate students t distribution

henryjac:
Implementation of Multivariate students t distribution in a very similar way to Multivariate normal. 
Includes sampling, mean, covariance, mode, pdf and log pdf functions. 

Testing with exact values from python scipy.stats.multivariate_t functions. Large degrees of freedoms does not work yet


henryjac:
I put you as a reviewer @YeungOnion, but you don't have to review this PR if you don't want to, but I'm just hesitant of merging my own PR now.

YeungOnion:
Thanks! I took a first glance.

I see the tests for large DOF issues, did you fix that or is the failure mode visible to a user for cases of "should be close enough, but not guaranteed"

Additionally, you have TODO suggesting more matrices to tests, but what else might you add? If I recall correctly, real positive semi-definite is a bijection to real symmetric which means any other tests would really be testing the precision of `nalgebra`, no?

henryjac:
About the large DOF issue, I don't remember exactly what I thought about back then, but as I see now the correspondence between mutlivariate normal and mutlivariate t for DOF quite high should go to zero (as it does for the f64::INFINITY test), but the exact difference in value for finite DOF isn't really known to me.

For the other point, more tests should increase the confidence to the user. And any real symmetric matrix is not necessarily positive semi-definite (we are only interested in positive definite though), take 
$`A = \begin{bmatrix}0&1\\1&0\end{bmatrix}`$
which has eigenvalues $`\lambda = \pm 1`$, so it isn't definite. 
I realize I have added tests for some matrices that aren't just diagonal, but I'm adding some more for 3x3 matrices at least, with using the expressions in [scipy/tests/test_multivariate_t](https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/stats/tests/test_multivariate.py#L2381) (based on MATLAB).

I am also doing some more minor changes, but should be good soon.

Also, I don't know the best place to discuss it openly, but we should define what we wish to do with existing PR and issues and in what direction to take the crate.

YeungOnion:
Well it's certainly not necessary to aim for better than scipy's implementation, and you've tested the asymptotic behavior. 

Regarding the theorem on symmetric matrices, I certainly recalled that wrong. 

Oh yeah, adding tests definitely adds confidence, I think that's great. The test you have, including the one you most recently added seems sufficient for implementing a new future. Perhaps opening an issue that's good for first-time contributors to add tests might be valuable.

I think you're right about an open place to discuss. Perhaps this is a little meta but I think we can open an issue to discuss direction and we can update the readme with any goals we (the community) establish.

YeungOnion:
@henryjac do you want to update this to use the static allocator syntax with nalgebra before merging #209?

Issue #168, opened 2022-06-13
I wonder why `MultivariateNormal::new` need for `Vec<f64>`


chatblanc-ciel:
I thinking using `nalgebra`'s vector and matrix is good.
For example,
```
fn new_from_nalgebra(mean :nalgebra::DVector<f64>, cov :nalgebra::DMatrix<f64>) -> statrs::Result<Self> {
/* here is not necessary
        let mean = DVector::from_vec(mean);
        let cov = DMatrix::from_vec(mean.len(), mean.len(), cov);
*/
        let dim = mean.len();
        // Check that the provided covariance matrix is symmetric
// omitted others
```

What do you think about this?

PR #177, opened 2022-12-25
Use nalgebra structs for constructing multivariate dist

henryjac:
Implementation of #168,
Move the checks from `new` to `new_from_nalgebra` and call `new_from_nalgebra` from `new` after converting to DVector/DMatrix


YeungOnion (approved):
Looks good to me, approved!

@henryjac we didn't hash out how we want to go through some of the approval process beyond not reviewing our own PRs since there are two of us, so feel free to provide me any feedback here.

PR #179, opened 2022-12-31
Add default implementation of survival functions

henryjac:
Adds a default implementation of survival functions as $1-cdf(x)$ with
generics, which is 
```
T::one() - self.cdf(x)
```


henryjac:
See reference: [PR 172](https://github.com/statrs-dev/statrs/pull/172/files#r951988536)

PR #187, opened 2023-03-28
Update nalgebra to 0.32

kalev:


Heeten:
Any change of getting this merged? @kalev just curious what you're doing in the meantime as I need this and am wondering if I should use my own fork for the time being.

kalev:
I ended up patching it downstream in Fedora in the mean time, https://src.fedoraproject.org/rpms/rust-statrs/c/3b588181bafc5bb693f334a2c0bde0fd910e4c47?branch=rawhide

YeungOnion:
Hi, I'm a new co-maintainer! I have no opposition to this, nothing acts up as I increase version dependence incrementally by minor release and none of nalgebra's [changelog](https://github.com/dimforge/nalgebra/blob/dev/CHANGELOG.md) seems to an issue for what we have here.

However, as @henryjac (our other maintainer) and I just accepted the role, we haven't figured out how we'll go about maintaining this, and I think this should be part of decisions about a next release. What do you think @henryjac?

henryjac:
I think we should add some of the other existing PRs like this together with some more pressing issues. [Issue 195](https://github.com/statrs-dev/statrs/issues/195) is probably something we should look into as it concerns copyright issues.

Overall I think getting out a release soon would be a good first milestone and show that the crate is still being maintained.

YeungOnion:
@henryjac, sounds like it'll be useful to make a gh milestone for 0.16 release and add things to it so we can discuss on those issues. I'll do that and add this issue and the NIST one for now. Let me know if you have a different suggestion.

henryjac:
Sounds good! I'll see what other reasonable PRs that could be added. 

Also, I think the release would be 0.17. It's a bit confusing with the numbering with it being both 0.15 and 0.16 at some places, but the latest release on [crates.io](https://crates.io/crates/statrs) is 0.16.

PR #190, opened 2023-06-07
Fix: Exponential distribution inverse CDF accuracy & performance

tessob:
Hello,

While working on my project, I found that this function is calculated with a rather large margin of error. So, I had replaced the generic implementation with the [specific one](https://en.wikipedia.org/wiki/Exponential_distribution#Quantiles). If you comment out my implementation of the trait method, you will see what I am talking about from the test errors.

Also have find a bit weird behaviour of `cargo test` command when it simply ignores my tests, and probably some other tests too.

BR,
.

### DIFF
--- a/src/distribution/exponential.rs
+++ b/src/distribution/exponential.rs
@@ -120,7 +120,7 @@ impl ContinuousCDF<f64, f64> for Exp {
     ///
     /// where `p` is the probability and `λ` is the rate
     fn inverse_cdf(&self, p: f64) -> f64 {
-        -(1.0 - p).ln() / self.rate
+        -(-p).ln_1p() / self.rate
     }
 }
 
--- a/src/distribution/exponential.rs
+++ b/src/distribution/exponential.rs
@@ -109,6 +109,19 @@ impl ContinuousCDF<f64, f64> for Exp {
             (-self.rate * x).exp()
         }
     }
+
+    /// Calculates the inverse cumulative distribution function.
+    ///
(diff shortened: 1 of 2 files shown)


src/distribution/exponential.rs
```diff
@@ -109,6 +109,19 @@ impl ContinuousCDF<f64, f64> for Exp {
...
+    ///
+    /// # Formula
+    ///
+    /// ```ignore
+    /// -ln(1 - p) / λ
+    /// ```
+    ///
+    /// where `p` is the probability and `λ` is the rate
+    fn inverse_cdf(&self, p: f64) -> f64 {
+        -(1.0 - p).ln() / self.rate
```
WarrenWeckesser, line 123: To avoid loss of precision when `p` is small, this should use `ln_1p`:
```suggestion
        -(-p).ln_1p() / self.rate
```

Magnetuz:
Hi @tessob.

I cloned your branch and ran the tests with `cargo +nightly test --features nightly --lib`.

I can see your test `test_inverse_cdf` running and passing. So it is fine.

Best regards.

tessob:
Hi @Magnetuz,

Thank you. I tested my code before commit. It just took some time to comprehend what was necessary to do. I think that a `CONTRIBUTION.md` file is needed in this repository.

YeungOnion:
Thanks for the PR, we'd like to implement `inverse_cdf` over search where possible.

PR #192, opened 2023-08-07
Fix invalid `k` letter in Poisson formula docs.

iago-lito:
For some reason, these formulae used a mixture of `k` and `x` for the same parameter. I haven't checked whether this problem is spread more widely within the docs though.

### DIFF
--- a/src/distribution/hypergeometric.rs
+++ b/src/distribution/hypergeometric.rs
@@ -140,8 +140,8 @@ impl DiscreteCDF<u64, f64> for Hypergeometric {
     /// # Formula
     ///
     /// ```ignore
-    /// 1 - ((n choose k+1) * (N-n choose K-k-1)) / (N choose K) * 3_F_2(1,
-    /// k+1-K, k+1-n; k+2, N+k+2-K-n; 1)
+    /// 1 - ((n choose x+1) * (N-n choose K-x-1)) / (N choose K) * 3_F_2(1,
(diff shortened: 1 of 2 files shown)


YeungOnion:
Thanks for catching this, seems like there's similar - but less erroneous imo - usages of $k$ where the docs previously reference `x`, there's more consistency in the docs using `x` for the parameter of functions characterizing a distribution.

Would you be willing to add the ones in the `distribution::hypergeometric` in this PR as well?

---

Generally, I have no opposition to readings of statements in the manner that always uses $x$ in terms of a function parameter,

$$
K \sim \textrm{Pois}(\lambda)
$$

$$
\implies \textrm{Pr}(K = x) = f_\lambda(x)
$$

As it's also good for other 1D random variables, even if multi parameter,

$$
X \sim \textrm{HypGeom}(N,n,\vec{p}); |\vec{p}|_1 = K
$$

$$
\implies \textrm{Pr}(X = x) = f_{N,n,\vec{p}}\ (x)
$$

as well as continuous _and_ to the same notation for CDF as well, which _could_ be defined for float even on discrete distributions (of appropriate support).

iago-lito:
Hello. I am not exactly sure I understand your request, but maybe this additional commit is what you expect?

YeungOnion:
Yes! Thanks for figuring it out, I now realize how I could have better demonstrated what I meant.

Issue #194, opened 2023-10-15
[request] Update nalgebra dependency


saona-raimundo:
I was using the [MultivariateNormal](https://docs.rs/statrs/0.16.0/statrs/distribution/struct.MultivariateNormal.html).
Specifically, the [continuous trait](https://docs.rs/statrs/0.16.0/statrs/distribution/struct.MultivariateNormal.html#impl-Continuous%3C%26%27a%20Matrix%3Cf64%2C%20Dynamic%2C%20Const%3C1%3E%2C%20VecStorage%3Cf64%2C%20Dynamic%2C%20Const%3C1%3E%3E%3E%2C%20f64%3E-for-MultivariateNormal).

In my case, I had the data in ndarray, so I passed it to nalgebra. 
But the `nalgebra` has a new version of 1d arrays. 

For reference, this is the extract from rustc
```
                              gaussians[i].pdf(&nalgebra::base::DVector::from(point.to_vec())))
    |                                                --- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&Matrix<f64, Dynamic, ..., ...>`, found `&Matrix<f64, Dyn, Const<1>, ...>`
    |                                                |
    |                                                arguments to this method are incorrect
    |
    = note: `Matrix<f64, Dyn, Const<1>, ...>` and `Matrix<f64, Dynamic, Const<1>, ...>` have similar names, but are actually distinct types
note: `Matrix<f64, Dyn, Const<1>, ...>` is defined in crate `nalgebra`
   --> ...\nalgebra-0.32.3\src\base\matrix.rs:175:1
    |
175 | pub struct Matrix<T, R, C, S> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: `Matrix<f64, Dynamic, Const<1>, ...>` is defined in crate `nalgebra`
   --> ...\nalgebra-0.29.0\src\base\matrix.rs:157:1
    |
157 | pub struct Matrix<T, R, C, S> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: perhaps two different versions of crate `nalgebra` are being used?
```

saona-raimundo:
As a second thought, would it not be better to `impl<'a> Continuous<&'a [f64]> for MultivariateNormal`?

Heeten:
[...]

PR #199, opened 2024-01-30
feature: Implement Continuous<Vec<f64>> for MultivariateNormal

saona-raimundo:
Solves #194

The problem was the public dependency on `nalgebra::Dvector`:
Although statrs does not update its dependencies, users should easily sample from MultivariateNormal.

The users were using an updated version of nalgebra and the types had clashes. 
Therefore, this implements the same trait for a type with guaranteed backward compatibility: `Vec`.

Maintaining or not the `nalgebra::DVector` in the public API is an open question that we did not touch.

### DIFF
--- a/src/distribution/multivariate_normal.rs
+++ b/src/distribution/multivariate_normal.rs
@@ -202,6 +202,28 @@ impl<'a> Continuous<&'a DVector<f64>, f64> for MultivariateNormal {
     }
 }
 
+impl Continuous<Vec<f64>, f64> for MultivariateNormal {
+    /// Calculates the probability density function for the multivariate
+    /// normal distribution at `x`
+    ///
+    /// # Formula
+    ///
+    /// ```ignore
+    /// (2 * π) ^ (-k / 2) * det(Σ) ^ (1 / 2) * e ^ ( -(1 / 2) * transpose(x - μ) * inv(Σ) * (x - μ))
+    /// ```
+    ///
+    /// where `μ` is the mean, `inv(Σ)` is the precision matrix, `det(Σ)` is the determinant
+    /// of the covariance matrix, and `k` is the dimension of the distribution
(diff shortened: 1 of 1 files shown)


YeungOnion:
Hmm, I believe I merged this in a rush. I'd yet to get input here on #209 which reintroduces the coupling to internal and at API usage of `nalgebra` and I was considering the option for an `multidimensional_nalgebra` feature to allow using this without using an internal `nalgebra` which would be a taller order.

saona-raimundo:
As far as I could see, the problem of #194 (requiring the user to use possibly more than one version of `nalgebra`) is also present on #209 
If this is the case, then I see no problem in accepting both interfaces: `Vec` can always be used, while `nalgebra` can be used easily if there is no dependency clashes. 
Before we make it efficient or convenient, it should "just work", and I think `Vec` is needed in this case.

YeungOnion:
I think you're right that both will work for now as it (and similar for later `ContinuousCDF`) solves the usage of trait `Continuous`.

However, I think the traits that return `OVector` (e.g. `MeanN`, `VarianceN`, `rand::distributions::Distribution`) would return library constrained version of `OVector`. That sound right? Maybe that'd be another issue?

saona-raimundo:
Ah! You are right.
I would say they are different kind of problems.
One thing is the user needing to input an `nalgebra` struct of a specific version, while another is `statrs` returning an `nalgebra` struct of a specific version.
They both introduce a possible dependency problem, but they have different solutions.

I have not tested it but the second might not be much of a problem.
If the struct that `statrs` returns has a method to construct a `Vec` with the underlying data, and the user does not need to add a dependency for that, then this is not much of a problem. 
We can mention it in the documentation.

YeungOnion:
Ah yes, it's on the other side. I'll write it as a separate issue, thanks for your input!

PR #211, opened 2024-04-08
Remove lazy-static and make FCACHE a proper const

anergictcell:
This minor change removes the `lazy_static` dependency by making the `FCACHE` a `const`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ rand = "0.8"
 nalgebra = { version = "0.29", features = ["rand"] }
 approx = "0.5.0"
 num-traits = "0.2.14"
-lazy_static = "1.4.0"
 
 [dev-dependencies]
 criterion = "0.3.3"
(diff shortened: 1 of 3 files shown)


YeungOnion:
Dropping another dependency would be nice. Do you know if this will still be static or will it work like inlining?

FreezyLemon:
A `const` is not static in the C/C++ sense of the word. You should expect inlining when you use a `const` variable.

That said, you don't have to switch to `const`. `static` will mostly work the same way (just declare it as `static FCACHE`) and will instead put it into some memory location in the binary instead of inlining.

YeungOnion:
Thanks for commenting on this one @FreezyLemon, I didn't know there's more chance of optimizations, so I learned something new.

I think FCACHE appears only twice in the entire crate. An additional `[f64; 170]` is a lot smaller than the crate. Seems like a great idea to change to `const`. Will merge!

Issue #210, opened 2024-03-16
Bug: Mode calculation for Gamma distribution


tessob:
Gamma distribution with the following parametrization:
```rust
Gamma::new(0.2422859569300416, 0.0000033637229958478293)
```
produces really weird mode value – `Some(-225260.53542615683)`.

[In Wolfram there is no mode for this distribution](https://www.wolframalpha.com/input?i=GammaDistribution%5B0.2422859569300416%2C+1%2F0.0000033637229958478293%5D)

YeungOnion:
resolved by #212

PR #212, opened 2024-04-08
fix: mode for gamma distribution is 0 for shape<=1

YeungOnion:
addresses #210, opt for mode at singular point of x=0 in lieu of None.


tessob:
Maybe it should be `Option::None` instead? As negative shape is not valid value to construct Gamma.

YeungOnion:
Yeah, now that you mention it, `None` for `alpha<1.0` since it's singular and one sided makes more sense.

PR #213, opened 2024-04-12
Various chores

YeungOnion:
A few sets of chore commits
- a set of edits to documentation concerning intro to crate, usage, and contrib
- update Cargo.toml to ~point at correct repo~ not explicit docs, since all on docs.rs
- update examples, update syntax and link to docs.rs
- first response to [this PSA](https://users.rust-lang.org/t/psa-check-if-your-cargo-crates-are-clean-and-tagged/109264) on Rust user forum


FreezyLemon:
Can I suggest disabling the `macros` feature in nalgebra?

```toml
nalgebra = { version = "0.29", default-features = false, features = ["std", "rand"] }

# or:

[dependencies.nalgebra]
version = "0.29"
default-features = false
features = ["std", "rand"]
```

It looks like it is unnecessary for `statrs` and reduces the amount of dependencies & compile time a bit

YeungOnion:
Considering making all multivariate nalgebra dependencies a feature so it can be opted out of, and I think #209 will only use the macros in tests.

This PR is already non-specific, but mostly related to docs, but the idea is great; I'll call out the mention in #209.

FreezyLemon:
Thanks! Another small thing, this one might be suited for this PR: The README uses CRLF line endings, which is a bit unexpected because all other files in the project seem to use LF only. Bit of a nitpick, but basically free to fix

YeungOnion:
reordered commits in hopes to seem a little more organized.
I also missed one of the math texts that was listed as `ignore` instead of `text`

Issue #195, opened 2023-10-19
non-free NIST data


decathorpe:
We are packaging this crate for Fedora Linux as a dependency of some other applications. During technical and legal review, it was discovered that this crate redistributes part of some NIST data set - which appears to be covered by a license that makes it not freely redistributable.

We are not able to distribute this data as part of the package for the statrs crate, so we need to repackage a "cleaned" version of this crate, and cannot use the version published on crates.io directly.

It is likely also not allowed for this crate to redistribute this dataset by publishing it on crates.io (or even here on GitHub). It might be a good idea to at least exclude the NIST data files from published crates.

For reference, the discussion thread on the Fedora "legal" mailing list is here:
https://lists.fedoraproject.org/archives/list/legal%40lists.fedoraproject.org/thread/LSM6MO6TAHTIDNF5COCA6UWQDHWRF3AH/

YeungOnion:
Two things:

1. Do you have alternatives to suggest? Would there be any issues downloading it and using it in our CI so that we're not distributing it?

2. Also, noticed that Apache Commons ([link to folder](https://github.com/apache/commons-math/tree/master/commons-math-legacy/src/test/resources/org/apache/commons/math4/legacy/stat/data)) is also using this dataset, didn't see anything regarding licensure in [the commit history](https://github.com/apache/commons-math/tree/master/commons-math-legacy/src/test/resources/org/apache/commons/math4/legacy/stat/data). Would you happen to know enough about licenses to know if I should let them know?

decathorpe:
> Two things:
> 
> 1. Do you have alternatives to suggest? Would there be any issues downloading it and using it in our CI so that we're not distributing it?

I think this would be a good solution, yes.

> 2. Also, noticed that Apache Commons ([link to folder](https://github.com/apache/commons-math/tree/master/commons-math-legacy/src/test/resources/org/apache/commons/math4/legacy/stat/data)) is also using this dataset, didn't see anything regarding licensure in [the commit history](https://github.com/apache/commons-math/tree/master/commons-math-legacy/src/test/resources/org/apache/commons/math4/legacy/stat/data). Would you happen to know enough about licenses to know if I should let them know?

Based on the information I have (the legal review in the mailing list thread I linked above), yes, that would probably be a good idea.

YeungOnion:
#214 dropped this from the crate, so the 0.17.0 release will not distribute the data from crates.io

[...]

PR #214, opened 2024-04-16
Use `include` in Cargo.toml to remove some files from crates.io package

FreezyLemon:
`cargo package --list` shows the contents that would be included in a crate file (e.g. when running `cargo publish`), and by diffing master vs PR it shows that the following files are no longer included in the package:

```
.github/workflows/test.yml
.gitignore
benches/order_statistics.rs
codecov.yml
data/nist/lew.txt
data/nist/lottery.txt
data/nist/mavro.txt
data/nist/michaelso.txt
data/nist/numacc1.txt
data/nist/numacc2.txt
data/nist/numacc3.txt
data/nist/numacc4.txt
rustfmt.toml
```

I think all of these except maybe the benchmark are pretty obvious to remove from the distributable package. Even if the NIST data is removed from the repo entirely (see #195), I'd argue it makes sense to use an allowlist for the distributable package instead of including everything by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,8 @@ homepage = "https://github.com/boxtown/statrs"
 repository = "https://github.com/boxtown/statrs"
 edition = "2018"
 
+include = ["CHANGELOG.md", "LICENSE.md", "src/"]
+
 [lib]
 name = "statrs"
 path = "src/lib.rs"


YeungOnion:
I had the same thought, since it addresses part of the license violation with NIST's data. Approved.

PR #216, opened 2024-04-17
Fix simple clippy warnings

FreezyLemon:
There are a few more clippy warnings, but those aren't as clear-cut as the ones I fixed in this PR.

### DIFF
--- a/src/distribution/beta.rs
+++ b/src/distribution/beta.rs
@@ -3,7 +3,6 @@ use crate::function::{beta, gamma};
 use crate::is_zero;
 use crate::statistics::*;
 use crate::{Result, StatsError};
-use core::f64::INFINITY as INF;
 use rand::Rng;
 
 /// Implements the [Beta](https://en.wikipedia.org/wiki/Beta_distribution)
@@ -326,13 +325,13 @@ impl Continuous<f64, f64> for Beta {
             0.0
         } else if self.shape_a.is_infinite() {
             if ulps_eq!(x, 1.0) {
-                INF
+                f64::INFINITY
             } else {
                 0.0
             }
         } else if self.shape_b.is_infinite() {
             if is_zero(x) {
-                INF
+                f64::INFINITY
             } else {
                 0.0
             }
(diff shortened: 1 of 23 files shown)


YeungOnion:
Thanks!

Any thoughts on how to approach the one in `uniform`? I don't think uniform distribution is very well defined over an infinite (or semi-infinite) interval, but I'm not a formal mathematician.

If you agree, I'll ensure boundedness in `new` and drop the additional checks. 

After all, calls to our `sample` would fail in [`UniformFloat::new`](https://docs.rs/rand/latest/src/rand/distributions/uniform.rs.html#832) in the `rand` crate and there's not a way to define `inverse_cdf`.

FreezyLemon:
> Any thoughts on how to approach the one in uniform?

I think the logic itself is fine, clippy just doesn't like that there are two `else if` branches because that could be done in one:

```rs
} else if x >= self.max || x.is_infinite() && self.max.is_infinite() {
```

That's less readable though, so I would just keep it as-is and add an `#[allow]` attribute.

YeungOnion:
Regarding the logic, is it not redundant?

If the first case is false, then `self.min > NEG_INF` and `x > NEG_INF`
If the third is true, then `x = INF` and `self.max` is either `NEG_INF` (implies `x>`) or `INF` (implies `x=`)

Both options would have satisfied the second, no? 

---
Can you share what you think that use case might be? I'll take a chance to expand my brain.

FreezyLemon:
> Regarding the logic, is it not redundant?
> 
> If the first case is false, then `self.min > NEG_INF` and `x > NEG_INF` If the third is true, then `x = INF` and `self.max` is either `NEG_INF` (implies `x>`) or `INF` (implies `x=`)
> 
> Both options would have satisfied the second, no?

You're right. The other cases already neatly handle bounds being +/- infinity. I only skimmed the code because clippy is complaining about something else, but the entire block can be removed AFAICT.

Makes me wonder if the `else if self.max.is_infinite()` below that is also redundant, but I can't be bothered to do the logic in my head right now :P

> 
> Can you share what you think that use case might be? I'll take a chance to expand my brain.

Well, my original argument was going to be about API flexibility and that even if it might not be mathematically defined, it could still be useful to "just handle it" transparently instead of erroring. That said, it does look like the uniform distribution requires finite bounds from what I can find online.

YeungOnion:
Okay, I'll fix that separately since it's a behavior change instead of linting.

FreezyLemon:
Added some more trivial clippy fixes, some from nightly.

YeungOnion:
Think this is all of them that we can handle as lints then, right? Merging, then I'll put in the change for Uniform.

Not sure how I want to handle that one in Empirical. It does seem that `ordered_float`'s `NotNan<T>` has settled since `Empirical` was introduced; perhaps we can let clippy allow it for merging CI.

PR #218, opened 2024-04-21
Ensure Uniform constructed on bounded interval

YeungOnion:
In looking at lints (#216, has some of the discussion there) adjacently found that Uniform would allow infinite support. I don't know enough to say if this is well defined mathematically, but a few implications I thought of:
- `sample` will fail while using `rand`'s UniformSampler
- `cdf` and `sf` have odd end behaviors for the unbounded end, is F(x = INF) = 1 if X ~ Uni(a, INF)? Equals INF isn't good math, but is valid value for floats

There may be others, but this PR would disallow construction of such a distribution, `sf` was implemented to have behavior for infinite support, but `cdf` was not. If someone can declare how it should work, then we can reallow it and implement all the behavior for it.

### DIFF
--- a/src/distribution/uniform.rs
+++ b/src/distribution/uniform.rs
@@ -31,7 +31,7 @@ impl Uniform {
     ///
     /// # Errors
     ///
-    /// Returns an error if `min` or `max` are `NaN`
+    /// Returns an error if `min` or `max` are `NaN` or unbounded
     ///
     /// # Examples
     ///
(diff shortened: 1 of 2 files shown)


YeungOnion:
I removed tests that didn't require that `a, b < ∞` AND `a<b`  but I noticed there were tests that specifically tested such behavior, most notably things like
- $f_U(x;a=b) \equiv f_δ(x-a)$ δ being the dirac delta
- testing cdf at x=∞ for uniform without upper bound.

The blame shows @boxtown writing some of those tests 8 years ago. Don't suppose you recall why you chose that before, do you? I'd like to know more about what I'm considering removing.

YeungOnion:
More discussion is needed on what's expected behavior for such degenerate cases, #102 will host future discussion, for now I'll exclude this degenerate case.

PR #219, opened 2024-04-22
Make `rustfmt` happy

FreezyLemon:
Most of this should be uncontroversial.

[Rustfmt page](https://rust-lang.github.io/rustfmt/?version=v1.6.0) (rules can be looked up here).

I've seen `fn_params_layout = "Compressed"` in other projects, but all the code in statrs already adhered to the other rule (`"Tall"`), so I don't see why this should be `"Compressed"`. Plus it only affects a few functions, so this wouldn't save many lines anyways.

All the other rules that I removed seem unconventional to me, and again, all code in statrs at the moment adheres to these rules being `false` or not set. So I don't see a reason to keep these.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -12,9 +12,6 @@ use_small_heuristics = "Default"
 # TODO: single line functions only where short, please?
 # https://github.com/rust-lang/rustfmt/issues/3358
 fn_single_line = false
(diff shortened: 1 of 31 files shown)


YeungOnion:
Don't expect this to break most other PR's, but if it does, should all be easy changes.

Issue #185, opened 2023-02-27
Inverse CDF for binomial distribution doesn't seem to converge


derekdreery:
It currently falls back to what looks like interval bisection, but it then doesn't seem to converge, which is strange since the inverse cdf is monotonic IIUC.

troublescooter:
Can you provide parameters that do not converge?

derekdreery:
Hi sorry yes I should have given more information. The one I'm using is `Binomial::new(0.018, 465).inverse_cdf(0.0003472)`. In the function itself, the lower bound settles on `0` while the upper bound settles on `1`.

derekdreery:
According to `R`, the correct answer is `1`.

ImJeremyHe:
Don't converge here
```rs
#[cfg(test)]
mod tests {
    use statrs::distribution::{Binomial, DiscreteCDF};

    #[ignore]
    #[test]
    fn binomial() {
        let result = Binomial::new(0.5, 6).unwrap().inverse_cdf(0.75);
        assert_eq!(result, 6);
    }
}
```

YeungOnion:
Took a look and submitted a PR. Thanks all, please provide me feedback there.

ImJeremyHe:
@YeungOnion Really thanks for your work! Would you mind releasing a new version?

YeungOnion:
Sure, a fix release would be a good way for me to make the minor release smooth too. Source is now on branch `fix-issue-185` but I'm having issues publishing.

---

@statrs-dev/owners do any of you have publish access to crates.io or can here help? I get a 403 error when I run `cargo publish`
```text
~/p/statrs @8b1439e 25.5s ❱ cargo publish
    Updating crates.io index
   Packaging statrs v0.16.1 (/home/orion/projects/statrs)
   ... # truncated
   Compiling statrs v0.16.1 (/home/orion/projects/statrs/target/package/statrs-0.16.1)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 26.21s
    Packaged 63 files, 686.0KiB (133.6KiB compressed)
   Uploading statrs v0.16.1 (/home/orion/projects/statrs)
error: failed to publish to registry at https://crates.io

Caused by:
  the remote server responded with an error (status 403 Forbidden): this crate exists but you don't seem to be an owner. If you believe this is a mistake, perhaps you need to accept an invitation to be an owner before publishing.
```

[...]

PR #220, opened 2024-04-27
Fix for issue #185

YeungOnion:
fix: default inverse_cdf method incorrectly implements generic integer bisection
also adds additional tests


YeungOnion:
fixes #185

PR #222, opened 2024-04-28
Make PartialOrd implementation canonical for `NonNAN<T>`

FreezyLemon:
Fixes clippy error:

```rs
warning: non-canonical implementation of `partial_cmp` on an `Ord` type
  --> src/distribution/empirical.rs:14:1
   |
14 | /  impl<T: PartialOrd> PartialOrd for NonNAN<T> {
15 | |      fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
   | | _____________________________________________________________-
16 | ||         self.0.partial_cmp(&other.0)
17 | ||     }
   | ||_____- help: change this to: `{ Some(self.cmp(other)) }`
18 | |  }
   | |__^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#non_canonical_partial_ord_impl
   = note: `#[warn(clippy::non_canonical_partial_ord_impl)]` on by default
```

This is the normal way of implementing `PartialOrd` when a manual `Ord` implementation exists because it ensures that they will return the same result.


YeungOnion:
Might be better to derive `PartialOrd` instead, but the behavior change is good by me, it should not be constructed and it's a lot lighter than `ordered_float::NotNan`.

Since you're messing with it, would you rename to `NonNan` or `NotNan` to match PascalCase? It's not public, so this is just preference.

PR #227, opened 2024-05-04
Fix inverse_cdf for Gamma

YeungOnion:
Don't think there's a closed form expression for inverting the CDF, so this improves on the bisection approach with a few iterations of Newton-Raphson since we have `pdf` 

also adds tests

### DIFF
--- a/src/distribution/gamma.rs
+++ b/src/distribution/gamma.rs
@@ -609,6 +609,25 @@ mod tests {
     }
 
     #[test]
+    fn test_cdf_inverse_identity() {
+        let f = |p: f64| move |g: Gamma| g.cdf(g.inverse_cdf(p));
+        let params = [
+            (1.0, 0.1),
+            (1.0, 1.0),
+            (10.0, 10.0),
+            (10.0, 1.0),
+            (100.0, 200.0),
+        ];
+
+        for param in params {
+            for n in -5..0 {
+                let p = 10.0f64.powi(n);
+                test_case(param, p, f(p));
+            }
+        }
+    }
+
+    #[test]
     fn test_sf() {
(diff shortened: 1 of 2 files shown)


src/distribution/gamma.rs
```diff
@@ -146,6 +147,51 @@ impl ContinuousCDF<f64, f64> for Gamma {
             gamma::gamma_ur(self.shape, x * self.rate)
         }
     }
+
+    fn inverse_cdf(&self, p: f64) -> f64 {
+        const MAX_ITERS: (u16, u16) = (8, 4);
```
FreezyLemon, line 152: Is there a special reason for defining the number of iterations like this? It's probably more natural to just
```rs
for _ in 0..8 {
```

since the complexity is pretty low

YeungOnion: I thought,
> 8 and 4 sound like nice numbers, what does that give me. 8 bisection is 2 digits in $p$ and NR is quadratic convergence, so 4 gives 16 digits, lucky me. Hmm, it's helpful to name things, so `const` it is. 

Then I thought,
> I don't want to name two close but separate things, tuple it is. 

I.e. I would be fine changing it to literal.

However, since my attention is on it, I think I might instead do the initial bracketing a little differently and try some math to back up the number of iterations I select.

FreezyLemon: I see. It's not a problem either way, I just wanted to know why you used that const tuple there.

YeungOnion:
Since it works and I've not already reviewed how to make it "better" I don't think I will for now.

If we get a handle on provided or emitting precisions then I think property testing shrinking could use some of that precision information to help find larger parameter space with correct behavior.

PR #228, opened 2024-05-06
Add a way to easily get the standard normal distribution.

RationalAsh:
I've been using statrs and I love the project. I find myself creating the standard normal distribution frequently. So I thought it might be convenient to add a way to get the standard normal in one function call rather than unwrapping. Hope this is useful for others too.

### DIFF
--- a/src/distribution/normal.rs
+++ b/src/distribution/normal.rs
@@ -51,6 +51,26 @@ impl Normal {
             Ok(Normal { mean, std_dev })
         }
     }
+
+    ///  Constructs a new standard normal distribution with a mean of 0
+    /// and a standard deviation of 1.
+    ///
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use statrs::distribution::Normal;
+    ///
+    /// let mut result = Normal::standard();
+    /// ```
+    pub fn new() -> Normal {
+        let mean: f64 = 0.0;
+        let std_dev: f64 = 1.0;
+        Normal {
+            mean,
+            std_dev
+        }
+    }
 }
 
 impl ::rand::distributions::Distribution<f64> for Normal {
--- a/src/distribution/normal.rs
+++ b/src/distribution/normal.rs
@@ -63,7 +63,7 @@ impl Normal {
     ///
     /// let mut result = Normal::standard();
     /// ```
-    pub fn new() -> Normal {
+    pub fn standard() -> Normal {
         let mean: f64 = 0.0;
         let std_dev: f64 = 1.0;
         Normal {
--- a/src/distribution/normal.rs
+++ b/src/distribution/normal.rs
@@ -52,7 +52,7 @@ impl Normal {
         }
     }
 
-    ///  Constructs a new standard normal distribution with a mean of 0
+    /// Constructs a new standard normal distribution with a mean of 0
     /// and a standard deviation of 1.
     ///
     ///
@@ -63,12 +63,10 @@ impl Normal {
     ///
     /// let mut result = Normal::standard();
     /// ```
-    pub fn standard() -> Normal {
-        let mean: f64 = 0.0;
-        let std_dev: f64 = 1.0;
+    fn standard() -> Normal {
         Normal {
-            mean,
-            std_dev
+            mean: 0.0,
+            std_dev: 1.0,
         }
     }
 }
@@ -306,6 +304,15 @@ pub fn ln_pdf_unchecked(x: f64, mean: f64, std_dev: f64) -> f64 {
 /// draws a sample from a normal distribution using the Box-Muller algorithm
 pub fn sample_unchecked<R: Rng + ?Sized>(rng: &mut R, mean: f64, std_dev: f64) -> f64 {
     mean + std_dev * ziggurat::sample_std_normal(rng)
+}
+
+
+impl std::default::Default for Normal {
+    /// Returns the standard normal distribution with a mean of 0
+    /// and a standard deviation of 1.
+    fn default() -> Self {
+        Self::standard()
+    }
 }
 
 #[rustfmt::skip]
--- a/src/distribution/normal.rs
+++ b/src/distribution/normal.rs
@@ -540,8 +540,8 @@ mod tests {
     fn test_default() {
         let n = Normal::default();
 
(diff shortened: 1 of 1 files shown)


YeungOnion:
I've wondered if this kind of thing would be useful and I thought to use Default, but I was unsure if it would be obvious for any other distributions, I thought perhaps the exponential and uniform would be guessable. The motive for default if useful for other distributions would be that the name is reusable.

RationalAsh:
My typical experience of working with these distributions is that many (most?) distributions don't really do "standard" as much as the Normal Distribution. So it would be a property that is unique to the normal distribution. Personally, I would add it as something that is specific only to the normal distribution. 

But I do agree with you that the Default trait is more elegant as a solution. So if that's what we should go with, then I can open an updated PR.

FreezyLemon:
I would argue that `Normal::standard()` is more obvious at first glance than `Normal::default()`. That said, you can also just implement both and have `Normal::default()` return the result of `Normal::standard()`.

src/distribution/normal.rs
```diff
@@ -51,6 +51,26 @@ impl Normal {
             Ok(Normal { mean, std_dev })
         }
     }
+
+    ///  Constructs a new standard normal distribution with a mean of 0
```
FreezyLemon, line 55:
```suggestion
    /// Constructs a new standard normal distribution with a mean of 0
```

src/distribution/normal.rs
```diff
@@ -51,6 +51,26 @@ impl Normal {
...
+    ///
+    /// let mut result = Normal::standard();
+    /// ```
+    pub fn standard() -> Normal {
+        let mean: f64 = 0.0;
+        let std_dev: f64 = 1.0;
+        Normal {
+            mean,
+            std_dev
+        }
```
FreezyLemon, line 72:
```suggestion
        Normal {
            mean: 0.0,
            std_dev: 1.0,
        }
```

FreezyLemon (commented):
I added two formatting nitpicks. CI does not enforce formatting rules yet, but it's planned (see #215) and following them now will avoid having to fix them later

RationalAsh:
> I added two formatting nitpicks. CI does not enforce formatting rules yet, but it's planned (see #215) and following them now will avoid having to fix them later

Thanks! I've added in these changes in addition to implementing the default trait for the Normal distribution. Also added a test case. for it. Seems OK. Let me know if any further changes are needed!

src/distribution/normal.rs
```diff
@@ -508,4 +535,17 @@ mod tests {
         test_almost(5.0, 2.0, 10.0, 1e-14, inverse_cdf(0.9937903346742238648330218954258077788721022530769078));
         test_case(5.0, 2.0, f64::INFINITY, inverse_cdf(1.0));
     }
+
+    #[test]
+    fn test_default() {
+        let n = Normal::default();
+
+        n_mean = n.mean().unwrap();
+        n_std  = n.std_dev().unwrap();
```
FreezyLemon, line 544: These two lines need a `let`:

```suggestion
        let n_mean = n.mean().unwrap();
        let n_std  = n.std_dev().unwrap();
```

YeungOnion:
Whoops, didn't realize I left this as a draft. Cool with keeping `impl Default`.

---

I wholeheartedly agree that `Normal::standard()` likely only means one thing.

Now that I've considered more on what I suggested, it may only add value if there's a use case for needing `Default` as a trait bound.

src/distribution/normal.rs
RationalAsh: Oops not sure how I missed this one (cargo test ran??). Fixed.

FreezyLemon: If you scroll up, you'll see that the entire tests module is only run when the `nightly` feature flag is enabled. So you'd need to run `cargo +nightly test -F nightly`, which is easy to miss

YeungOnion:
Thanks for reviewing @FreezyLemon!

YeungOnion:
@RationalAsh Since you didn't notice the testing issue, do you have any feedback on making that more apparent that you'll need to run on nightly compiler with `-Fnightly`?

We have it in the `Contributing` section of the README, but it's not obvious if something you add isn't run unless you pay close attention to the number of tests and how it changes as you add them and I'm wondering if you've got suggestions?

Depending on workflow and context of edits, I'll target unit tests in specific module passing the module to `libtest` like so
```sh
cargo +nightly test -F nightly -- distribution::normal # tests in the module you contributed to in this PR
```

RationalAsh:
> @RationalAsh Since you didn't notice the testing issue, do you have any feedback on making that more apparent that you'll need to run on nightly compiler with `-Fnightly`?
> 
> We have it in the `Contributing` section of the README, but it's not obvious if something you add isn't run unless you pay close attention to the number of tests and how it changes as you add them and I'm wondering if you've got suggestions?
> 
> Depending on workflow and context of edits, I'll target unit tests in specific module passing the module to `libtest` like so
> 
> ```shell
> cargo +nightly test -F nightly -- distribution::normal # tests in the module you contributed to in this PR
> ```

The way I typically do it for my own projects that need non-standard testing / compile commands is to have a test.sh script that contains the commands needed to run all tests locally.

Also, this might be a really dumb question because I didn't go to deep into rust / statrs internals  -- Is there a need for the tests to run using the nightly compiler? Or in other words - is it possible to refactor so that most tests do not require the nightly compiler to run and instead use the latest stable version?

YeungOnion:
> Also, this might be a really dumb question because I didn't go to deep into rust / statrs internals -- Is there a need for the tests to run using the nightly compiler? Or in other words - is it possible to refactor so that most tests do not require the nightly compiler to run and instead use the latest stable version?

Was hoping this exact question would be asked without my mentioning it. I don't see a _need_ for it beyond how the test suite is written, but we'd have to change tests that use the `testing_boiler` macro or find a way to write the macro differently. Personally, I don't like the way `testing_boiler` is used as failed assertions don't clearly provide the line where the test is written.

I'll make an issue for it so we can merge and close this PR.

_Aside_: I think the possibility of a `nightly` _feature_ is useful for things we're figuring out. I'm expecting the multivariate API will start as `nightly` for the crate, but not `rustc`

Issue #147, opened 2021-07-06
Restore codecov integration


vks:
This was broken when we moved our CI from Travis to GitHub actions. We should consider restoring it.

PR #229, opened 2024-05-15
feat: reintroduce coverage and upload to codecov.io

YeungOnion:
Bringing back codecov #147 

I went with tarpaulin because it seemed easy to setup and use, especially with the container on dockerhub. I don't believe that coverage reports need the same level of confidence as passing the tests that are written, but can be useful in deciding how to focus efforts.

But I'm new to this, so feedback would be appreciated.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,38 @@
+name: Coverage
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+jobs:
+  coverage:
+    name: Coverage
+    runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: -D warnings
+      CARGO_TERM_COLOR: always
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: llvm-tools-preview
+
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: nextest
+      - uses: taiki-e/install-action@v2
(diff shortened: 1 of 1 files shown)


YeungOnion:
It appears that tarpaulin is accurately reading tests that rely on the `testing_boiler` macro, I don't know where else to check for inaccuracy, so others' input would be helpful.

FreezyLemon (commented):
I don't have any experience with tarpaulin, but it seems straightforward enough. Some actions are outdated though.

YeungOnion:
Should we get tests back to >90% before the release or not fail ci to get the release out?

_Note_: since v14 we've been below 90%, thinking maybe changing it to 80%?

FreezyLemon:
BTW: Looks like moving from nightly to stable caused a permissions error. Could be xd009642/tarpaulin#406 and may be solved by moving to a non-docker solution. FYI [`taiki-e/install-action`](https://github.com/taiki-e/install-action) supports tarpaulin, so that could be a good alternative.

YeungOnion:
_Note_: Still looking into this option, but I force-pushed to test the action.

In looking into [taiki-e/install-action](https://github.com/taiki-e/install-action) I found their [`nextest`](https://nexte.st/) project, which appears more friendly and designed for CI.

It and the wrapper they developed to compiling with instrument-coverage, [`cargo-llvm-cov`](https://crates.io/crates/cargo-llvm-cov) it seems capable of removing the need to rerun tests for collecting coverage data.

YeungOnion:
I've let this sit awhile, let's try it out.

PR #230, opened 2024-05-24
Handle tests with NIST data as integration tests

YeungOnion:
Unsure how best to handle these tests, but I think separating them is a good start to not have the data in the repo per #195. 

Test suite will rely on environment variable, `"STATRS_NIST_DATA_DIR"` set as `const NIST_DATA_DIR_ENV`, to specify path where the files, some listed below, are located.
- lew.txt
- mavro.txt
- numacc1.txt
- ...

_Note_: I am not married to the usage of an environment variable or its name.


FreezyLemon:
I like the approach. Maybe add a helper script (bash?) to download the NIST files? Downloading the files manually could be a bit tedious.

YeungOnion:
We didn't distribute the integration tests with the last release, but if we did, would we want to share the script along with these tests?

FreezyLemon:
I would logically see the script as a part of that test suite, so yeah.

YeungOnion:
Will add to CI in another PR, any feedback on this one?

FreezyLemon:
I tried it on my machine and the script looks good. I think the environment variable works well, no reason to overthink this for a test IMO.

Few ideas to make the experience a bit smoother:
- Add `tests/*.dat` to the `.gitignore` so the downloaded files don't show up in git
- Add a bit of output to the script (e.g. `Downloading <file>...`, `Processing <file>...`) so we know what's going on. This gets more important when `.gitignore` includes the data files
- Short comment (top of the `nist_tests.rs`) that explains how you're intended to actually run the tests would be nice. `cargo test -- --include-ignored nist` is not obvious IMHO
- Use the current working dir when `STATRS_NIST_DATA_DIR` is unset. That'd make it easier to `cd tests; ./gather_nist_data.sh; cargo test` without requiring the env variable

YeungOnion:
Thanks for the feedback, think it will be clearer and nicer. Opted for default target directory to be `tests` since manifest and ignore target it and set some sense of working directory.

Issue #233, opened 2024-05-25
Change compile requirements for running `cargo test`


YeungOnion:
Do we _need_ the nightly compiler just for the test suite? The requirement is for feature `fn_traits` (I believe this requires  feature `unboxed_closures`) is from `statrs::testing_boiler` but nowhere else.

Feedback seems to be,
- non-obvious to test with nightly for new contributors (even when familiar with rust projects) 
- assertion error messages reference call to macro `statrs::testing_boiler`, not clear what line test is defined on.

Open to other feedback, but what I'm imagining is a few proc macros that expand to unwraps or assertions
- `create(dist_family_name, tuple_of_dist_parameters, ...error msg)`
- `bad_create(...` sim to previous asserting is_err
- `test_almost(dist_family_name, tuple_of_dist_parameters, dot_method_call_with_args, ...error_msg)`

FreezyLemon:
Related: #232 

I'm pretty sure the `testing_boiler` macro can be changed so that it doesn't need any nightly compiler features. 

I wonder about the second part. Does `RUST_BACKTRACE=1` not give any useful information? How would an optional error message, e.g. in `test_almost(/* .. */, "error message here")`, help with [test suites like these](https://github.com/statrs-dev/statrs/blob/358aafb58bd25038a2379aefa3249ecb62037228/src/distribution/cauchy.rs#L360-L392)? There must be a better way than adding error messages "line 1", "line 2", ..., right?

YeungOnion:
`RUST_BACKTRACE=1` does provide useful info, perhaps I'm a little silly. I think the messages would be convenient, but I think changing what we have for existing would be better.

YeungOnion:
resolved by #234

PR #234, opened 2024-05-27
Update testing_boiler macro to not need a nightly compiler

FreezyLemon:
Addresses part of #233.

This makes the macro a bit more complicated but removes the need for the `fn_traits` feature.

Instead of 
```rs
testing_boiler!((f64, f64), Beta);

#[test]
fn some_test() {
    let arg = (1.0, 2.0);
    try_create(arg);
}
```

it's now

```rs
testing_boiler!(a: f64, b: f64; Beta);

#[test]
fn some_test() {
    let arg_a = 1.0;
    let arg_b = 2.0;
    try_create(arg_a, arg_b);
}
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,20 +41,12 @@ jobs:
     strategy:
       matrix:
           os: [ubuntu-latest, macos-latest, windows-latest]
-          toolchain: [stable, nightly]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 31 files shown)


YeungOnion:
haha, same approach I took, [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=27d9ce8336d3ae5bddaa84375e72e7c7)
Thanks for this, that was quick.

PR #235, opened 2024-05-30
release 0.17.0

YeungOnion:
I think this is okay for getting some more users back.

Open to feedback on the changelog, I used `release-plz` to generate the log for this release and made edits for further format

Remaining active work that is related to CI will be updated and likely minor version releases to [crates.io], included but not limited to:
- code coverage reports. Note that I dropped the codecov badge on the readme for this release since it's inaccurate.
- testing NIST data

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,49 @@
-Unreleased
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [0.17.0](https://github.com/statrs-dev/statrs/compare/v0.16.0...v0.17.0) - 2024-05-30
+
+### Added
+- specializes `inverse_cdf()` for Uniform (#166)
+- Add way to get standard normal distribution easily. (#228)
+- reject constructing Uniform of infinite support (#218)
+- extend `StatsError` for finiteness (#218)
+- default implementation of survival function with generics (#179)
+- update `MultivariateNormal` API
+  - construct from nalgebra with `MultivariateNormal::new_from_nalgebra` (#177)
+  - support `std::vec` vector input in addition to `nalgebra` vectors (#199)
+
+### Fixed
+- Update nalgebra to 0.32 (#187)
+- for Gamma with shape<1 there is no mode, returns `None` instead of some negative number (#212)
+- fix precision of <Gamma as ContinuousCDF>::inverse_cdf with some newton raphson steps (#227)
+  - adds test case from #200
+- fix integer bisection for default implementation of `<D as DiscreteCDF>::inverse_cdf` (#220)
+  - also add tests from (#185)
+
+### Other
+- Remove "nightly" feature and drop testing requirement for `nightly` (#234)
+- Allow some imprecision in specific test case (#215)
+- Update CI (#215)
+  - Check formatting in CI via rustfmt
+  - Expand CI test job
+  - Add clippy job to CI
+- update README with formatting and adding to "Contributing" (#213)
+- Add test asserting that `StatsError` is Sync & Send (#226)
+- Rename private struct NonNAN<T> to NonNan<T> (#222)
+- Remove `lazy-static` dependency and make FCACHE a proper const (#211)
+- crate examples shall be in docstrings instead of README (#213)
+- alias `inverse_cdf` as "quantile function" in docs (#213)
+- docstrings with math shall be `text` instead of `ignore` (#213)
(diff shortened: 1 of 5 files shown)


YeungOnion:
Noting that at some point (#209) @FreezyLemon mentioned the idea of introducing MSRV to this project; I planned to delay adding that policy to the release that introduces that change that motivated it since this release is focused to get the project back on its feet.

YeungOnion:
I got some advice that if we're going to introduce the msrv may as well not wait on it. 

I did some testing with [`cargo-hack`](https://crates.io/crates/cargo-hack) and found that `cargo check` fails up through 1.60 on nalgebra. Note that 1.61 will also suffice for #209's proposed changes regarding.

I'll make these changes now.

FreezyLemon:
Any MSRV should *really* be checked in CI, since it's very easy to accidentally break it while updating dependencies. It would also be nice to document it somewhere, e.g. at the bottom of the README (including when and for what reasons it will be changed, and how the crate version will reflect MSRV changes)

YeungOnion:
Seems like I only checked the lib in my local testing, our benches rely on `criterion` relying on `rayon` which has `rust-version = 1.64`

I'm okay excluding those benchmarks from the package, but I've not found a way to split the dependencies for benchmarking apart from testing, unsure if this is possible.
@FreezyLemon Do you know a way around this other than making a separate crate for testing?

FreezyLemon:
> Do you know a way around this other than making a separate crate for testing?

Another way of achieving this would be a crate feature that needs to be enabled for benchmarking, but it's not great for developer ergonomics.

It's also possible to just disable the default features for criterion which should remove rayon and possibly reduces the MSRV.

YeungOnion:
I can get it down to 1.65 with changing the features criterion uses and keeping the version pinned, which aligns around the 0.16 statrs release. However, my opposition is that the MSRV is higher than what the statrs lib itself needs, so I find it a little misguiding. I would clearly mention this in the docs. 

I don't like the option to gate benches with a feature, but given the below reasons, I think we could get by.
- the features benched aren't actively being developed it may not impede any developer
- benchmarks should be reported with version compiled with
- an MSRV higher than what our lib needs could remove it from consideration, only a few minor versions (5 months) newer.

Thoughts?

YeungOnion:
made a [post on the forums](https://users.rust-lang.org/t/rust-version-and-development-dependencies/112608) to get others' input if this would be unexpected.

.github/workflows/msrv.yml
```diff
@@ -0,0 +1,20 @@
...
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: "-Dwarnings"
+
+jobs:
+  msrv:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: taiki-e/install-action@cargo-hack
+    - run: cargo hack check --rust-version --workspace --all-targets --ignore-private && diff Cargo.lock Cargo.lock.msrv
```
FreezyLemon: I feel like this is going to fail quickly. If any dependency releases a new crate version that is MSRV-compatible, the `Cargo.lock` created here will differ from the checked-in `Cargo.lock.msrv`. Or am I missing something here?

.github/workflows/msrv.yml
```diff
@@ -16,5 +16,4 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: taiki-e/install-action@cargo-hack
-    - run: cargo hack check --rust-version --workspace --all-targets --ignore-private && diff Cargo.lock Cargo.lock.msrv
-
+    - run: cp Cargo.lock.msrv Cargo.lock && cargo hack check --rust-version --workspace --all-targets --ignore-private && diff Cargo.lock.msrv Cargo.lock
```
FreezyLemon, line 19: Why does this `diff Cargo.lock.msrv Cargo.lock` at the end? Isn't this going to break as soon as any dependency releases an MSRV-compatible update?

(also this should use --locked)

FreezyLemon:
I honestly thought that disabling the criterion default features would solve the problem. The benches feature doesn't feel great... I'd honestly just set the expectation that the MSRV is meant for crate users (`cargo add statrs`) and not for crate devs (`git clone <statrs>`) and just be done with it. `dev-dependencies` are not something to consider for normal crate usage, and any dev should have a recent Rust version anyways.

YeungOnion:
I agree, and it makes me wonder if it's better to
- hold off on an MSRV
- disable/remove benches to make sure that CI is checking behavior that a lib user would see
- have CI only check MSRV only for the lib target, not its tests (or benchmarks)

I'm leaning toward the last option, but it's potentially hiding a source of inconsistency in test results.

YeungOnion:
Think I'll hold off on it. Since inclusion of it is only policy, I imagine it would only change the fix release version.

Issue #198, opened 2024-01-19
Alternative parameterization for negative binomial distributions?


diaorch:
Currently, the `struct NegativeBinomial` uses the parameters `r` and `p`, as in the interpretation of negative binomial distribution as the distribution of the number of failures in a sequence of Bernoulli trials with probability of success `p` that continue until `r` successes occur. 

Another way to describe a negative binomial distribution is through the mean and the size/dispersion parameters. This is worth implementing especially given that a lot of the statistical questions center around the mean of the distribution. 

`r` and `p` can be calculated from the mean and the dispersion parameter. 

Perhaps the alternative parameterization of negative binomial distributions can be a useful addition to the crate's functionality?

YeungOnion:
I expect that you'd want this at initialization, but I don't have the background to know when else a different parametrization would be useful. Can you help me understand such a scenario?

diaorch:
Thank you for getting back to me!

Here's a quick summary of what parameterizations we are discussing: the current parameterization of negative binomial distribution as implemented uses `r` and `p`, which have specific meanings in terms of Bernoulli trials. The alternative parameterization I am interested in uses mean and dispersion as parameters, which is described in more detail in the last two paragraphs of ["Alternative formulation" of negative binomial distribution on Wikipedia page](https://en.wikipedia.org/wiki/Negative_binomial_distribution#Alternative_formulations). Also, I appreciate the warning in the current documentation to note carefully the meaning of the parameters.

The mean-and-dispersion is more widely used in regression analysis, because the explanatory variables can be linked to the mean, similar to linear regression. This is also, in my opinion, more intuitive when it comes to interpreting the effects of explanatory variables on the negative binomial counts. (Personally, I don't think I have seen negative binomial regression done with the `r`-and-`p` parameterization.) In my work, having the mean in parameterization also helps me compare a negative binomial distribution with a Poisson distribution of the same mean, since I can more quickly evaluate the dispersion level in the negative binomial distribution. I would also imagine that the mean-and-dispersion parameterization is more useful outside of a frequentist statistics context.

[...]

PR #247, opened 2024-06-23
hierarchy of errors

YeungOnion:
looking for some feedback on possible error hierarchy.
I don't think I did the source error pattern nicely, but I think there's something here worth looking at.


src/distribution/gamma.rs
```diff
@@ -1,9 +1,78 @@
...
+            DegenerateSuccessCount => Self::DegenerateRate(0.0),
+        }
+    }
+}
+
+/// holds a valid parametrization of the gamma distribution in shape and rate.
+pub struct Parameters {
+    shape: f64,
+    rate: f64,
+}
```
FreezyLemon, line 44: Can you explain why you added this abstraction? You could also check the parameters inside `Gamma::new` and return a suitable error from there, right?

FreezyLemon (commented):
This looks pretty nice. The error contains a lot of information (maybe too much? I can't estimate if this level of detail is needed) and the source annotation from `thiserror` should create better dev feedback for little effort.

There's one thing I don't really understand but I like this in general

src/distribution/gamma.rs
YeungOnion: Mostly just playing around with having input checking in another type to support the use of infallible `new` methods. 

But more interestingly perhaps it _could_ be a way to get alternative parametrizations #198 as well as `new_from_moments` to do something like,

```rs
let m = Gamma::new(shape, rate).into_moments();
let p = Poisson::new_from_moments(m);
```

But I've lots of thoughts floating around about that none motivated by something I've actually needed thus far. As a solution to #198 could be adding just one more method.

YeungOnion:
#265 will do this without unused complexity. Thanks @FreezyLemon

PR #251, opened 2024-07-20
Update uniform.rs

avhz:
Add a `standard()` method to Uniform Distribution, following the same idea as the `Normal::standard()` method. 

If you're happy with this change I can also add the same thing for other distributions that have an accepted 'standard' form.

### DIFF
--- a/src/distribution/uniform.rs
+++ b/src/distribution/uniform.rs
@@ -4,6 +4,7 @@ use crate::{Result, StatsError};
 use rand::distributions::Uniform as RandUniform;
 use rand::Rng;
 use std::f64;
+use std::fmt::Debug;
 
 /// Implements the [Continuous
 /// Uniform](https://en.wikipedia.org/wiki/Uniform_distribution_(continuous))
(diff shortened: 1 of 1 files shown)


YeungOnion:
In #228 we also implemented the `Default` trait since that would contribute to API, i.e. adds functionality without newtype, so I'd like this to also `impl Default` if we have a `standard` or `unit` construction method and if you added those trivial tests, I'd merge that.

After all, if most users had to write a `new_standard_normal()->Normal` and `new_standard_uniform()->Uniform` I don't think that would break their development experience. 

---

So, wanting to ensure `Default` behaves expectedly (since most distributions don't have a `T::new()->T`), I am certain Normal and Uniform are unambiguous in their choice, I am less certain for others.

For example, Exponential distribution is parametrized by a single scale parameter, but I've not heard someone use the term, "standard exponential distribution"/"unit exponential distribution" even if I'd infer it to be $\lambda=1$, some arguments could be made for other distributions as well. 

I suppose time and feedback will make this clear if other distributions have this, the author of #228 did not think so.

avhz:
I've included `Default` and some basic unit tests as requested :)

YeungOnion (approved):

YeungOnion:
Looks good. Thanks for the contribution; hope you continue to provide feedback!

PR #256, opened 2024-08-07
Replace `is_zero(x)` with `x == 0.0`

FreezyLemon:
```rs
fn is_zero(x: f64) -> bool {
    ulps_eq!(x, 0.0, max_ulps = 0)
}
```

[turns into](https://docs.rs/approx/latest/src/approx/macros.rs.html#61-68)

```rs
::approx::Ulps::default().max_ulps(0).eq(x, 0.0)
```

[which turns into](https://docs.rs/approx/latest/src/approx/ulps_eq.rs.html#34-71)

```rs
if (x - 0.0).abs() <= f64::EPSILON {
  return true;
}

if x.signum() != 0.0.signum() {
  return false;
}

x.to_bits() - 0.0.to_bits() == 0
```

The actual code is not quite as clear as this, it took me a few minutes to find the relevant places and shorten it to this.

`is_zero` [was introduced](https://github.com/statrs-dev/statrs/commit/b186804e9a61b2023d72ce0cb3a06feb9a06c5dc) in order to silence clippy warnings/errors and was a 1:1 replacement for `x == 0.0`. 

I would argue that `is_zero` is unnecessarily complicated, not transparent and possibly incorrect (one example: `0.0` == `-0.0` evaluates to `true` while `is_zero(-0.0)` will evaluate to `false`) and was not the right solution to the clippy warnings in the first place.

### DIFF
--- a/src/distribution/beta.rs
+++ b/src/distribution/beta.rs
@@ -1,6 +1,5 @@
 use crate::distribution::{Continuous, ContinuousCDF};
 use crate::function::{beta, gamma};
-use crate::is_zero;
 use crate::statistics::*;
 use crate::{Result, StatsError};
 use rand::Rng;
@@ -339,7 +338,7 @@ impl Continuous<f64, f64> for Beta {
                 0.0
             }
         } else if self.shape_b.is_infinite() {
-            if is_zero(x) {
+            if x == 0.0 {
                 f64::INFINITY
             } else {
                 0.0
@@ -377,7 +376,7 @@ impl Continuous<f64, f64> for Beta {
                 f64::NEG_INFINITY
             }
         } else if self.shape_b.is_infinite() {
-            if is_zero(x) {
+            if x == 0.0 {
                 f64::INFINITY
             } else {
                 f64::NEG_INFINITY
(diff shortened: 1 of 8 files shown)


YeungOnion:
The purpose it was added seems clear and no longer relevant. This means all cases of negative zero will likely be handled as if they were positive zero since we seem to only use
- inequality checking, either explicitly or via Range `contains` or
- `signum`

instead of checking sign bit.

Is adding some negative 0 test cases needed here? It would ensure that behavior that checks the sign bit is consistent. Or is there an alternative way to remember and enforce this decision against checking sign bit for zero values?

FreezyLemon:
Hm.. on revisit, I think I missed something fairly trivial, don't ask me how.

`is_zero(-0.0)` **does return true**, not false like I said above. I came to the conclusion while I was looking at the code and didn't test it afterwards. Sorry for that.

Anyways, 

> Is adding some negative 0 test cases needed here?

I think that my mistake above changes the context a bit. It's probably not necessary to test this...?

> Or is there an alternative way to remember and enforce this decision against checking sign bit for zero values?

How often does `statrs` use bitwise operators or otherwise check the contents of specific bits of numbers? AFAICT, it's usually just "normal" number operators/functions. I guess some dependencies do this (as seen in the `ulps_eq` macro), but it should probably be assumed these are correct, right?

PR #257, opened 2024-08-07
Remove unused imports

FreezyLemon:
I don't think there's a good reason to allow this in the entire crate. There weren't many unused imports to remove anyways, and if necessary it can always be re-allowed for specific files or modules.


YeungOnion:
seems like a reasonable chore

Issue #221, opened 2024-04-27
Consider improving error handling and `StatsError` type


FreezyLemon:
`enum StatsError` is supposed to be an

> Enumeration of possible errors thrown within the `statrs` library.

This indicates that it should not try to be a generic error type for any kind of statistics calculations, but instead only concern itself with the errors produced in `statrs`.

With that basic assumption, there are currently some inconsistencies and outdated practices though ([API guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#c-good-err) for reference):
- Add test that `StatsError` implements both `Sync` and `Send` (this is more of a formality, but is trivial to implement and good future-proofing)
- [`Error::description`](https://doc.rust-lang.org/std/error/trait.Error.html#method.description) is deprecated and should not be implemented
- There are a few unused variants (`ArgNotNegative`, `ArgIntervalExclMin`, etc.), these are leftovers from older versions and should be removed because they're no longer needed by `statrs`.
- There's at least one case of a function returning a `Result<T>` that cannot return an error: `Empirical::new`. There's no real reason for this infallible API to return a `Result<T, E>`. There might be others.

I realize that most of these are breaking changes, but seeing that the crate is pre-1.0, I don't think there's a big problem doing this.

Other things that could be improved:
`StatsError` is big: 40 bytes on a Linux x64 target. This is because there are variants which contain 2x `&str` (2 x 16 = 32 bytes plus discriminant and padding). Is it really necessary to have strings in the error type? The implementation could be replaced mostly 1:1 with some sort of `ArgName` enum, but there might be an even better solution that does not need this argument name wrapping at all.

All `new` functions seem to just return `StatsError::BadParams` whenever the params are outside of the defined/allowed range. Is there a good reason for these to be so vague when compared to the more specific errors returned by other functions? After all, the more specific errors already exist, why not return more exact error information? There might even be value in providing multiple error types, to have errors that are more specific to the exact API being used.

YeungOnion:
I do see good reason for all of these and I'd be open to making changes for all but the infallible `new` not returning a result. 

All other public structs defined in the `distribution` module have a `new` method implemented that returns Result so it does provide consistency. Perhaps if `Empirical` were in a different module or it's `new` had a different name?

FreezyLemon:
> All other public structs defined in the distribution module have a new method implemented that returns Result so it does provide consistency. Perhaps if Empirical were in a different module or it's new had a different name?

[...]

PR #258, opened 2024-08-07
Remove `StatsError` type & replace `Result<T, StatsError>` with `Option<T>`

FreezyLemon:
Mainly looking to move the error discussion forward by seeing how this would look like.

This is a possible solution to #221, and effectively the extreme opposite of #247.

Why do this? Well, a lot of functions look something like this:

```rs
fn new(a: f64, b: f64) -> Result<T> {
  if params_valid(a, b) {
    Ok(T { a, b })
  } else {
    Err(StatsError::BadParams)
  }
}
```

Even if they have more lines, functions often only return one specific error variant, which can be replaced by `Option::None`. A user can then check the documentation to quickly figure out why it's returning `Option::None` in most cases (invalid params of some sort).

I opened this as a draft because there are some functions that should probably have a concrete error type, either for future-proofing or because they already return multiple error variants.


FreezyLemon:
I think #265 is more likely at this point, and rebasing this would be a hassle. I'll close this, if it's still on the table I'll re-open or re-do entirely.

PR #260, opened 2024-08-12
Add some more inverse CDF functions

ColmTalbot:
I've been using `statrs` for a project I'm working on, and noticed that many distributions don't have an explicit implementation of the inverse cdf when they can be simply implemented.

I wasn't sure what the expectations are for documentation and tests, I added some tests for a couple of them, but thought I should just ask if there is a standard I'm missing before writing a bunch of boilerplate.

If this contribution is welcome/desired, I'm happy to put in any docs/tests.

### DIFF
--- a/src/distribution/bernoulli.rs
+++ b/src/distribution/bernoulli.rs
@@ -3,6 +3,8 @@ use crate::statistics::*;
 use crate::Result;
 use rand::Rng;
 
+use super::FullDiscrete;
+
 /// Implements the
 /// [Bernoulli](https://en.wikipedia.org/wiki/Bernoulli_distribution)
 /// distribution which is a special case of the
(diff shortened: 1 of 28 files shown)


YeungOnion:
Thanks for the contribution!

For docs we've been writing the mathematical form for expression for inverse cdf's where possible, it more clearly distinguishes them from search implementations.

I think what you've done for tests is good.

It's been mentioned before to use a "nice" way to test fwd and inverse cdf since a triple of `(parameters, x, p)` specifies both ways if not specifying precision, but I'd much prefer having these implemented than having a clearer set of tests.

YeungOnion (approved):
This all looks good to me!

Thanks for fixing some of the other docs and catching the precision typo on the test. 10^+14 relative precision isn't any good.

PR #265, opened 2024-08-16
Implement concrete Error types for each distribution's `new` function

FreezyLemon:
See https://github.com/statrs-dev/statrs/issues/221#issuecomment-2291250694.

Ignore the massive amount of commits, that's for my organisation and can/will be squashed later.

Outstanding issues:
- [x] ~~Clippy (`Empirical`): Returning a `Result<_, ()>` is not best practice. clippy suggests an `Option` which makes less sense here, so either ignore or return a `Empirical` directly (the function is infallible)~~ ignored

- [x] ~~`MultinomialError` has not been implemented (yet). I noticed that the parameter validation is moved to the `internal` module which implies the `check_multinomial` function will be reused somewhere else, so I didn't want to start changing stuff before that project is done~~ Changed my mind on this. Should be fine to change, it's not much code to replicate inside `new` (e.g. Categorical does basically the same check inside `new`)

- [x] Unit Tests
- [x] Proofread
- [ ] After everything is done: Check `StatsError` to see where it is still used and if it can be replaced by something else

### DIFF
--- a/src/stats_tests/fisher.rs
+++ b/src/stats_tests/fisher.rs
@@ -105,6 +105,7 @@ pub enum FishersExactTestError {
 }
 
 impl std::fmt::Display for FishersExactTestError {
+    #[cfg_attr(coverage_nightly, coverage(off))]
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match self {
             FishersExactTestError::TableInvalidForHypergeometric(hg_err) => {
--- a/src/stats_tests/fisher.rs
+++ b/src/stats_tests/fisher.rs
@@ -1,6 +1,5 @@
 use super::Alternative;
-use crate::distribution::{Discrete, DiscreteCDF, Hypergeometric};
-use crate::StatsError;
+use crate::distribution::{Discrete, DiscreteCDF, Hypergeometric, HypergeometricError};
 
 const EPSILON: f64 = 1.0 - 1e-4;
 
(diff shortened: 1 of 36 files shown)


FreezyLemon:
#269 would reduce the effort needed to add tests in this PR significantly, so I'll wait for that.

src/stats_tests/fisher.rs
YeungOnion, line 1: I think you're right that it's better to provide the error in the context of the table's elements, but I'd prefer `Error` over `Option` here as it's less specific to say that there's not a t-statistic for this table so much as this isn't a valid contingency table.

FreezyLemon: I think the line number got lost here. You can click or drag on the line number when reviewing to place the comment on a specific line/lines.

Looking at the changes in this file, I assume you'd want to have a custom error type for the `fishers_exact` and `fishers_exact_with_odds_ratio` functions? As far as I can tell, the only way these can fail is if the passed `population`, `successes` and `draws` (extracted from the `table` parameter) do not form a valid `Hypergeometric` distribution. So a new error type for these functions would effectively be 

```rs
enum FisherTestError {
  InvalidParameters
}
```

which is fine, don't get me wrong. The reason I went with `HypergeometricEeror` here is because it was the simplest solution (just let it propagate with `?`)

FreezyLemon:
- rebased
- added a new test function `test_create_err` which expects a specific error variant
- added a lot of tests
- added `FishersExactTestError` (only one variant atm, which wraps around Hypergeometric -> bit boilerplate-y, but most flexible)

src/stats_tests/fisher.rs
FreezyLemon: Should be resolved with the latest changes. I added a newtype (more info in the other comment)

YeungOnion:
Regarding coverage, our CI is running nightly cargo, so we could use this attribute on functions ([from cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov?tab=readme-ov-file#exclude-function-from-coverage))

in `lib.rs`
```rs
/// ... docs

#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
```

in modules
```rs
impl std::fmt::Display for D {
    #[cfg_attr(coverage_nightly, coverage(off))]
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        //...
    }
}
```

in manifest
```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage,coverage_nightly)'] }
```

Let me know if you want me to do the proofreading as part of my review and I'll check that off when I've done it.

FreezyLemon:
> Let me know if you want me to do the proofreading as part of my review and I'll check that off when I've done it.

Thanks for offering. I plan on going through the docs and Display impls once myself, there's some inconsistencies that I know of which I want to fix.

Feel free to proofread when reviewing, but I'd wait until I've done my first pass.

FreezyLemon:
Alright, I changed the things I wanted to and finally added the [`#[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) attribute to the error types. Should be good to go now

YeungOnion (commented):
Gave it all a read, everything matches up and the language is consistent, thanks for this.

YeungOnion:
added a Send + Sync check like in #226

FreezyLemon:
Yeah. I refactored it a bit.. I realized after 226 that it's trivial to do `T: Sync + Send`, only requiring one helper function which makes it a bit shorter.

YeungOnion (approved):

PR #268, opened 2024-08-27
Implement and document MSRV

### LINKED ISSUE #242: Support MSRV concept
We tried this out for the #235, but ran into some trouble when dev-dependencies were not able to satisfy the version that the lib can. Since it's not tested while on crates.io, perhaps this is a non-issue, but it does mean that the CI will fail because of dependencies that we currently use for benchmarking. It would be nice to ensure that the tests will run on our MSRV. Any other input welcome here regarding what one would expect of the MSRV information and what might be unexpected or where you would expect to look for information if something isn't going your way when using this dependency.

FreezyLemon:
Closes #242.

I figured out the current MSRV via manual testing. `first_key_value()` (found in [empirical.rs](https://github.com/statrs-dev/statrs/blob/aa276c8ef00b39937d56aba0f6e7ca950befe316/src/distribution/empirical.rs#L155)) was [stabilized in 1.66.0](https://releases.rs/docs/1.66.0/), which was released in December 2022.

It's probably not very difficult to lower the MSRV a bit, but it wasn't a priority for this PR.

I added a lockfile containing MSRV-compatible dependency versions to the project. A change in MSRV is [not considered to be breaking](https://github.com/rust-lang/api-guidelines/discussions/231) by most people AFAICT, so the msrv CI job uses the lockfile to ensure nothing breaks over time.[^1]

Note that this file has no effect on compilation unless you manually `mv Cargo.lock.MSRV Cargo.lock`.

This new lockfile however means that changes to the `[dependencies]` in Cargo.toml require updating the `Cargo.lock.MSRV` (CI is run with `--locked`, so a mismatch between `Cargo.toml` and `Cargo.lock` will result in a compile-time error) if you want CI not to fail. It's relatively easy to do if you have access to a nightly toolchain:
`cargo +nightly update -Zmsrv-policy && cp Cargo.lock Cargo.lock.MSRV`.
If you don't, another possibility is just updating all dependencies (`cargo update`) and then manually downgrading packages that are incompatible (`cargo update --precise <msrv-compatible version> <broken package>`). That is pretty tedious though.

Not 100% sure about the note in the README. It's the usual way I've seen it described and done so for other projects myself. Since statrs is pre-1.0, a MINOR version increase means a breaking change, so MSRV should be considered a PATCH version increase. Unless of course it's bundled with other changes, which often makes sense.

[^1]: For example: `approx = 0.5.0` is resolved to version 0.5.1 now and compatible with 1.66.0. If version 0.5.2 is released and has a MSRV of 1.70, then `cargo update` or a fresh `cargo build` (like CI) will resolve to the newer version by default, meaning compilation will break on 1.66.0. Locking approx to version 0.5.1 prevents this from happening.


.github/workflows/test.yml
```diff
@@ -35,8 +35,21 @@ jobs:
       - name: Run rustfmt --check
         run: cargo fmt -- --check
 
+  msrv:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      # rust-version from Cargo.toml
+      - name: Install Rust 1.66.0
```
YeungOnion, line 43: perhaps we could use `cargo hack --rust-version` reading `rust-version` from the manifest instead of managing this manually?

```
- uses: taiki-e/install-action@cargo-hack
```

README.md
```diff
@@ -48,6 +48,10 @@ cargo test
 If you'd like to modify where the data is downloaded, you can use the environment variable,
 `STATRS_NIST_DATA_DIR` for running the script and the tests.
 
+## Minimum supported Rust version (MSRV)
+
+This crate requires a Rust version of 1.66.0 or higher. Increases in MSRV will result in a semver PATCH version increase.
```
YeungOnion, line 53: perhaps some verbiage like,
> Increases in MSRV will be considered a semver non-breaking API change (MINOR after 1.0.0, PATCH until 1.0.0) and require a version increase.

YeungOnion (changes requested):
I think [cargo hack](https://github.com/taiki-e/cargo-hack) could be valuable here to not require consistency between a value in CI and the manifest. But I'd be open to other options to do the same deduplication.

YeungOnion:
Not as familiar as I'd like to be with GitHub's review tools, so let me know if something is unclear.

.github/workflows/test.yml
FreezyLemon: Hmm it's possible, but it makes caching a bit more complicated. I'll see if I can do it..

FreezyLemon: Oh, nevermind. The obvious solution actually works fine. Not sure what I expected

FreezyLemon:
I agree that cargo hack seems useful for this purpose. Added it and rebased on top of current master. Also reworded the README similar to how you suggested, hope it's OK for you.

> let me know if something is unclear.

No worries, everything was clear

YeungOnion (approved):
Thanks again, merging.

FreezyLemon:
Looks like CI is stuck due to https://github.com/rust-lang/rust/issues/130056

Not sure there's much to do about it at the moment. It's just nightly being affected atm

Issue #108, opened 2020-05-04
Change floating point tests to be approximate instead of exact


ghost:
IEEE754 does not lead to reproducible results across platforms, so tests of numerical
routines should be whether the results lay within predetermined bounds instead of testing exact values.

jpgoldberg:
I have used [`assert_approx_eq!`](https://docs.rs/assert_approx_eq/1.1.0/assert_approx_eq/macro.assert_approx_eq.html) for this kind of thing.

troublescooter:
`statrs` has a similar function in `prec.rs`.
At the time of this issue the pattern was more widespread that some tests compared to exact floating point numbers, while others used absolute difference within some epsilon. I've left this issue open because we've been moving away from absolute differences to relative ones which are more insightful over different orders of magnitude. That transition has not been finished.

vks:
We already have an [`assert_almost_eq!` macro](https://github.com/statrs-dev/statrs/blob/e45779c779fcd962ab1180ae2c3e445bb048ae52/src/lib.rs#L42) for this.

YeungOnion:
The cleanup would be nice, also appears heavily in the docstrings. I can start chipping away at that.

here's a list of what I found on a cursory glance
- `function`
  - [ ] `beta`
  - [ ] `erf`
  - [ ] `evaluate`
  - [ ] `exponential`
  - [ ] `factorial`
  - [ ] `gamma`
  - [ ] `harmonic`
  - [ ] `logistic`
- `testing_boiler`
  - [x] in implementation of self
  - [ ] use boiler in distribution tests
- `distribution`
  - [ ] all docstrings
  - whatever is remaining

YeungOnion:
never closed this from before.

PR #269, opened 2024-09-04
Update testing_boiler and unify test helpers

FreezyLemon:
The unit tests are a bit messy at the moment. Duplicate code, subtly inconsistent behavior (`test_case`), bad naming, no documentation etc.

I tried to clean it up a bit. This also makes the problems mentioned in #108 easier to spot (ctrl-f "test_exact"). List of changes:

- Changed some function names:
  - `try_create` and `create_case` -> `create_ok` (dropped some verification but that really shouldn't matter[^1]. If it does, I can change this)
  - `bad_create_case` -> `create_err` (new: `create_err` returns the error, can be used for verification when more specific errors are returned)
  - `test_case` -> either `test_exact` or `test_relative`, depending on old behavior (some implementations used `assert_eq`, others used `assert_relative_eq`)
  - `test_almost` -> `test_absolute`
  - `get_value` -> `create_and_get`
- Added documentation to the test helpers
- Improve messages for test failures in common functions
- Rewrote some `#[should_panic]` tests to use `test_none` instead. Reads better IMHO

There are a few smaller things that could help further (re-enabling rustfmt for unit tests, dis-allowing `clippy::excessive_precision`), but those are less clear-cut, so I kept it to this changeset for now.

[^1]: The type of verification being done was extremely rudimentary, e.g. `let dist = Dist::new(7.5, 3.5);` and then `assert_eq!(dist.a(), 7.5);` plus `assert_eq!(dist.b(), 3.5);`.


FreezyLemon:
Just noticed I missed some `#[should_panic]` tests that I still want to rewrite. And I guess I can also write some tests for the testing_boiler itself, now that it's being used for almost all unit tests

FreezyLemon:
The changed lines have 100% coverage, but total coverage still went down because the PR removes circa 700 lines.

YeungOnion:
I really appreciate this, I think it'll add a lot of value for contributors to have a consistent approach for tests. 

I've no qualms with choices of names. I believe at this point, all arguments in `new` methods are stored in the struct. If that changes (sim to #198), then tests should be added with it. 

I've reviewed the minimal functional changes to the boiler and I think it's fair to specify `f64` until we actually implement over generic or multiple floats. These error messages will be more specific now as well, thanks 👍!

---

I believe that the only way this PR could allow behavior of the library to change in the future is if tests aren't the same, so my plan is to do some semantic diff and make sure everything in a pre-existing test module is a rename (or deletion, in the cases you footnote).

Sound reasonable?

FreezyLemon:
Yeah, of course.

YeungOnion (approved):
reviewed locally, looks good!

FreezyLemon:
I think the auto-merge broke something here, `master` doesn't build atm.

FreezyLemon:
See #277.

PR #270, opened 2024-09-04
Make nalgebra optional

### LINKED ISSUE #196: Adding Feature Flags and optional dependency?
Hi,

This is an amazing package with high quality code. However, I find it hard to include it in my work for the following reasons:

1. I am trying to slim down my binary and if more statistical distributions are needed, I want to add them incrementally by adding more features.
2. For my purpose, I almost never need multivariable distributions. I am already depending on Faer crate for linear algebra and I do not want another package for the same stuff. Again, I think nalgebra should be an optional dependency.

I am using statrs's code right now, which I copied. I am creditting the project. I mostly only need survival functions so I stripped down the code. 

You can find my project here: https://github.com/abstractqqq/polars_ds_extension

FreezyLemon:
Closes #196.

With this, users can remove the `nalgebra` dependency by changing:

```toml
[dependencies]
statrs = "0.17"
```

to

```toml
[dependencies]
statrs = { version = "0.17", default-features = false }
```

This will exclude all distributions depending on `nalgebra` from compilation (and the public API):

```sh
$ cargo public-api --no-default-features | grep -c nalgebra
0
```

There has been some talk in the linked issue about adding a feature for multivariate distributions, but I think for now it's simple enough to hide all nalgebra-dependent types behind this flag. If it becomes necessary, this can still be changed.

The new feature is enabled by default to avoid breakage when upgrading the package, but it probably wouldn't be a big deal to disable it by default either.

MSRV-related note: The `dep:` prefix requires 1.60.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,9 @@ jobs:
       - name: Run cargo clippy
         run: cargo clippy --all-targets
 
+      - name: Run cargo clippy without default features
+        run: cargo clippy --no-default-features --all-targets
+
   fmt:
     runs-on: ubuntu-latest
     steps:
@@ -48,5 +51,5 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
 
       - name: Test default features
-        run: cargo test --all-targets
+        run: cargo test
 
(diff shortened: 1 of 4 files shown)


YeungOnion:
Thanks for this, looks good!

PR #271, opened 2024-09-04
Fix several rustdoc warnings

FreezyLemon:
See `cargo doc` for the warnings. Just some trivial things, like URLs broken across multiple lines and empty code blocks.


YeungOnion (commented):
References can italicize with markdown underscore instead of HTML considering we use backticks for code. 

All else is good, thanks again!

YeungOnion (approved):

PR #272, opened 2024-09-05
Update rustfmt config to not require nightly

FreezyLemon:
This PR updates the rustfmt configuration.. by deleting `rustfmt.toml`.

There are only two or three options that do anything at all. The others are set to their respective default value or otherwise ineffective.

Copied from the commit message:
```text
Options set to their default value:
- wrap_comments
- use_small_heuristics
- fn_single_line

Unnecessary options:
- edition = "2021" (read from Cargo.toml)
- comment_width (wrap_comments = false)

Actual config changes:
- blank_lines_upper_bound: 2 -> 1
- reorder_impl_items: true -> false
```

This config includes unstable options which make a `nightly` toolchain necessary to run `cargo fmt`, which feels unnecessary. As you can see, there is effectively no change to the actual code formatting after this configuration change.


YeungOnion (approved):

YeungOnion:
Thanks!

PR #273, opened 2024-09-05
Add `into_params()` to multivariate distributions

### LINKED ISSUE #180: Mutable/Movable parameters for multivariate normal
Hello, I'm studying the possibility of using this crate for Markov Chain Monte Carlo (MCMC) based inference. In this use case, the log-density of a distribution is evaluated repeatedly at different parameter values. To do that currently, the crate requires re-creating the distributions at each iteration. This isn't much of a problem for scalar distributions, but for the multivariate normal, I have to re-allocate the mean vector and covariance matrix at each iteration (since distributions are immutable), which impacts performance.

Allowing the user to re-set the parameters separately would work:

```
pub fn set_mean(&mut self, mu : &[f64]);
pub fn set_variance(&mut self, var : &[f64]);
```

But a solution that moves the parameters out of the struct would also work (therefore preserving the intended immutable API):

```
pub fn take_parameters(self) -> (DVector<f64>, DMatrix<f64>);
```

Are there any plans to offer something like that?

FreezyLemon:
Closes #180.

Multinomial could technically also benefit from something like this, but the parameter is mutated inside `Multinomial::new`, so there's no easy way to return the original `p` (reversing the mutation could result in a different `p` due to floating-point imprecisions).

The `into_` naming was chosen due to [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) and [convention in std](https://doc.rust-lang.org/std/?search=into_).

### DIFF
--- a/src/distribution/dirichlet.rs
+++ b/src/distribution/dirichlet.rs
@@ -150,6 +150,37 @@ where
             -gamma::ln_gamma(sum) + (sum - self.alpha.len() as f64) * gamma::digamma(sum) - num;
         Some(entr)
     }
+
+    /// Consumes the [`Dirichlet`] and returns the `alpha` parameter
+    /// originally passed to [`new_from_nalgebra`][Self::new_from_nalgebra]
+    /// to construct it.
+    ///
+    /// This can be used to avoid allocations when creating the same
+    /// distribution multiple times.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use statrs::distribution::Dirichlet;
+    /// use nalgebra::dvector;
+    ///
+    /// let alpha = dvector![0.1, 0.3, 0.5, 0.8];
+    /// let dir_1 = Dirichlet::new_from_nalgebra(alpha).unwrap();
+    /// assert_eq!(dir_1.entropy(), Some(-17.46469081094079));
+    ///
+    /// let mut alpha = dir_1.into_params();
+    /// alpha[1] = 0.2;
+    /// alpha[2] = 0.3;
+    /// alpha[3] = 0.4;
+    ///
+    /// let dir_2 = Dirichlet::new_from_nalgebra(alpha).unwrap();
+    /// assert_eq!(dir_2.entropy(), Some(-21.53881433791513));
+    /// ```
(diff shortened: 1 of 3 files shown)


YeungOnion:
I think this is a good start, but allocations for decompositions (if random sampling) and decomp's inverse are necessary to fully address #180 in regards to reallocation. `into_...` a tuple of all parameters and precomputed values and `new_with_precompute` may be a little messy, definitely leaks implementation, and requires logic errors to be upheld by the caller. I realize the last of these is not really an issue.

Can you think of an approach that would keep the buffers for the precomputed values so they can be overwritten on update? I think `nalgebra` has some support for this without directly using `MaybeUninit`. I'll look into it as well.

FreezyLemon:
> Can you think of an approach that would keep the buffers for the precomputed values so they can be overwritten on update?

Well, I'd generally assume that distributions are immutable. So anything we want to re-use (both values and/or allocations) would need to be passed back into some API.

It's a bit convoluted but one way that would almost definitely work would be adding something like:
```rs
impl<D> MultivariateNormal<D>
/* nalgebra trait bounds */
{
    pub fn new_from_nalgebra(mean: OVector, cov: OMatrix) -> Result<Self, Error> {
        Self::new_from_nalgebra_with_buffers(mean, cov, empty_matrix(), empty_matrix())
    }

    pub fn new_from_nalgebra_with_buffers(
        mean: OVector,
        cov: OMatrix,
        precision: OMatrix,
        cov_chol_decomp: OMatrix,
    ) -> Result<Self, Error>
    {
        // Parameter checking etc. ...

        // Real construction logic, but using the provided buffers (precision & cov_chol_decomp)
        // that are then stored inside this struct
        Ok(Self {
            ...
        })
    }
}
```

FreezyLemon:
Maybe some builder pattern with some back-and-forth possibilities.. Tried implementing my idea from the last comment, and it's not very nice. Returning such a large tuple with repeated types makes the API confusing.

FreezyLemon:
I'll think of a better solution for this..

PR #277, opened 2024-09-05
fix: missing/moved type import after auto-merge

FreezyLemon:
#269 was merged after #270 without a manual rebase/merge, resulting in a slightly broken automatic merge. This should fix everything

### DIFF
--- a/src/distribution/internal.rs
+++ b/src/distribution/internal.rs
@@ -277,9 +277,9 @@ pub mod test {
     }
 
     pub mod boiler_tests {
-        use super::*;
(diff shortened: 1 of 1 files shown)


YeungOnion:
Good catch.

Do you think it's better practice to update source branches (with rebase) and see if checks still pass before merging? Or should I just check the merge rebase locally?

FreezyLemon:
Hm.. I'd usually rebase the PR. But this kind of problem is fairly rare in my experience, so it's not something I often think about.

PR #278, opened 2024-09-07
feat: experiment for reusing allocated buffers for Dyn multivariate

YeungOnion:
takes approach of making a few fields `OnceCell`, allowing for init-ing
precomputed values while allow taking them out to modify without
reallocate

Implementing similar way to update location via `set` or `set_with` would also be implemented.
Would also implement nearly identical in StudentsT.

### DIFF
--- a/src/distribution/multivariate_normal.rs
+++ b/src/distribution/multivariate_normal.rs
@@ -9,7 +9,7 @@ use std::f64::consts::{E, PI};
 /// # Panics
 /// will panic on dimension mismatch
 #[inline]
-pub(super) fn pdf_exponent_unchecked<D>(
+pub(super) fn pdf_exponent<D>(
     mu: &OVector<f64, D>,
     precision: &OMatrix<f64, D, D>,
     x: &OVector<f64, D>,
@@ -23,11 +23,11 @@ where
     -0.5 * (precision * &dv).dot(&dv)
 }
 
-/// computes the argument of the normalization term in the normal distribution
+/// computes normalization term in the normal distribution
 /// # Panics
 /// will panic on dimension mismatch
 #[inline]
-pub(super) fn pdf_const_unchecked<D>(mu: &OVector<f64, D>, cov: &OMatrix<f64, D, D>) -> f64
+pub(super) fn pdf_const<D>(mu: &OVector<f64, D>, cov: &OMatrix<f64, D, D>) -> f64
 where
     D: DimMin<D, Output = D>,
     nalgebra::DefaultAllocator: nalgebra::allocator::Allocator<f64, D>
@@ -165,7 +165,7 @@ where
             cov_chol @ Some(_) => Ok(MultivariateNormal {
                 precision: cov_chol.as_ref().map(|ll| ll.inverse()),
                 cov_chol,
-                pdf_const: pdf_const_unchecked(&mean, &cov),
+                pdf_const: pdf_const(&mean, &cov),
                 mu: mean,
                 cov,
             }),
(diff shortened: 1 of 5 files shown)


YeungOnion:
@FreezyLemon an alternative to #273, what are your thoughts?


---
Noting as a tangential discussion, since #180 was referring to MCMC, perhaps there would also be a need for many more variables than `nalgebra` is well suited to. We've mentioned ideas for supporting different backends before but I figure if we can specify some of the API for nalgebra, then we can emulate that where possible for a different matrix crate.

FreezyLemon:
> @FreezyLemon an alternative to https://github.com/statrs-dev/statrs/pull/273, what are your thoughts?

This seems like it will result in a nicer API than 273 for sure. Hm.. without `OnceLock`s or some other thread synchronisation, this will make the struct non-`Sync` and non-`Send` though. And I can see complexity becoming a problem, with the internal state needing to be valid in any circumstance.

YeungOnion:
I realized this morning all of this works because of `take`, so `Option` would be viable

Also, you mentioned how a distribution should be immutable, and I agree, so semantically it would make more sense to have `into_new_from_nalgebra` that takes self and relies on `take` and `clone_into`.

YeungOnion:
redid this with `Option`, regarding async, I think this would be fine for an RWLock style of code since the only times the `Option`s are `None` are when there's a mutable reference to them.

I considered a type that represented this kind of "field that owns a data" that always logically stores a value, but is just a wrapper to an `Option`.

src/distribution/multivariate_normal.rs
```diff
@@ -5,76 +5,39 @@ use nalgebra::{Cholesky, Const, DMatrix, DVector, Dim, DimMin, Dyn, OMatrix, OVe
...
 /// computes the argument of the exponential term in the normal distribution
-/// ```text
-/// ```
-/// # Errors
-/// will error on dimension mismatch
+/// # Panics
+/// will panic on dimension mismatch
 #[inline]
-pub(super) fn density_distribution_exponential<D>(
+pub(super) fn pdf_exponent_unchecked<D>(
```
FreezyLemon, line 12: I'm only skimming the changes at the moment, but the [convention](https://doc.rust-lang.org/stable/std/?search=_unchecked) seems to be that `_unchecked` denotes a function which is `unsafe` and can cause undefined behavior if used incorrectly.

There are *`checked`* functions for number primitives that differ from their non-checked counterparts (`add` vs. `checked_add` which checks for overflow), but functions that panic often don't have a function pre/suffix as far as I know. Maybe there's a place for `pdf_exponent_checked` which returns an `Option`, and then `pdf_exponent` which just uses `pdf_exponent_checked.unwrap()`.

YeungOnion: Following style from std seems good. I suppose documenting panics is all that's needed. Could also implement `try_... -> Result` instead. Since it's internal and fairly transparent, I'll just have the panicking version.

FreezyLemon: Oops. I missed that these were not exposed to the public API, skimmed too fast.

Doesn't matter that much for non-public API IMO

YeungOnion: Agreed. Will mention in #284 as well

FreezyLemon:
Have you benched this? I've tried implementing a builder pattern for some time and benches have always shown it's not really any faster than just reallocating.

Unless the amount of variables (and thereby allocation size) is *much* larger than I've tested, I doubt this makes much difference at all performance-wise.

Some napkin math (please double-check) reveals that for `n` variables, we need to allocate 3 `n x n` matrices and one `n x 1` matrix/vector. For 10 variables, that's 310 `f64`s (8 bytes each), so 2480 bytes i.e. less than 2.5KiB. Such small allocations will usually be incredibly fast unless on memory-constrained hardware.

Even with 1000 variables, we're talking about ~3MiB which really isn't that much. That should fit into CPU cache on most modern hardware..

FreezyLemon:
I did a rudimentary profiling run for 1000 variables, just running `MultivariateNormal::new` 40 times (~10 secs runtime):

https://share.firefox.dev/4enR0hX

The code is basically:

```rs
fn main() {
    let (mean, cov) = create_mean_and_cov();

    for _ in 0..40 {
        MultivariateNormal::new_from_nalgebra(mean.clone(), cov.clone()).unwrap();
    }
}
```

If you look for allocating functions (mostly `clone()`), you'll find that apart from some calls very deep inside `nalgebra` that are probably not avoidable by `statrs`, the allocations that are done don't seem to be very heavy at all. Even the manual clones inside the loop in `fn main` are less than 0.5% of the total runtime.

YeungOnion:
Oh this is a smarter way to do preliminary benchmarks. I was going to check if this in-place mutation was faster 🙃. Thanks for doing this!

What did you use to profile and view with Firefox profiler?

Changing from nalgebra right now does seem a bit like it wastes effort, so I'm averse to it until we get some more feedback.

I'll close this and share thoughts about something benefitting MCMC on the initial issue.

FreezyLemon:
> What did you use to profile and view with Firefox profiler?

I used [samply](https://github.com/mstange/samply), it's really easy to use after a bit of setup and does the firefox profiler stuff for you. `perf record` also works, but is a bit more barebones ([flamegraph](https://github.com/flamegraph-rs/flamegraph) can convert the artifacts from that into nice graphs)

PR #279, opened 2024-09-09
run coverage on stable rust

YeungOnion:


YeungOnion:
at the moment, the [coverage job for 275](https://github.com/statrs-dev/statrs/actions/runs/10776565255/job/29890559629?pr=275) doesn't finish compiling, hangs while reporting `Compiling criterion v0.5.1`.

Wanting to exclude issues from `rustc`, using using updated nightly and updated stable compiled the tests on my machine.

Tested here with stable, and the compile passes, next text is doing it on the branch where the issue was seen.

YeungOnion:
stable passes compile for the rand feature as well, trying nightly.

YeungOnion:
reproduced as expected, as this is identical to #275 
and of course, revert to master and check nightly

YeungOnion:
quoted from #268,
> Looks like CI is stuck due to [rust-lang/rust#130056](https://github.com/rust-lang/rust/issues/130056)
> 
> Not sure there's much to do about it at the moment. It's just nightly being affected

I'm going to approve this because _it compiles on my machine_, unless you think I should hold off, I hope within the week the coverage reports will be good again.

FreezyLemon:
Seems fine.

PR #280, opened 2024-09-10
Revert "run coverage on stable rust"

YeungOnion:
oh dear, I merged the experiment instead of the actual branch.


FreezyLemon:
I mean, it *does* fix the issue while nightly still has the problem, right?

YeungOnion:
Oh you're right.

Pinning a specific date of nightly is probably fine as long as we notice when the nightly works again. I'll add a job step with a timeout so nightly has a chance to succeed and the pinned nightly will provide info.

PR #281, opened 2024-09-10
Reduce MSRV to 1.61

FreezyLemon:
I think it's worth reducing MSRV if the changes are as small as this.

Going lower than 1.61 is probably impossible because it looks like nalgebra requires 1.61 to compile.


YeungOnion (approved):

YeungOnion:
Agreed, thanks

PR #283, opened 2024-09-10
ci: pin nightly rust as temporary fix

YeungOnion:
pins a nightly which isn't ultra-slow (>100mins) to obtain coverage report
also has a true nightly with timeout to see when that works again


.github/workflows/coverage.yml
```diff
@@ -7,15 +7,25 @@ on:
...
 jobs:
   coverage:
+
+    strategy:
+      matrix:
+        include:
+          - toolchain: nightly
+            timeout-minutes: 20
+          - toolchain: nightly-2024-08-29
+    timeout-minutes: ${{ matrix.timeout-minutes }}
```
FreezyLemon, line 17: This might not run because `timeout-minutes` is unset in one of the matrix configs. Maybe just set `timeout-minutes: 20` for all coverage jobs for now, and add a comment to remove it when nightly is fixed?

YeungOnion: ah, thanks. I was thinking that might've been how I voided the whole workflow.

PR #284, opened 2024-09-10
Replace all instances of `StatsError` with custom error types or `Option<T>`

### LINKED ISSUE #221: Consider improving error handling and `StatsError` type

FreezyLemon:
Closes #221, I think.

Added some new error types similar to the ones introduced in #265.

Some functions only got the simpler `Option<T>` return type if they only have one error condition. The functions in MultivariateNormal also have a simple Option return type because they're non-public, and it's expected that the preconditions are fulfilled on every call anyways. No need for fancy error messages (and can be easily changed if need be)

### DIFF
--- a/src/distribution/multivariate_normal.rs
+++ b/src/distribution/multivariate_normal.rs
@@ -1,42 +1,37 @@
 use crate::distribution::Continuous;
 use crate::statistics::{Max, MeanN, Min, Mode, VarianceN};
-use crate::StatsError;
 use nalgebra::{Cholesky, Const, DMatrix, DVector, Dim, DimMin, Dyn, OMatrix, OVector};
 use std::f64;
 use std::f64::consts::{E, PI};
 
-/// computes both the normalization and exponential argument in the normal distribution
-/// # Errors
-/// will error on dimension mismatch
+/// Computes both the normalization and exponential argument in the normal
+/// distribution, returning `None` on dimension mismatch.
 pub(super) fn density_normalization_and_exponential<D>(
     mu: &OVector<f64, D>,
     cov: &OMatrix<f64, D, D>,
     precision: &OMatrix<f64, D, D>,
     x: &OVector<f64, D>,
-) -> std::result::Result<(f64, f64), StatsError>
+) -> Option<(f64, f64)>
 where
     D: DimMin<D, Output = D>,
     nalgebra::DefaultAllocator: nalgebra::allocator::Allocator<f64, D>
         + nalgebra::allocator::Allocator<f64, D, D>
         + nalgebra::allocator::Allocator<(usize, usize), D>,
 {
-    Ok((
+    Some((
         density_distribution_pdf_const(mu, cov)?,
         density_distribution_exponential(mu, precision, x)?,
     ))
 }
 
-/// computes the argument of the exponential term in the normal distribution
-/// ```text
-/// ```
-/// # Errors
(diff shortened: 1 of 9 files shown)


YeungOnion:
I would drop [9fa380a](https://github.com/statrs-dev/statrs/pull/284/commits/9fa380a213625e5807fe34cfa8b1e60673c178d4) since we'll go with a panic only in #278 since it's internal.

YeungOnion:
Wondering if it's valuable to emit a `Result<f64, ValueWithConvergence{value: f64, rel_tol: f64}>` to describe what convergence was like at the final iteration, or perhaps that would make sense as a part of returning values that emit precision as well as converged value.

FreezyLemon:
> I would drop [9fa380a](https://github.com/statrs-dev/statrs/pull/284/commits/9fa380a213625e5807fe34cfa8b1e60673c178d4) since we'll go with a panic only in https://github.com/statrs-dev/statrs/pull/278 since it's internal.

I'll leave this PR as-is for now and update it then.. If i revert any of the changes here then I also have to revert the removal of `StatsError`, which is kind of the end goal of this

YeungOnion:
> I'll leave this PR as-is for now and update it then.. If i revert any of the changes here then I also have to revert the removal of StatsError, which is kind of the end goal of this

that makes sense, odd that deleting `error.rs` was a conflict, but I did the rebase locally and merged that last commit removing it.

FreezyLemon:
@YeungOnion Not sure what happened, but `error.rs` was not deleted in the last commit. Maybe something went wrong during the rebase?

YeungOnion:
Presumably, the commit I force pushed added it back into the tree 🙃

PR #286, opened 2024-09-11
noncentral t distribution

### LINKED ISSUE #267: Add noncentral t distribution
Could you add noncentral t distribution? The noncentral t distribution is required in statistical power calculation.

YeungOnion:
I have some other minor edits in this one
- **lint: `use` should not fully qualify scope**
- **doc: support KaTeX in docs**

and I wrote the function signatures out. The motive for adding the rendering with katex is that these formula are getting unruly, the only reason for all the specificity is that the nomenclature isn't completely qualified for most references.

would close #267

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,3 +56,9 @@ features = ["macros"]
 level = "warn"
 # Set by cargo-llvm-cov when running on nightly
 check-cfg = ['cfg(coverage_nightly)']
+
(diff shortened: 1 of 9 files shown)


YeungOnion:
image

the linked word "here", is [this link](https://en.wikipedia.org/wiki/Beta_function#Incomplete_beta_function) to wikipedia

PR #291, opened 2024-09-13
allow ci to continue on error for newest nightly

YeungOnion:
- **ci: continue past errors of true nightly for coverage**


YeungOnion:
this should be completely resilient against future issues with nightly and we can pin different nightlies that work for us.

I'd like the red x at the PR level to go away, but perhaps this will change sometime (see actions/runner#2347)

FreezyLemon:
This is probably the best solution :+1: Feels bad to have CI fail just because of a broken nightly

PR #293, opened 2024-09-16
Docs: add `#[doc(cfg(feature = "...'))` to feature-gated structs and trait impls

FreezyLemon:
This will add a note "Available on crate feature <...> only." to feature-gated structs, impls etc.

One random example of another crate that uses this is [tokio](https://docs.rs/tokio/latest/tokio/#modules). If you want to try it locally, you can run

```sh
cargo +nightly rustdoc --open -- --cfg docsrs
```

and it should show that some distributions need `nalgebra`, and that the trait implementations for `rand::Distribution` need the rand feature.


YeungOnion (approved):

YeungOnion:
Thanks for this, was wondering if there was a way to not manually include it as a note in the docstring.

PR #294, opened 2024-09-17
fix: Use exp_m1 instead of subtracting directly

soumyasen1809:
- Added fix to use exp_m1 function to compute sf for Gumbel distribution more accurately


YeungOnion:
figured I'd also add tests to make this a notable difference. These two tests fail the previous implementation and pass this one.

PR #298, opened 2024-09-22
refactor: rewrite Dirichlet::ln_pdf

FreezyLemon:
The new code should be easier to read and avoids looping through x twice.


YeungOnion:
Initially, I thought it made sense to fail before computing any gamma functions, but the failure is panicking, so this seems suitable. Will merge.

Issue #136, opened 2021-05-08
Beta skewness limits incorrect


Bi-Modal:
Hello everyone,

I believe I've spotted a bug with the skewness for a beta distribution. 

In distribution::beta the `skewness` function returns incorrect values when `shape_a` or `shape_b` are infinite. Currently this function gives -2.0 for infinite `shape_a` and 2.0 for infinite `shape_b` but these are not the correct limits for the skewness of a beta distribution.

The skewness of a beta distribution with shape parameters α, β is:
[ 2(β - α)/(α + β + 2) ]  * sqrt( (α + β + 1) / (αβ) )

The limit of this as α tends to infinity is 2/sqrt(β). And the limit of this as β tends to infinity is -2/sqrt(α). I've attached a short proof of these. 
[beta_skewness_limits.pdf](https://github.com/boxtown/statrs/files/6446403/beta_skewness_limits.pdf)

If I have access I'll make a PR to fix this.

references:
https://en.wikipedia.org/wiki/Beta_distribution#Skewness
https://mathworld.wolfram.com/BetaDistribution.html

troublescooter:
Thanks for the report, this is correct with that choice of the limit. However it's not clear that's the obvious thing to do. The edge-cases in this library are generally a bit wonky and have been discussed before #102 .
A mathematical limit is defined if lim X -> Y f(X) is the same as f(lim X -> Y) = f(Y) with a reasonable choice of lim X -> Y. The space of all Beta distributions has two parameters, but in this case the limit is an artifact of the parameterization: Established [convergence definitions of random variables](https://en.wikipedia.org/wiki/Convergence_of_random_variables) are defined to be independent of this.
Barring any math errors:
The limit as a distribution for alpha or beta towards inf is the point distribution at the edges 0, 1.
Let's define f(X) = skewness(X) = E((X-mu)^3)/(E((X-mu)^2)^(3/2)
Then one sequence of random variables converging to X = 1 is a=1,2,3.. B(a,b) while different sequence converging to X = 1 is Xn = 1 - 1/n. One sequence leads to the limit skewness(B(inf,b)) = -2/sqrt(b) and the other leads to lim n -> oo skewness(Xn) = 1. Thus the limit does not exist.
 In light of these problems, I intended to remove all the special, degenerate cases, but haven't gotten around to it.

PR #299, opened 2024-09-23
refactor!: forbid α=inf || β=inf in Beta distr.

FreezyLemon:
Currently, `new` only rejects both shapes being infinite. Now, it rejects any of the shapes being infinite.

There's been some discussion (#102) about supporting degenerate parameters for distributions. For `Beta` specifically, almost every function (cdf, sf, pdf, mode, mean, ...) needs one or two special cases for infinite parameters, which doesn't seem sensible. For users needing a shape=+/-INF, it makes more sense to create their own struct and then implement the relevant statrs traits, performance- and complexity-wise.

This fixes `skewness()` returning an incorrect value for numeric limits (#136) and simplifies most `Beta` functions.

Also clarifies and fixes up some docs. Breaking because it removes an enum variant from `BetaError`, which is public.

### DIFF
--- a/src/distribution/beta.rs
+++ b/src/distribution/beta.rs
@@ -26,23 +26,19 @@ pub struct Beta {
 #[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]
 #[non_exhaustive]
 pub enum BetaError {
-    /// Shape A is NaN, zero or negative.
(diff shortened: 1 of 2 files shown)


YeungOnion:
Looks great, would support removing parameters that yield degenerate distributions for other constructors as well

PR #301, opened 2024-09-23
refactor: Categorical now stores normalized values for cdf and sf

FreezyLemon:
(De)normalization needs to be done on every cdf- or sf-related API call at the moment. It's much simpler to handle it like the pmf vector and just do it once in `new`.

This breaks multinomial because it uses some of the `Categorical` helper functions in its implementation. This PR depends on another one resolving this, e.g. #287.


YeungOnion:
pulled into #396, made it under 2 years 🙃

PR #302, opened 2024-09-24
cli example usage

YeungOnion:
figured text output would be quick for verifying implementations with external programs, will help with verifying the implementation of #287

EDIT
[^redact]Figured this is also an example of using the features, but not extending on top of the interface.

[^redact]: removed commit messages from text

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,9 +43,15 @@ optional = true
 default-features = false
 features = ["std"]
 
+[[example]]
+name = "random_clap"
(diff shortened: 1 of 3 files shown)


YeungOnion:
Didn't even think about how dev-dependencies would affect the lockfile. The workflow for MSRV should generally avoid the dev dependencies. Will get back to this after the 0.18 release.

day01:
i think we can close it, lets start with statrs-cli

PR #304, opened 2024-09-26
replace [Discrete]Distribution trait for traits for moments and entropy separately

YeungOnion:
- **feat: define new traits Covariance, Entropy, and CentralMoment**
- **feat!: replace Distribution with new traits for binomial**
- **feat!: replace Distribution with new traits for Gamma**
- **feat!: replace Distribution with new traits for uniform**
- **feat!: replace DiscreteDistribution for new traits for negative binomial**
- **feat!: replace Distribution for new traits with F-distribution**
- **feat!: replace Distribution for new traits for laplace**
- **feat!: replace Distribution for new traits for normal**
- **feat!: replace Distribution for new traits for multivariate vector distributions**


YeungOnion:
@FreezyLemon this removes the `statistics::DiscreteDistribution` and `statistics::Distribution` traits.
I'm relying a bit on `unimplemented` here - was wondering if I should have meaningful associated types for those trait methods as some are unit.

src/statistics/traits.rs
```diff
@@ -56,6 +57,191 @@ pub trait DiscreteDistribution<T: Float> {
...
+
+    /// returns the determinant of the covariance matrix, det(Sigma)
+    fn determinant(&self) -> T;
+}
+
+#[cfg(feature = "nalgebra")]
+mod multivariate {
+    use nalgebra::{Cholesky, Dim, OMatrix, OVector, U1};
+
+    impl<D> super::Covariance<f64> for OVector<f64, D>
```
FreezyLemon, line 78: Should probably have a few tests for this

FreezyLemon (commented):
Looks really good, I like this API. 

> I'm relying a bit on unimplemented here - was wondering if I should have meaningful associated types for those trait methods as some are unit.

Hm.. good question. Setting the correct type feels right, but might be unexpected for a user because the unimplemented!() macro will only panic at runtime.

Probably better to set it to `()` for now, and maybe comment that it's unimplemented ("should be f32 but is unimplemented")? Something like that

Edit: The `#[deprecated("...")]` attribute could also be (mis)used to denote a function as unimplemented. That would create a compile-time warning which should be fairly easy to spot.

YeungOnion:
Would be nice to have the Never `!` type, but that's not stable. Will do this for now. I'll write out the tests next.

YeungOnion:
Okay, I'll write the tests for Covariance now, but I realized that I didn't have to find a way to express in associated type if something is unimplemented, I could just separate the traits.

Aside: I learned about [`ast-grep`](https://ast-grep.github.io/) and it was useful for the rewrite.

YeungOnion:
I also dropped the `sparse` trait method from the Covariance trait and renamed the trait to `CovarianceMatrix` perhaps there's another name more suitable, but the intent was to support various types as if they expressed covariance as an operator and how it acts on states. It felt a little ambiguous across types unlike the `dense` method.

YeungOnion:
Not sure what I'm misunderstanding about colorize/whiten and its testing regarding this.
Since square roots of matrices aren't unique, I'm relying on the quadratic forms like seen in distribution functions that have covariance matrices, like in the gaussian $-\frac{1}{2}x^T \Sigma^{-1}x$
This is also how [scipy does their tests](https://github.com/scipy/scipy/blob/main/scipy/stats/tests/test_multivariate.py#L136)

PR #305, opened 2024-10-02
Implement `FromIterator<f64>` for `Empirical` and remove `from_vec`

FreezyLemon:
The implementation is identical, but expands support to any source of type `T: IntoIterator<f64>`.


YeungOnion:
Thanks, this is a good API choice, even if we don't expose it as a collection. Maybe we can later implement `Extend`

Was kind of thinking that a streaming empirical distribution should be distinct from a static one. Optimizations for inserts/removals would be tree-like data but searches could be faster[^1] flat data over contiguous memory

[^1]: up to details

PR #306, opened 2024-11-04
refactor!: Chi: Use u64 for `freedom`

FreezyLemon:
Removes all degenerate cases for `Chi` (see #102).

Degrees of freedom `k` is usually defined as a positive integer. Storing `k` as a NonZeroU64 avoids all degenerate cases from the start (k < 0, k == 0, k == NaN), and the overhead of casting to an f64 for some operations should be minimal.

NonZeroU64 also allows memory layout optimizations (similar to the nullptr optimization). E.g. `std::mem::size_of::<Result<Chi, ChiError>>()` is now `== 8`, when before it was greater than 8.

Also removed all tests that used k = 0, k = NaN, k = +/- INF and other non-integer k (e.g. k == 2.5)

### DIFF
--- a/src/distribution/chi.rs
+++ b/src/distribution/chi.rs
@@ -2,6 +2,7 @@ use crate::distribution::{Continuous, ContinuousCDF};
 use crate::function::gamma;
 use crate::statistics::*;
 use std::f64;
+use std::num::NonZeroU64;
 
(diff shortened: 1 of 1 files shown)


YeungOnion:
Statically is a great way to remove those cases, thanks!

PR #308, opened 2024-11-10
Refactor Empirical distribution

FreezyLemon:
Nothing here should be breaking. Some of the changes are opinionated code quality improvements.

Two noticeable improvements from this:
- Struct size reduced (platform-dependent, but `Option<(f64, f64)>` is larger than two `f64`s. Up to 8 bytes on x64)
- Avoids `BTreeMap::remove` inside `fn remove` if value > 1

Probably also removes a branch in `fn add`, but I honestly haven't checked the generated ASM or checked via benchmarking. Should I?

### DIFF
--- a/src/distribution/empirical.rs
+++ b/src/distribution/empirical.rs
@@ -2,6 +2,7 @@ use crate::distribution::ContinuousCDF;
 use crate::statistics::*;
 use non_nan::NonNan;
 use std::collections::BTreeMap;
+use std::ops::Bound;
 
 mod non_nan {
     use core::cmp::Ordering;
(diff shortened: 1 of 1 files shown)


FreezyLemon:
Pushed a breaking change on top of the previous ones. [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) is a useful type for this exact purpose, statically proves that no `Err` variant can ever exist of this result *and* allows a layout optimization (`size_of::<Result<Empirical, Infallible>>() == size_of::<Empirical>()`)

YeungOnion:
It looks good, and the tests were definitely needed for the statistics. Thank you, I appreciate what you do.
TIL there's a way around not using nightly's `!`!

YeungOnion (approved):

PR #314, opened 2024-12-06
Add `#![no_std]` support

### LINKED ISSUE #165: no_std support
It would be nice if we can add no_std support for at least the parts don't need rng. It's a good idea to define a rng trait and allow the user to pass it to the library.

FreezyLemon:
Closes #165.

Standard library types that exist in libcore are usually just reexported through `std::`, so changing from e.g. `std::f64` to `core::f64` results in no change in behavior.

There are a few things to note:
- All `nalgebra`-dependent features require std at the moment. I'm not sure `OVector`/`OMatrix` can be used without std because of `Dyn`, but nalgebra does support `no_std` and `SVector`/`SMatrix` should work without libstd. Out of scope here though.
- Some other features also require libstd, so they're disabled in `no_std`. E.g. all functions that have a `Vec` parameter/return type like `OrderStatistics::ranks`.
`core::error::Error` exists, but was only recently stabilized in 1.81.0. I decided to not use this yet and instead not implement the Error trait in `no_std`.
- Tests have worse error messages in no_std. I don't know how to this type of string formatting without `String`, so I just dropped it. Dev machines should usually have it available anyways.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,6 +59,9 @@ jobs:
       - name: Install Rust stable
         uses: dtolnay/rust-toolchain@stable
 
+      - name: Test no_std
+        run: cargo test --no-default-features -F rand
+
       - name: Test default features
         run: cargo test
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,11 +22,13 @@ path = "src/lib.rs"
 [[bench]]
 name = "order_statistics"
 harness = false
-required-features = ["rand"]
+required-features = ["rand", "std"]
 
 [features]
-default = ["nalgebra", "rand"]
-nalgebra = ["dep:nalgebra"]
+default = ["std", "nalgebra", "rand"]
+std = ["nalgebra?/std", "rand?/std"]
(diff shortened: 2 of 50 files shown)


YeungOnion:
I forgot why I didn't merge this earlier, but I also see a few items we can add for future development, 
- empirical as an immutable distribution (I've already been considering this since it's the only distribution with this behavior), an array would be sorted in place, then a second array would be buffered to the same size and filled with the values and cumulative counts.
- OrderStatistic could have an API that outputs iterables instead of explicit Vec, this also means that if I only wanted to know where in a sorted list the first k items in an unsorted list appears, I could have the benefit of only obtaining the rank of the first three values, while worst case is unchanged, I think typical case would improve meaningfully for larger datasets.

Thanks for doing this, I think since I took a while with this there may be more changes required and then we'll have to update PRs.

FreezyLemon:
Quite a few things were added between when I opened this and now.. this probably should've been updated before merging, as e.g. `no-default-features` is now broken on `master`. I'll try to fix those and open a new PR.

PR #315, opened 2025-01-02
Add some hypothesis test functions

mdahlin:
Adds functions for 
* Mann-Whitney U test (per https://users.rust-lang.org/t/mann-whitney-u-test/95005)
* One Way ANOVA F-test
* One Sample t-test

Functions are generally based on the scipy version. I tried to align with existing formatting/setup from the fisher test, but wanted to make sure I was on the right track (in terms of structure, level of documentation, desire for this capability, etc.) before thinking about adding more tests.

### DIFF
--- a/src/stats_tests/mannwhitneyu.rs
+++ b/src/stats_tests/mannwhitneyu.rs
@@ -58,9 +58,8 @@ pub enum MannWhitneyUMethod {
 }
 
 /// ranks data and accounts for ties to calculate the U statistic
-fn rankdata_mwu<T: PartialOrd>(xy: Vec<T>) -> Result<(Vec<f64>, Vec<usize>), MannWhitneyUError> {
-    let mut j = (0..xy.len()).collect::<Vec<usize>>();
-    let mut y = xy;
+fn rankdata_mwu<T: PartialOrd>(mut y: Vec<T>) -> Result<(Vec<f64>, Vec<usize>), MannWhitneyUError> {
+    let mut j = (0..y.len()).collect::<Vec<usize>>();
 
     // check to make sure data can be compared to generate the ranks
     for i in 0..y.len() {
--- a/src/stats_tests/mannwhitneyu.rs
+++ b/src/stats_tests/mannwhitneyu.rs
@@ -476,6 +476,31 @@ mod tests {
         assert_eq!(result, Err(MannWhitneyUError::ExactMethodWithTiesInData));
     }
     #[test]
+    fn test_automatic_asymptotic() {
+        // compare to running same inputs in scipy
+        // both samples len > 8
+        let (statistic, pvalue) = mannwhitneyu(
+            &[19, 22, 16, 29, 24, 28, 7, 10, 30],
+            &[20, 11, 17, 12, 5, 31, 18, 2, 34],
+            MannWhitneyUMethod::Automatic,
+            Alternative::TwoSided,
+        )
+        .unwrap();
+        assert_eq!(statistic, 49.0);
+        assert!(prec::almost_eq(pvalue, 0.47992869214595724, 1e-9));
+
+        // ties in data
+        let (statistic, pvalue) = mannwhitneyu(
+            &[1, 2, 3, 4, 5, 6],
+            &[6, 7, 8, 9, 10],
+            MannWhitneyUMethod::Automatic,
+            Alternative::TwoSided,
+        )
+        .unwrap();
+        assert_eq!(statistic, 0.5);
+        assert!(prec::almost_eq(pvalue, 0.010411098147110422, 1e-9));
+    }
+    #[test]
     fn test_rankdata_mwu() {
         let data = Vec::from([1, 4, 3]);
         let (rank, t) = rankdata_mwu(data).expect("data is good");
--- a/src/stats_tests/mannwhitneyu.rs
+++ b/src/stats_tests/mannwhitneyu.rs
@@ -1,0 +1,503 @@
+//! Provides the [Mann-Whitney U test](https://en.wikipedia.org/wiki/Mann–Whitney_U_test#) and related
+//! functions
+
+use num_traits::clamp;
+
+use crate::distribution::{ContinuousCDF, Normal};
+use crate::stats_tests::Alternative;
+
+/// Represents the errors that can occur when computing the mannwhitneyu function
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]
+#[non_exhaustive]
+pub enum MannWhitneyUError {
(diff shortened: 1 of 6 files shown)


YeungOnion:
Sorry that I've taken a bit to reply here.

So far, these are great. We have mentioned the idea of a nan policy in regards to analytical functions (as opposed to empirical functions) and I think following the scipy approach is good because of developer expectations.

We don't really have enough tests to have a sense of uniform API for tests and having the policy as an argument is useful for establishing that.

*I think the direction you're going in is good and I would approve this* once I look into why the `nightly`-dependent workflow in the CI won't compile. I'm open to you continuing on this PR or opening a dependent PR. 

However, regarding license, to what degree would you say you referred to the scipy source? I don't wish to complicate the license we distribute with, nor do I want to use a license that's not _typical_ for crates on crates.io

mdahlin:
Hey thanks for the response and feedback.  

In terms of the nightly piece. I found the same error locally. A day or so later I updated nightly and everything worked just fine, so it seems like it was an issue specific to nightly.

---

In terms of how much I "referred to the scipy source", it's been a pretty loose reference for the most part but I'll provide some relevant links if you want to form your own opinion.

### one-way ANOVA F-test
- relevant part of the implementation in this PR: https://github.com/statrs-dev/statrs/blob/100e726f4a887c89e15d808746172bda57f0f733/src/stats_tests/f_oneway.rs#L114
- scipy: https://github.com/scipy/scipy/blob/92d2a8592782ee19a1161d0bf3fc2241ba78bb63/scipy/stats/_stats_py.py#L4173

*My conclusion*: commonality with scipy is mainly just the function signature as I leveraged a [statsdirect](https://www.statsdirect.com/help/analysis_of_variance/one_way.htm) page for logic

### One Sample t-test  
- this PR: https://github.com/statrs-dev/statrs/blob/100e726f4a887c89e15d808746172bda57f0f733/src/stats_tests/ttest_onesample.rs#L78
- scipy: https://github.com/scipy/scipy/blob/6e246d0b54dd55dc69232a0caae6772228a7ac25/scipy/stats/_stats_py.py#L6092

*My conclusion*: again mainly just the function signature as I used the logic from this [jpm](https://www.jmp.com/en_us/statistics-knowledge-portal/t-test/one-sample-t-test.html) page

### Mann Whitney U 

Here we'd probably want to look at the two main pieces of logic, being the different methods for calculating the test's statistic, separately

#### Exact
- this PR: https://github.com/statrs-dev/statrs/blob/100e726f4a887c89e15d808746172bda57f0f733/src/stats_tests/mannwhitneyu.rs#L156
- scipy: https://github.com/scipy/scipy/blob/v1.15.0/scipy/stats/_mannwhitneyu.py#L83

*My conclusion*: These are very different from each other. The scipy version is doing a lot of 2-d array stuff and matrix operations that I didn't get into in my implementation.

#### Asymptotic
- this PR: https://github.com/statrs-dev/statrs/blob/100e726f4a887c89e15d808746172bda57f0f733/src/stats_tests/mannwhitneyu.rs#L126
- scipy: https://github.com/scipy/scipy/blob/92d2a8592782ee19a1161d0bf3fc2241ba78bb63/scipy/stats/_mannwhitneyu.py#L149

*My conclusion*: This is probably the **only case worth your review/thoughts**. The scipy version is ~10 LOC and the version in this PR is basically a 1:1 copy of those lines. There isn't too much room for alternative implementation here, but happy to re-write it to avoid any potential issues. Also for reference, the scipy function references [this section](https://en.wikipedia.org/wiki/Mann–Whitney_U_test#Normal_approximation_and_tie_correction) in the Mann Whitney U wiki article.

YeungOnion:
re: nightly
looks good there

---

Thanks for the well-structured reply.

I took a look at each and I agree on your points regarding the implementations. To be fair, had you not done this I would still trust your assessment completely since the author has the best idea of what level they referred to sources.

As a non-expert in licensing, I believe we're in the clear here, especially since it's laid out so clearly in the Wikipedia article.

src/stats_tests/mannwhitneyu.rs
```diff
@@ -0,0 +1,503 @@
...
+    AsymptoticInclContinuityCorrection,
+    /// calculate an approximated (via normal distribution) p-value excluding a continuity
+    /// correction
+    AsymptoticExclContinuityCorrection,
+}
+
+/// ranks data and accounts for ties to calculate the U statistic
+fn rankdata_mwu<T: PartialOrd>(xy: Vec<T>) -> Result<(Vec<f64>, Vec<usize>), MannWhitneyUError> {
+    let mut j = (0..xy.len()).collect::<Vec<usize>>();
+    let mut y = xy;
```
YeungOnion, line 63: hoping this comes off less of a nitpick and more of an "in case you didn't know"
You can bind arguments to be `mut` in the function header without affecting the signature and it can remove this kind of line.

mdahlin: wasn't aware of this. Still pretty new to Rust so appreciate any comments on the more idiomatic things I might be missing. Should be addressed in 50dcaab

mdahlin:
> I'm open to you continuing on this PR or opening a dependent PR.

In regards to this, I just added a couple of more tests. I'm happy to keep working on this PR adding more tests for a bit longer. Or I can just up a new PR for more additions. I don't have too much a preference, and it doesn't seem like you do either.

YeungOnion:
I think I now prefer that the next things be a new PR since these tests could be used for sampling verification which I was working on a bit ago. 

I forgot that we have some `pub use` in the module header, which I'm not certain how I feel about. The names are unlikely to be shadowed. But I can certainly say that not having it is okay as adding it won't break.

Thanks for adding these hypothesis tests!

YeungOnion:
Oh and I didn't get back to you on due diligence for attributing data.

As much as possible, you should provide the source for the dataset where scipy is specifying what that source is; I believe it demonstrates good faith toward the primary source. But after that I'll definitely accept the PR.

mdahlin:
> Oh and I didn't get back to you on due diligence for attributing data.
> 
> As much as possible, you should provide the source for the dataset where scipy is specifying what that source is; I believe it demonstrates good faith toward the primary source. But after that I'll definitely accept the PR.

Sure thing. Are you looking for a full reference section?

YeungOnion:
I think adding a line comment with an inline reference on the relevant tests would suffice.

After all, scipy had their references within their docs, so consuming that information is more common. This is for appropriate attribution, so I think it's fine if it's a little hairy.

YeungOnion (approved):
Thanks for making the attribution changes!

YeungOnion:
Thank you for your contribution! Having tests will be great!

PR #316, opened 2025-01-15
feat: accessor methods to various distribution

### LINKED ISSUE #186: create accessors for all distributions
Some distributions have this (like `Exp`), but others (like `LogNormal`) don't.

for example:
```rust
impl LogNormal {
  pub fn location(&self)->f64 { self.location }
  pub fn scale(&self)->f64 { self.scale }
}
```

this will make it far easier to interoperate these distributions in a variety of ways.

Qazalbash:
Hi, this PR fixes #186. I have implemented the accessor methods for the following distributions,

- [x] Dirac
- [x] Discrete Uniform
- [x] Log Normal
- [x] Triangular
- [x] Uniform
- [x] Multivariate Normal

### DIFF
--- a/src/distribution/dirac.rs
+++ b/src/distribution/dirac.rs
@@ -60,6 +60,20 @@ impl Dirac {
             Ok(Dirac(v))
         }
     }
+
+    /// Returns the value `v` of the dirac distribution
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use statrs::distribution::Dirac;
+    ///
+    /// let n = Dirac::new(3.0).unwrap();
+    /// assert_eq!(n.v(), 3.0);
+    /// ```
+    pub fn v(&self) -> f64 {
+        self.0
+    }
 }
 
 impl std::fmt::Display for Dirac {
--- a/src/distribution/discrete_uniform.rs
+++ b/src/distribution/discrete_uniform.rs
@@ -66,6 +66,44 @@ impl DiscreteUniform {
             Ok(DiscreteUniform { min, max })
         }
     }
+
+    /// Returns the minimum value in the domain of the discrete uniform
+    /// distribution
+    ///
+    /// # Remarks
+    ///
+    /// This is the same value as the minimum passed into the constructor
+    ///
+    /// # Examples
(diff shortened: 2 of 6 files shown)


YeungOnion:
Thanks for working on this. I think it all looks good, but I'm not sure if it's best to leak the implementation of multivariate normal with `precision` and `cov_chol`

If you can make these return cloned/owned values, it will be less binding in our API - there is no implication that the struct owns such data. While I have no expectations for this to change, but we're still fairly new. 

As far as cost goes, the intended matrix sizes for nalgebra are small (dim<5) will be cheap for potential alloc and copies, so an accessor with cloning is cheaper than computing cholesky decomp or inversion.

A possible intermediate for those who need the precomputed values as is: a `take` method that returns a struct where all fields are `pub` containing the same fields as the "`take`n" `MultivariateNormal` but none of statistics `impl`s.

YeungOnion:
And I just realized that I'm giving you mixed messages on copy being opt-in, aren't I?

🤔 Hmm. I think I need someone else's input here. In what cases might you want to view the precompute outside of the features we have already implemented?

Qazalbash:
Hi, I will keep an eye on this PR. Let me know the conclusion.

YeungOnion:
The precompute for the precision matrix is fairly common, so I'm no longer concerned that it's leaking implementation unnecessarily, so a borrow seems good there.

I don't know if it's common enough to store the cholesky $L$, it likely is common. However, the ambiguity I'm wanting to avoid is in whether we would have `OMatrix` for $L$, or if we would have the `Cholesky` type. I don't want to revert the change later, so can we hold off on an accessor for it until we get the specific need.

YeungOnion:
Hey, sorry just getting back to this.

Would you change the cholesky decomp "accessor" to be named,  `clone_cov_chol_decomp` and make a clone instead?

This way it can be there if it's convenient but we aren't tied into having that precompute represented exactly by that type.

Qazalbash:
Hi, I have updated the name along with relevant refactoring.

YeungOnion:
Great! Thanks for this

PR #320, opened 2025-01-26
test: add check for derivative of CDF to ensure it matches PDF

Qazalbash:
I have used finite difference approximation of derivatives. It seems like there are some numerical inaccuracies!

### DIFF
--- a/src/distribution/internal.rs
+++ b/src/distribution/internal.rs
@@ -449,7 +449,9 @@ pub mod test {
 
     /// Does a series of checks that all continuous distributions must obey.
     /// 99% of the probability mass should be between x_min and x_max.
-    pub fn check_continuous_distribution<D: ContinuousCDF<f64, f64> + Continuous<f64, f64>>(
+    pub fn check_continuous_distribution<
+        D: ContinuousCDF<f64, f64> + Continuous<f64, f64> + std::panic::RefUnwindSafe,
+    >(
         dist: &D,
         x_min: f64,
         x_max: f64,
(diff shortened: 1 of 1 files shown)


YeungOnion:
Hmm, I see that this appears for
- Beta
- Student T
- Uniform

I believe the Uniform one makes sense for the error when there's an evaluation over the stepped edges. 

The other two CDF rely on some special functions which could have insufficient precision for finite differences over some steps. That is,

For δ, the precision error in CDF evaluation and ε, the roundoff error from subtraction and representation, and step size h, we have some error from finite difference techniques $h^2$ and from floating point $\epsilon/h$ and the second term is contributing significantly. 

Could you rule out that possibility? Note that our tests for special functions are typically checking at least 12 digits[^*]. 

Let me know if you need any other guidance or advice. 

[*]: referring to a loose notion of the number of base 10 digits, but details are in implementation and our use of [RelativeEq](https://docs.rs/approx/latest/approx/trait.RelativeEq.html#impl-RelativeEq-for-f64) and [AbsDiffEq](https://docs.rs/approx/latest/approx/trait.AbsDiffEq.html#impl-AbsDiffEq-for-f64)

Qazalbash:
It is resolved by comparing,

$$
|F(x+h)-F(x-h)-2hf(x)|<\epsilon
$$

instead of,

$$
\left|\frac{F(x+h)-F(x-h)}{2h}-f(x)\right|<\epsilon
$$

seems like we lost precision due to floating point division.

Qazalbash:
I have set at least one of the checks should pass, i.e. integration of pdf should be equal to cdf or derivative of cdf should be equal to pdf!

src/distribution/internal.rs
```diff
@@ -418,9 +418,40 @@ pub mod test {
         assert!(sum <= 1.0 + 1e-10);
     }
 
+    /// cdf should be the integral of the pdf
```
YeungOnion, line 421: I think this comment is copied over, maybe 
> pdf should be derivative of cdf

src/distribution/internal.rs
```diff
@@ -432,7 +463,16 @@ pub mod test {
...
+        let result_integration = std::panic::catch_unwind(|| {
+            check_integrate_pdf_is_cdf(dist, x_min, x_max, (x_max - x_min) / 100000.0);
+        });
+        let result_differentiation = std::panic::catch_unwind(|| {
+            check_derivative_of_cdf_is_pdf(dist, x_min, x_max, (x_max - x_min) / 100000.0);
+        });
+
+        if result_integration.is_err() && result_differentiation.is_err() {
+            panic!("Integration of pdf doesn't equal cdf and derivative of cdf doesn't equal pdf!");
+        }
```
YeungOnion, line 475: can use the `Result::or` method as a part of result chaining idioms.

```rs
if catch_unwind(|| check_int()).or(catch_unwind(|| check_diff()).is_err() {
    panic!("OOF");
}

YeungOnion:
I think the multiplication is a good solution!

I made some comments, one for an error chaining idiom, and the other on docs.

I'm against using `catch_unwind` unless there's not a clear alternative. However, this is a test and not the lib code, so catching unwinding thrown by anything other than the assertion would be fairly strange behavior. So I'll leave it up to you if that change to the check functions will be in this PR, I know that you also want to get #317 through as well.

YeungOnion (approved):

PR #327, opened 2025-02-12
Switched sample to be correct for gamma rate parameterization

KOVALW:
**Description**
The `gamma` call is parameterized to sample using `rate` rather than `scale`. The original input was the scale as a function of `self.p`, leading to unexpected sample behavior. The `sample_unchecked` is now supplied with the rate calculated from `self.p`.


KOVALW:
Premature PR, replaced with #328

PR #328, opened 2025-02-12
Fix broken negative binomial sampling

KOVALW:
**Description**

The negative binomial distribution sampler is not working because `gamma` is parameterized to sample using `rate` rather than `scale`. The original input was the scale as a function of `self.p` leading to incorrect samples for the specified distribution. The `sample_unchecked` is now supplied with the rate calculated from `self.p` and a test is provided to ensure behavior is in agreement with the theoretical mean and variance of the specified distribution.

### DIFF
--- a/src/distribution/negative_binomial.rs
+++ b/src/distribution/negative_binomial.rs
@@ -489,6 +489,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "rand")]
     fn test_sample() {
         use crate::prec;
         use rand::{distributions::Distribution, SeedableRng, rngs::StdRng};
--- a/src/distribution/negative_binomial.rs
+++ b/src/distribution/negative_binomial.rs
@@ -141,7 +141,7 @@ impl ::rand::distributions::Distribution<u64> for NegativeBinomial {
     fn sample<R: ::rand::Rng + ?Sized>(&self, r: &mut R) -> u64 {
         use crate::distribution::{gamma, poisson};
 
-        let lambda = gamma::sample_unchecked(r, self.r, (1.0 - self.p) / self.p);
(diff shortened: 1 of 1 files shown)


YeungOnion:
Thanks for catching this!

src/distribution/negative_binomial.rs
```diff
@@ -487,4 +487,25 @@ mod tests {
         let sf = |arg: u64| move |x: NegativeBinomial| x.sf(arg);
         test_absolute(3.0, 0.5, 5.282409836586059e-28, 1e-28, sf(100));
     }
+
+    #[test]
```
YeungOnion, line 491: Thanks for seeding so we don't have a flaky test! Would you be able to wrap this test with a `cfg` so that it only compiles with the `rand` feature?

src/distribution/negative_binomial.rs
```diff
@@ -487,4 +487,25 @@ mod tests {
...
+        let tol = 0.1;
+
+        let samples: Vec<u64> = dist.sample_iter(&mut rng).take(n_samples).collect();
+        let sample_mean = samples.iter().sum::<u64>() as f64 / n_samples as f64;
+        let sample_variance = samples.iter().map(|&x| (x as f64 - sample_mean).powi(2)).sum::<f64>() / n_samples as f64;
+
+        let theoretical_mean = dist.mean().unwrap();
+        let theoretical_variance = dist.variance().unwrap();
+
+        assert!(prec::almost_eq(sample_mean, theoretical_mean, tol));
```
YeungOnion, line 508: I have an optional suggestion here, would you use the hypothesis testing module to verify this sampler? I'd use chi square GoF. If not, I was thinking of adding these kinds of tests for sampling over time.

YeungOnion:
The only failing test is an unrelated one we're working on in #325, thanks for contributing a fix!

PR #329, opened 2025-02-16
feat(stats_tests): implement KS test

mdahlin:
Implements versions of the one-sample and two-sample KS test

### DIFF
--- a/src/stats_tests/ks_test.rs
+++ b/src/stats_tests/ks_test.rs
@@ -119,7 +119,7 @@ fn onesample_kolmogorov_twosided_pvalue(d: f64, n: f64) -> f64 {
 
 #[cfg(feature = "nalgebra")]
 fn onesample_marsaglia_et_al_twosided_pvalue(d: f64, n: f64) -> Result<f64, KSTestError> {
-    use nalgebra::DMatrix;
+    use nalgebra::{DMatrix, DVector};
     // Marsaglia, Tsang & Wang (2003)
     // `factorial` can only handle up to 170... could use ln factorial
     if n as usize >= 170 {
@@ -132,8 +132,6 @@ fn onesample_marsaglia_et_al_twosided_pvalue(d: f64, n: f64) -> Result<f64, KSTe
 
     let mut mm = DMatrix::<f64>::zeros(m, m);
 
-    // PERF: definitely a better way to fill the matrix. Also could cache the
-    // factorial calculations to save time as well
     for j in 0..m {
         for i in 0..m {
             if j == 0 {
@@ -153,13 +151,22 @@ fn onesample_marsaglia_et_al_twosided_pvalue(d: f64, n: f64) -> Result<f64, KSTe
             }
         }
     }
+    let mut nn = n as usize;
+    let k = k as usize - 1; // now have k reflect the index
 
-    let mut t = mm.clone();
-    for _ in 0..(n as usize - 1) {
-        t = &mm * t;
+    let mut v = DVector::<f64>::zeros(m); // unit vector
+    v[k] = 1.0;
+    let mut a = mm.clone();
+
+    while nn > 0 {
+        if nn % 2 != 0 {
+            v = &a * v;
+        }
+        a = &a * &a;
+        nn /= 2; //floor division
     }
 
-    Ok(t[(k as usize - 1, k as usize - 1)] * factorial::factorial(n as u64) / n.powi(n as i32))
+    Ok(v[k] * factorial::factorial(n as u64) / n.powi(n as i32))
 }
 
 /// Kolmogorov-Smirnov (KS) Test for one sample against [`ContinuousCDF`]
--- a/src/stats_tests/ks_test.rs
+++ b/src/stats_tests/ks_test.rs
@@ -1,0 +1,1084 @@
+//! Provides the [Kolmogorov-Smirnov (KS) test](https://en.wikipedia.org/wiki/Kolmogorov–Smirnov_test) and related
+//! functions
+
+use core::f64;
+use std::iter::zip;
+
+use nalgebra::DMatrix;
+use num_traits::clamp;
+
+use crate::distribution::ContinuousCDF;
+
+use crate::function::factorial::{binomial, factorial};
+
+use super::NaNPolicy;
+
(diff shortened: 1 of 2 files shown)


mdahlin:
https://github.com/statrs-dev/statrs/blob/5fad268627031618421fb1c10ef53f203133072c/src/stats_tests/ks_test.rs#L246

My implementation to calculate the exact p-value for a one sample KS test requires the use of nalgebra to do some matrix multiplication. I was wondering if there were suggestions for how to add the feature gate. I'm not too familiar with feature gating, but the two ideas that came to my mind were to
1. separate `ks_onesample` and `ks_twosample` and feature gate all of `ks_onesample`
2. feature gate specifically the enum variant (though this seems like it could cause some issues)

YeungOnion:
Thanks for this!

Since feature gating is for conditional compilation, we won't be able to just bar the enum, since that's from a specific code path, but if you walked down that code path and feature gated the whole path that would compile with the feature flag disabled.

I'll use the review tools to make the suggestions so they're near to the code.

src/stats_tests/ks_test.rs
```diff
@@ -0,0 +1,1084 @@
+//! Provides the [Kolmogorov-Smirnov (KS) test](https://en.wikipedia.org/wiki/Kolmogorov–Smirnov_test) and related
+//! functions
+
+use core::f64;
+use std::iter::zip;
+
+use nalgebra::DMatrix;
```
YeungOnion, line 7: either gate this or write the use statement near to usage

src/stats_tests/ks_test.rs
```diff
@@ -0,0 +1,1084 @@
...
+    Less,
+    /// uses [Birnbaum & Tingey (1951)](doi.org/10.1214/aoms/1177729550) to calculate the p-value
+    /// for the one-sided hypothesis test.
+    Greater,
+    /// uses [Marsaglia, Tsang & Wang (2003)](doi.org/10.18637/jss.v008.i18) to calculate the
+    /// p-value for the two-sided hypothesis test. This implementation can become slow for larger
+    /// `n`s and will error with if there are ties in the input data or the input data is too
+    /// large. The threshold for too large is data with length 170 lining up with the
+    /// implementation of [`factorial`] being used.
+    TwoSidedExact,
```
YeungOnion, line 78: gate this like you mentioned
I believe it can go before or after the docstring.

src/stats_tests/ks_test.rs
```diff
@@ -0,0 +1,1084 @@
...
+        if term.abs() < 1e-10 {
+            break; // break when added term is relatively small
+        }
+        k += 1.0;
+    }
+
+    2.0 * sum
+}
+
+fn onesample_marsaglia_et_al_twosided_pvalue(d: f64, n: f64) -> Result<f64, KSTestError> {
```
YeungOnion, line 119: this one would be gated

src/stats_tests/ks_test.rs
```diff
@@ -0,0 +1,1084 @@
...
+        return Err(KSTestError::ExactAndTooLarge);
+    }
+
+    let k = (n * d).ceil();
+    let m = 2 * k as usize - 1;
+    let h = k - n * d;
+
+    let mut mm = DMatrix::<f64>::zeros(m, m);
+
+    // PERF: definitely a better way to fill the matrix. Also could cache the
```
YeungOnion, line 132: There's certainly another way to fill this, but I think the readability would improve significantly, but I don't think the performance would significantly. I don't see any overwrites in my review. But some of the evaluations could be SIMD as you do have a few instances of, "set the first column by this expressions" and "set these antidiagonals by this expression" so it would be interesting to see how much the gains could be.

Some bigger gains could be improving calculating a matrix element of M^n, where M=`mm` especially where the number of bins is large (matmul is O(N^3)) as nalgebra is _primarily_ targeted for graphics. `faer` may be better choice for this or using `nalgebra-lapack` to connect to battle-tested fortran. But let's take that decision out of this scope. For computing a matrix element,

If we have the unit vector along dimension k to be u, then we can compute matrix element k,k as u^T M u, which could be expressed as below, where n/2 is floor division.

$$
u^T M^n u = \left(u^T M^{n/2}\right) \left(M^{n/2}u\right) = v^T v, \quad n \textrm{ mod } 2 == 0
$$

and for odd n, we could raise M to the floor of half the power and add the explicit multiply of M, again with n / 2 being floor divide,

$$
v^T \left(M^{n/2 + 1}u\right)
$$

diagonalization would be useful here as well, unsure at what size the tradeoff will benefit to do n/2 matmul vs diagonalize since both are O(N^3), but matmul seems like less bookkeeping to parallelize. 


And the factorial function is cached prior, so you're good there for less than 170!

mdahlin: Implemented. Thanks for the suggestion. 

|                 | Original | Updated |
|-----------------|----------|---------|
| d=0.15; n=8.0   | 638ns    | 521ns   |
| d=0.15; n=135.0 | 928μs    | 60μs    |
| d=0.62; n=8.0   | 4.4μs    | 3.0μs   |
| d=0.62; n=135.0 | 44.3ms   | 2.7ms   |

YeungOnion:
The test failure is related to something we're fixing in #325, so this is good to go. Thanks for the quick bench. If you've got that runner available, would you be able to share it? I'd use it to test if another optimization is significantly worthwhile.

mdahlin:
> If you've got that runner available, would you be able to share it?

It was just some basic code I pulled together with criterion, so not sure how helpful it is, but see below

```rust
fn bench_exacts(c: &mut Criterion) {
    let mut group = c.benchmark_group("Exact");
    for i in [(0.15, 8.0), (0.15, 135.0), (0.62, 8.0), (0.62, 135.0)].iter() {
        group.bench_with_input(
            BenchmarkId::new("Original", format!("{:?}", i)),
            i,
            |b, (d, n)| b.iter(|| original(*d, *n)),
        );
        group.bench_with_input(
            BenchmarkId::new("Alternative", format!("{:?}", i)),
            i,
            |b, (d, n)| b.iter(|| better_matmul(*d, *n)),
        );
    }
    group.finish();
}
criterion_group!(benches, bench_exacts);
criterion_main!(benches);

```

PR #331, opened 2025-03-08
chore: Bumped `rand` to 0.9.0 and mechanically updated usage

RobertJacobsonCDC:
This PR bumps the `rand` dependency to `rand@0.9.0`. This is a breaking change. The code has been mechanically modified to accommodate the API changes in `rand`. Any changes to functionality are unintentional errors. 

The one exception to this is the very first example in the lib's main doc comment, `src/lib.rs:14`, which originally used `rand`'s `OsRng`. In the latest version of `rand`/`rand_core`, `OsRng` does not implement the `RngCore` trait (the `Rng` trait) but rather the `TryRngCore` trait. As `Distribution::sample` requires the RNG to implement `RngCore`, we have to use an alternative. The natural choice seems to me to be `ThreadRng`, which is a variant of `StdRng` that reseeds from `OsRng`. 

The `narray` dependency does not yet have a release that is compatible with `rand@0.9`, but it's latest development code does. Therefore, I have `narray` pulled directly from its GitHub repo.

You might not want to merge this PR yet for this reason. I'm offering it in case it saves you a little bit of work when you're ready.


CHANGELOG.md
```diff
@@ -137,7 +137,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [v0.10.0](https://www.github.com/statrs-dev/statrs/compare/v0.9.0...v0.10.0)
 
 - upgrade `rand` dependency to `0.5`
-- Removes the `Distribution` trait in favor of the `rand::distributions::Distribution` trait
```
YeungOnion, line 140: and this change should revert

YeungOnion:
Thanks for going through this! I'll keep an eye out for nalgebra's next release on crates.io.

Looks like CI doesn't like our code for benching. I'd be okay updating that to be more generic. I may also modify CI so that's not a blocker for unit tests.

CHANGELOG.md
RobertJacobsonCDC: My changes were a little too mechanical.

benches/order_statistics.rs
```diff
@@ -4,15 +4,15 @@ use rand::prelude::*;
...
     let to_random_owned = |data: &[f64]| -> Data<Vec<f64>> {
-        let mut rng = thread_rng();
+        let mut rng = rand::rng();
         let mut owned = data.to_vec();
         owned.shuffle(&mut rng);
         Data::new(owned)
     };
-    let k = black_box(rng.gen());
-    let tau = black_box(rng.gen_range(0.0..1.0));
+    let k = black_box(rng.random_range(..=usize::MAX));
```
RobertJacobsonCDC, line 14: There is an odd issue: It turns out that [`Distribution<T>`](https://docs.rs/rand/latest/rand/distr/struct.StandardUniform.html#impl-Distribution%3Cu128%3E-for-StandardUniform) is implemented for all primitive numeric types except for `usize`. The `rand` maintainer [appears to be unwilling to support `usize` and describes some workarounds](https://github.com/rust-random/rand/issues/1563). In this particular case, identical functionality is provided by `Rng.random_range(..usize::MAX)`.

src/distribution/multivariate_students_t.rs
```diff
@@ -217,7 +217,7 @@ where
     /// `L` is the Cholesky decomposition of the scale matrix,
     /// `Z` is a vector of normally distributed random variables, and
     /// `μ` is the location vector
-    fn sample<R: ::rand::Rng + ?Sized>(&self, rng: &mut R) -> OVector<f64, D> {
+    fn sample<R: ::rand::Rng + ::rand::RngCore + ?Sized>(&self, rng: &mut R) -> OVector<f64, D> {
```
nataziel, line 220: Doesn't `::rand::Rng` imply `::rand::RngCore`?

OrionYeung: Oh that's true, since Rng is the user level trait, it requires core.

Cargo.toml
```diff
@@ -36,12 +36,12 @@ approx = "0.5.0"
 num-traits = "0.2.14"
 
 [dependencies.rand]
-version = "0.8"
+version = "0.9.0"
 optional = true
-default-features = false
+features = ["std", "os_rng"]
```
FreezyLemon, line 41: Please revert this line. Generally the rand code in `statrs` does not need `std` outside of some doc comments and nalgebra. This interaction is handled under the `[features]` table, where the `std` feature enables `rand/std` which includes `os_rng`:

```toml
std = ["nalgebra?/std", "rand?/std"]
# at the moment, all nalgebra features needs std
nalgebra = ["dep:nalgebra", "std"]
rand = ["dep:rand", "nalgebra?/rand"]
```

src/statistics/slice_statistics.rs
```diff
@@ -133,10 +135,8 @@ impl<D: AsMut<[f64]> + AsRef<[f64]>> Data<D> {
 }
 
 #[cfg(feature = "rand")]
-impl<D: AsRef<[f64]>> ::rand::distributions::Distribution<f64> for Data<D> {
+impl<D: AsRef<[f64]>> ::rand::distr::Distribution<f64> for Data<D> {
     fn sample<R: ::rand::Rng + ?Sized>(&self, rng: &mut R) -> f64 {
-        use rand::prelude::SliceRandom;
```
FreezyLemon, line 138: I think it makes more sense to move the `use rand::prelude::IndexedRandom;` here for consistency with other `impl rand::distr::Distribution<...>` blocks

FreezyLemon (changes requested):
The code changes are good, just one nitpick; see comment.

The feature matrix in `Cargo.toml` needs to be fixed though, otherwise `no_std` will become unusable for `rand` users.

This also needs a big MSRV bump because nalgebra 0.34.0 requires Rust 1.87.0:
- Update `rust-version` in Cargo.toml to 1.87.0
- Change version in the MSRV section of the README
- Update the MSRV lockfile: `cargo +nightly update -Zmsrv-policy && cp Cargo.lock Cargo.lock.MSRV` (needs nightly)

If you need help with anything (especially MSRV stuff can be tedious), let me know.

RobertJacobsonCDC:
@FreezyLemon Thanks! Take a look at the latest commit on this PR. I followed your instructions, but there's something suspicious going on with the `glam` dependency. It looks like `nalgebra` has per version features for `glam`. I don't understand what I'm looking at, though, so maybe it's ok.

k88hudson-cfa:
@YeungOnion We would still really love to see this merged if you're able to get to it!

Issue #330, opened 2025-02-18
Discrete inverse CDF panics if the correct result is the minimum of the distribution


PlasmaPower:
Consider the following testcase:

```rs
#[cfg(test)]
mod test {
    use statrs::distribution::{Binomial, DiscreteCDF};

    #[test]
    fn binomial() {
        let binomial = Binomial::new(0.05, 2).unwrap();
        dbg!(binomial.cdf(0));
        assert_eq!(binomial.inverse_cdf(0.5), 0);
    }
}
```

It panics with the following output:
```
---- test::binomial stdout ----
[src/main.rs:174:9] binomial.cdf(0) = 0.9024999999999995
thread 'test::binomial' panicked at /home/lee/.cargo/registry/src/index.crates.io-6f17d22bba15001f/statrs-0.18.0/src/distribution/mod.rs:220:81:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
             at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/core/src/panicking.rs:76:14
   2: core::panicking::panic
             at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/core/src/panicking.rs:148:5
   3: core::option::unwrap_failed
             at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/core/src/option.rs:2009:5
   4: core::option::Option<T>::unwrap
             at /home/lee/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:972:21
   5: statrs::distribution::DiscreteCDF::inverse_cdf
             at /home/lee/.cargo/registry/src/index.crates.io-6f17d22bba15001f/statrs-0.18.0/src/distribution/mod.rs:220:9
```

I believe the correct result is zero. In any case, this is not a described panic condition from the inverse_cdf docs. The issue seems to lie in the fact that `integral_bisection_search` returns the "smallest k such that f(k) >= z" but it actually returns None if `f(lb) > z` in its first if statement, causing `inverse_cdf` to panic.

PR #332, opened 2025-03-12
fix: one more condition in inverse_cdf to account for small enough probabilities.

geosarr:
Fix: #330

### DIFF
--- a/src/distribution/bernoulli.rs
+++ b/src/distribution/bernoulli.rs
@@ -321,4 +321,15 @@ mod testing {
         test_absolute(0.3, 0.3, 1e-15, sf(0));
         test_absolute(0.7, 0.7, 1e-15, sf(0));
     }
+
+    #[test]
+    fn test_inverse_cdf() {
+        let invcdf = |arg: f64| move |x: Bernoulli| x.inverse_cdf(arg);
+        test_exact(0., 0, invcdf(0.));
+        test_exact(0., 0, invcdf(0.5));
+        test_exact(1., 0, invcdf(0.));
+        test_exact(1., 1, invcdf(1.));
+        test_exact(1., 1, invcdf(1e-6));
+        test_exact(0.5, 0, invcdf(0.25));
+    }
 }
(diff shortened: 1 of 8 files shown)


PlasmaPower:
It'd probably be good to add a regression test to this as well

YeungOnion:
Seems like your condition is the general condition for evaluating to the minimum - it could replace the first condition of checking zero input. And the test would be great.

You need not do the test in the general sense where the default impl is tested on a type that's solely defined for the test, just adding it to those that use the default should suffice.

Let me know if you need anything!

geosarr:
Thanks for your comments @PlasmaPower and @YeungOnion.

I added some tests. But I noticed that the following test fails.

```
#[cfg(test)]
mod test {
    use statrs::distribution::{Bernoulli, DiscreteCDF};

    #[test]
    fn test_inverse_cdf() {
        assert_eq!(0, Bernoulli::new(0.5).unwrap().inverse_cdf(0.5));
    }
}
```
It seems to be due to the implementation of Bernoulli::cdf, which uses the output of function::beta::beta_reg. Currently Bernouilli::cdf(0) is  0.499... instead of 0.5, on my machine.

It may be probably better to use the "exact implementation" as mentionned in the doc of Bernoulli::cdf:

Bernoulli::cdf(x) = if x >= 1 { 1 } else { 1 - p }

YeungOnion:
I've been wondering when using the binomial distribution to represent the Bernoulli distribution would pose an issue. 

Lmk if you'll write the fix and/or refactor for implementing Bernoulli as its own case. If not, I can start and finish that real quick at some time within the next 24 hours.

geosarr:
@YeungOnion I added the implementation of Bernoulli::cdf.

As a general rule, It think it'd probably be better to use the exact implementations wherever possible, in particular for the statistics like mean, variance, median, skewness, etc of any simple enough distribution.

YeungOnion:
Thanks for making the change.

The only test failure is because #325 is not merged yet. If you're willing, I'd appreciate some feedback on that PR. If not, no pressure.

Thanks for the contribution!

Issue #98, opened 2019-05-27
Math.NET behaviour for INFINITY values


troublescooter:
What is the reason for adopting this odd behaviour? It seems to me that the most sensible thing to do would be to return 0 if `shape` isn't NaN or (+-)INFINITY. This is just one instance, as there seem to be a couple of other odd return values.

```
    impl Mean<f64> for Gamma {
    /// Returns the mean of the gamma distribution
    ///
    /// # Remarks
    ///
    /// Returns `shape` if `rate == f64::INFINITY`. This behavior
    /// is borrowed from the Math.NET implementation
    ///
    /// # Formula
    ///
    /// ```ignore
    /// α / β
    /// ```
    ///
    /// where `α` is the shape and `β` is the rate
```

boxtown:
A lot of these as indicated by the doc-string were borrowed behavior from `Math.NET` but I agree that not all of this behavior is necessarily correct. Looking through the commits of the `Math.NET` source I don't see a good reason for this particular method to not return `0`, I'm happy to accept any PRs addressing this and other odd behavior you've seen

troublescooter:
Fixed in #116

PR #333, opened 2025-03-19
fix(distribution): enforce finite params in Gamma::new

### LINKED ISSUE #103: Dis-allow `f64::INF` as valid a parameters for Gamma distribution
Relevant issues:
https://github.com/boxtown/statrs/issues/98
https://github.com/boxtown/statrs/issues/102

It does not necessarily make sense to allow INF as a valid value for Gamma distribution parameters as it just ends up leading to generate cases. We should instead return `Err(StatsError::BadParams)`

oOOo-YKS:
- before:
    - Gamma::new raises an error oly when both shape and rate are infinite
- after:
    - Add infinite checks for shape/rate parameters
    - Add test cases for invalid infinite params

Closes #103
Refs: #98, #102


oOOo-YKS:
sorry but I've made an error , it should be 
- before:
    -  Gamma::new raises an error only when both shape and rate are infinite

YeungOnion:
This is also the approach we took for Beta in #299, I think this is the least ambiguous route forward since our API is a floating point evaluation instead of a more thorough analysis approach.

Could you also,

1. deprecate the `GammaError::ShapeAndRateInfinite` variant?
1. update the `Display` impl for `GammaError` to mention finiteness restrictions?
1. drop the related checks for inf shape or rate in the other function implementations in this file?

oOOo-YKS:
For sure. I've finished all these three requirements and commit.

YeungOnion:
I've made one comment, but overall it looks good, thanks for the fix

YeungOnion (approved):

PR #336, opened 2025-04-09
perf: Various improvements to chisquare

FreezyLemon:
The current code always allocates a new `f_exp: Vec<f64>` just to iterate through it later. This is not necessary:
- if f_exp is `Some(..)`, we can iterate through the slice itself
- if f_exp is `None`, we don't need to iterate anything. we can just set a fixed value for `e` and use that.

This has the added benefit of being `no_std`-compatible.

The amount of iterations through `f_obs` can also be reduced in some cases. If we need to `f_obs.iter().zip(f_exp)` anyways, there's no reason to do `f_obs.iter().sum()` beforehand. We can just calculate the sum "on the way", so to speak. The same goes for `f_exp`.

### DIFF
--- a/src/stats_tests/chisquare.rs
+++ b/src/stats_tests/chisquare.rs
@@ -69,25 +69,41 @@ pub fn chisquare(
     if n <= 1 {
         return Err(ChiSquareTestError::FObsInvalid);
     }
-    let total_samples = f_obs.iter().sum();
-    let f_obs: Vec<f64> = f_obs.iter().map(|x| *x as f64).collect();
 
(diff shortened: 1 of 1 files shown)


YeungOnion:
whoops! used the web editor for the merge and missed the closing brace. Will fix and retry.

YeungOnion:
oh man, I boofed it. I think I'll need you to fix it. Missing semicolon closing src/stats_tests/chisquare.rs:90

FreezyLemon:
No worries

PR #337, opened 2025-04-09
Fix new `no_std` errors

FreezyLemon:
Fixes the errors introduced by #314. These are just the lowest effort way to get it to compile in `no_std` again, some of the new code can maybe be rewritten to become `no_std`-compatible. This should allow `master` to pass CI again though.

### DIFF
--- a/src/distribution/dirichlet.rs
+++ b/src/distribution/dirichlet.rs
@@ -369,7 +369,7 @@ where
 mod tests {
     use super::*;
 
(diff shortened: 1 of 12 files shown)


FreezyLemon:
Test failures seem to be unrelated...?

nataziel:
Yep that Levy test has been consistently flaky, shouldn't be a blocker for merge

PR #341, opened 2025-05-25
tests only require one numerical relationship between cdf and pdf to be close

YeungOnion:


YeungOnion:
hmm, the test failure for geometric is interesting. But these changes are solely around the numerical integration/differentiation relationship between cdf and pdf. I'll open an issue.

PR #343, opened 2025-05-31
feat: add specialised inverse cdf implementation for geometric distribution

### LINKED ISSUE #342: issue in default inverse_cdf, found with simple case of Geometric

nataziel:
closes https://github.com/statrs-dev/statrs/issues/342

### DIFF
--- a/src/distribution/geometric.rs
+++ b/src/distribution/geometric.rs
@@ -548,5 +548,14 @@ mod tests {
         test_exact(1., 1, invcdf(1.));
         test_exact(0.2, 1, invcdf(0.2));
         test_exact(0.004, 173, invcdf(0.5));
+        test_exact(0.5, u64::MAX, invcdf(1.));
+        test_exact(0.5, 2, invcdf(0.75));
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_inverse_cdf_panic() {
+        let invcdf = |arg: f64| move |x: Geometric| x.inverse_cdf(arg);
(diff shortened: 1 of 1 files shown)


nataziel:
@YeungOnion would you mind taking a look at this?

YeungOnion:
Thanks for the ping and for contributing!

Would you be able to write a few tests, specifically for
small p (2e-16) with x near 1, maybe 1-1e-15? Just wanting to know if there should be a warning output for small p regarding an upper bound to k, and if so, at what range of p.

Since the domain of CDF is natural numbers but the type is u64, want to ensure the inversion property is upheld for the range of values in u64. 

Does my request seem justified?

YeungOnion:
Let me know if you need any clarity on the request.

nataziel:
Request is fine, just need to find time to implement! Will try this weekend.

YeungOnion:
Hey, did you need anything here? I totally fell off the radar, if you ran into issues, I can work on a test case.

I'll probably test include it as a separate pr to see if it even passed with the search implementation. It can be tricky to know if the test is even right for things like this.

YeungOnion:
Merged manually via cherry-pick rebase onto master (4c3add9..178984b) to preserve author attribution and avoid the merge commit from the PR branch. All tests pass.

YeungOnion:
sorry I took forever on this one, I'll add in some pathological tests as well. Thanks for the fix! or feature I suppose, correctness under precision error is a whole thing.

PR #346, opened 2025-07-06
implemented anderson darling test

DanielLacina:


src/stats_tests/anderson_darling.rs
```diff
@@ -0,0 +1,101 @@
...
+                write!(f, "Sample size `n` must be greater than 0.")
+            }
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for AndersonDarlingError {}
+
+pub fn anderson_darling<T: ContinuousCDF<f64, f64>>(f_obs: &Vec<f64>, dist: &T) -> Result<(f64, f64), AndersonDarlingError> {
```
YeungOnion, line 23: We can choose the API to borrow slice (`&[f64]`) instead of borrowing Vec (`&Vec<f64>`).
One advantage of this is that users can call with borrowed slice or borrowed Vec, and this is because `Vec<T>` implements `Borrow<&[T]>` (read like _`Vec<T>` can be borrowed as `&[T]`_ ) and you'll be able to observe this in that your tests will not need to change.

DanielLacina:
@YeungOnion I resolved the issue.

YeungOnion (approved):

DanielLacina:
I fixed the formatting and added one test to coverage more code.

YeungOnion:
The clippy error isn't on you, thanks for this!

PR #347, opened 2025-07-20
ci: Use -x instead of -1.0 * x

FreezyLemon:
This throws clippy warnings in newer Rust versions.


FreezyLemon:
The coverage error seems related, which is surprising. I just used `cargo clippy --fix`. I'll investigate, this might be about operator precedence

FreezyLemon:
The Windows test failure should be unrelated, as it doesn't use the `integral` function.

PR #349, opened 2025-08-11
Windows CI failure

FreezyLemon:
Looks like something about `Geometric::cdf` is less accurate on Windows/MSVC than it is on other platforms, causing problems with inverse_cdf. This PR is just here to show the problem as I am not sure how to fix it.


day01:
Looks like this was fixed in `main` and Windows CI is passing now. Can @YeungOnion  this PR be closed as obsolete?

PR #354, opened 2025-09-24
fix Clippy failures in CI

nataziel:
This PR fixes the (non-approx_constant related) Clippy failures in CI. Just a pure application of the suggested fixes


YeungOnion:
Thanks for this! Closer to easy to parse CI  results.

PR #363, opened 2025-11-17
feat (knn/kde) density estimation

geosarr:
# Multivariate density estimation

This PR implements density estimation from a sample using kd trees and nearest neighbors approach.

Bandwith is optionally provided by the caller.

### DIFF
--- a/Cargo.lock.MSRV
+++ b/Cargo.lock.MSRV
@@ -334,6 +334,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "kdtree"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


YeungOnion:
This looks good, thanks for implementing!
As for knn wrapper on the tree and errors this seems good by me.

My only question is if you have some ideas on how you'd coordinate this with the Kernel trait to allow changing the kernel function from Gaussian?

geosarr:
I may generalize kde_pdf function by adding a `kernel` argument that implements Kernel trait. But I am not sure the output value converges to the true value if you change the gaussian kernel to another one, especially for multidimensional samples.

YeungOnion:
I moved the branch head to figure out some other changes.

YeungOnion:
time for a cool new feature!

PR #369, opened 2026-01-11
Add a MultivariateNormal constructor from the Cholesky decomposition

### LINKED ISSUE #351: Relax validation on `MultivariateNormal` or provide escape hatch
The [exact-equality symmetry validation](https://github.com/statrs-dev/statrs/blob/master/src/distribution/multivariate_normal.rs#L180) on `MultivariateNormal` is not appropriate for floating-point calculations, and is not required for the `Cholesky` decomposition from `nalgebra`. Here's a quick script to show an example:
```rust
use rand::Rng;

let mut exactly_sym = 0;
let mut almost_sym = 0;
let mut not_sym = 0;
let mut cholesky = 0;
let total = 10_000;
for _ in 0..total {
    // Matrix that is nicely behaved, even this doesn't always pass an exact equality check!
    let sym =
        nalgebra::debug::RandomSDP::new(nalgebra::U2, || thread_rng().r#gen::<f64>()).unwrap();
    assert_relative_eq!(
        sym.upper_triangle(),
        sym.lower_triangle().transpose(),
        epsilon = 1e-8
    );

    // Doesn't really matter how we construct this, we could even just use the identity matrix!
    let mut f = nalgebra::Matrix2::<f64>::identity();
    f[(0, 1)] = 1.0;

    // This should also be symmetric, by mathematical definition
    let also_sym = f * sym * f.transpose();
    if also_sym.upper_triangle() != also_sym.lower_triangle().transpose() {
        if also_sym.upper_triangle().relative_eq(
            &also_sym.lower_triangle().transpose(),
            1e-10,
            1e-10,
        ) {
            almost_sym += 1;
        } else {
            not_sym += 1;
        }
    } else {
        exactly_sym += 1;
    }
    match nalgebra::Cholesky::new(also_sym) {
        Some(_) => {
            cholesky += 1;
        }
        None => {}
    }
}
println!("Almost symmetric: {almost_sym}/{total}");
println!("Exactly symmetric: {exactly_sym}/{total}");
println!("Cholesky ok: {cholesky}/{total}");
assert_eq!(not_sym, 0);
```
Here are the results:
```shell
Almost symmetric: 2059/10000 # Notice that these two
Exactly symmetric: 7941/10000 # values sum to 10000
Cholesky ok: 10000/10000
```

It seems to me that it would be more appropriate for `MultivariateNormal` to either check for approximate equality, or simply pass through `nalgebra`'s assumption that the covariance is (sufficiently) symmetric. Alternatively, a constructor that allows users to pass in a `Cholesky` struct directly would also suffice.

akern40:
This is a new feature to close #351. We introduce the `new_from_cholesky` function to allow users to construct a `MultivariateNormal` from just the Cholesky decomposition. We also create two helper functions to let us delegate the checks and final construction.

For now, we do not allow users to pass _both_ a Cholesky decomposition and a covariance matrix. Although this might save a matrix multiplication (to recover the covariance from the Cholesky decomposition), it raises questions about how strictly to check the compatibility of the covariance and the decomposition.

### DIFF
--- a/src/distribution/multivariate_normal.rs
+++ b/src/distribution/multivariate_normal.rs
@@ -153,6 +153,93 @@ impl MultivariateNormal<Dyn> {
     }
 }
 
+/// Check that a covariance is square, perfectly symmetric, and non-NaN
+fn check_cov<D>(cov: &OMatrix<f64, D, D>) -> Result<(), MultivariateNormalError>
+where
+    D: DimMin<D, Output = D>,
+    nalgebra::DefaultAllocator: nalgebra::allocator::Allocator<D>
+        + nalgebra::allocator::Allocator<D, D>
+        + nalgebra::allocator::Allocator<D>,
+{
+    if !cov.is_square()
+        || cov.lower_triangle() != cov.upper_triangle().transpose()
+        || cov.iter().any(|f| f.is_nan())
+    {
+        Err(MultivariateNormalError::CovInvalid)
+    } else {
+        Ok(())
+    }
+}
+
+/// Check the mean, covariance, and cholesky decomposition for incompatibilities, and return all three.
+///
+/// Covariance and cholesky decomposition are computed from each other as necessary.
+///
+/// # Panics
+/// If both the `cov` and `cholesky` arguments are `None`; at least one must be `Some(_)`.
+fn normalize_constructor_arguments<D>(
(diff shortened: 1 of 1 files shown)


YeungOnion:
I like the decision to not support specifying both, thanks!

PR #374, opened 2026-04-18
update `rand` 0.9 to 0.10

qkniep:


YeungOnion:
Thanks for this, can you outline why the change to use qualified syntax? If it was around specifying motor function is called without using type resolution in a reader's head, a change to use RngExt at the call site would be accepted too.

PR #375, opened 2026-05-25
Removed redundant word "may"

Why-not-now:


YeungOnion:
Thanks for this! For a second there I saw the diff without looking at the context and thought this might have been a meme PR in some kind of way.

PR #377, opened 2026-06-05
Add clippy fixes

soumyasen1809:
Cleanup based on clippy complains

### DIFF
--- a/src/distribution/beta.rs
+++ b/src/distribution/beta.rs
@@ -193,6 +193,29 @@ impl ContinuousCDF<f64, f64> for Beta {
             beta::inv_beta_reg(self.shape_a, self.shape_b, x)
         }
     }
+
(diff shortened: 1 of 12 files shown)


YeungOnion:
Hey, since I'm catching up, I did a lot of toe-stepping, would you be willing to move this branch onto something that cherry picks just the `try_*` changes?

I've also wondering about migrating the trait API, so what do you think if this were instead a new trait that accepts a validated type? I think at some point it would be nice to have the old traits deprecated. That would move the error boundary to have input validation separate from failure in evaluating the behavior.

I've tossed around some ideas on a branch [here](https://github.com/YeungOnion/statrs/tree/experimental_api/src/experimental_api). Let me know if the API there is easier to reason around. I think the docs, tests, and doctests are pretty decent at expressing what user code would be.

YeungOnion:
thanks for your work, I put it into #391!

PR #378, opened 2026-07-01
fix: restore missing normalization of Dirichlet samples

### LINKED ISSUE #344: `Dirichlet::sample()` is incorrect
I noticed that #253 omitted the normalization step when refactoring the `sample()` method of the Dirichlet distribution. The `sum` local variable is not used after it is computed. For large dimensions of alpha, the magnitude of sampled vectors is noticeably shorter than those generated by statrs@0.17.1 or scipy.

andersk:
This was broken by commit a239ebe25006c985115bd1d127804a42413111d4 (#253, cc @YeungOnion).

- Fixes #344.

### DIFF
--- a/src/distribution/dirichlet.rs
+++ b/src/distribution/dirichlet.rs
@@ -209,7 +209,7 @@ where
                 sum += sample;
                 sample
             }),
-        )
+        ) / sum
     }
 }
 


YeungOnion:
Ah of course! Thanks for the fix.

YeungOnion:
If you would, another PR with a test for this would be great. You wouldn't even need fixed seed for this one since it's a constraint for each output.

andersk:
- #379

PR #381, opened 2026-07-08
fix: compute Multinomial pmf/ln_pmf in log space to avoid NaN for large n

### LINKED ISSUE #370: Multinomial PMF likely converts back from ln to exp too early
The current implementation of `Multinomial::pmf()` computes the multinomial coefficient via `factorial::multinomial()`, which itself computes the logarithm of the multinomial coefficient then gets back to the coefficient via an exponential. The result is then multiplied by the product of probabilities elevated at the appropriate power.

The problem with this implementation is that because the multinomial coefficient is `n!` divided by factorials with much smaller arguments, the coefficient grows quickly enough with `n` to overflow even an `f64`. When that happens `factorial::multinomial` will overflow to `inf`, the product of small probabilities will underflow to `0.0`, and therefore the product of those two will be `NaN`.

This outcome can be avoided by computing the logarithm of the multinomial coefficient, subtracting from it the sum of the logarithms of the probabilities multiplied by their occurence count (which I suspect is also likely to be significantly more precise than elevating probabilities to a huge power, at the risk of underflowing to zero), and only then using the exponential to get to the final result.

But I'm not confident enough in my applied math skills to be 100% sure that this way of computing is numerically stable enough. I only know enough to know that computing the exponential of a sum with terms of opposite signs in it should be approached with a lot of care. So if someone around here feels more confident in their floating-point game, a theoretical cross-check of this approach is more than welcome before I proceed with submitting a PR that changes to that implementation :)

teddytennant:
## What

`Multinomial::pmf` and `Multinomial::ln_pmf` now compute the distribution entirely in log space, so valid large-`n` inputs return finite values instead of `NaN`/`+inf`.

- `pmf` delegates to `self.ln_pmf(x).exp()` (the length/sum guards are preserved: `ln_pmf` still panics on a length mismatch and returns `NEG_INFINITY` when the counts don't sum to `n`, and `NEG_INFINITY.exp() == 0.0`).
- `ln_pmf` builds the multinomial coefficient directly with `factorial::ln_factorial` (`ln n! - Σ ln x_i!`) instead of `factorial::multinomial(...).ln()`.
- Zero-count categories are skipped in the probability sum so an allowed zero probability no longer produces `0 * ln(0) = NaN`.

## Why

`factorial::multinomial` returns the coefficient itself (it exponentiates the stable log value before returning), which overflows `f64` to `+inf` once the log exceeds ~709.78. For `n = 2000`:

- `pmf` computed `inf * prod(p_i^x_i)`, and since the probability product simultaneously underflows to `0.0`, the result was `inf * 0.0 = NaN`.
- `ln_pmf` computed `inf.ln() = +inf`.

Both are wrong for perfectly valid inputs; SciPy returns finite values for the same parameters. Fixes #370.

## Testing

Added `test_pmf_large_n_no_overflow` (n=2000, p=[0.5, 0.5], x=[1000, 1000]), which asserts the pmf is finite and matches the lgamma-computed reference values (pmf ≈ 0.0178390111, ln_pmf ≈ -4.0263675824). Verified it fails on the previous implementation (`assert!(p.is_finite())` panics) and passes after the fix. The existing `test_pmf` cases still pass under the log-space implementation.

```
$ CARGO_BUILD_JOBS=2 cargo test -p statrs multinomial
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 716 filtered out
```

`cargo clippy -p statrs --all-targets -- -D warnings` is clean; `cargo fmt --check` reports no diffs in the touched file (the only fmt diffs on this tree are pre-existing in `geometric.rs`).

### DIFF
--- a/src/distribution/multinomial.rs
+++ b/src/distribution/multinomial.rs
@@ -303,19 +303,7 @@ where
     /// `x_i` is the `i`th `x` value, and `k` is the total number of
     /// probabilities
     fn pmf(&self, x: &OVector<u64, D>) -> f64 {
-        if self.p.len() != x.len() {
-            panic!("Expected x and p to have equal lengths.");
-        }
-        if x.iter().sum::<u64>() != self.n {
-            return 0.0;
-        }
-        let coeff = factorial::multinomial(self.n, x.as_slice());
-        coeff
-            * self
-                .p
-                .iter()
-                .zip(x.iter())
-                .fold(1.0, |acc, (pi, xi)| acc * pi.powf(*xi as f64))
+        self.ln_pmf(x).exp()
     }
 
     /// Calculates the log probability mass function for the multinomial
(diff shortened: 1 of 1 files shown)


YeungOnion:
Hey, this is great! Really appreciate this!

Issue #373, opened 2026-03-31
Inverse gamma cdf numerical problems


benjamin-lieser:
The inverse gamma cdf introduced in #227 does not work on all inputs.

For example
```
let gamma = Gamma::new(0.18,0.18).unwrap();
gamma.inverse_cdf(0.25);
```
we get `Nan`

I see that this is numerically tricky, but for example scipy has no problem giving the correct answer of `0.0016167`, even in way more extreme cases is still gives finite answers.

I will try looking into this myself, but maybe someone has more experience with this. It seems the newton iterations are not working anymore. I also don't know in which range the cdf function is correctly working.

benjamin-lieser:
If you are fine with adding `spec_math` as a dependency there is the `gdtri ` which seems to be stable.

day01:
@benjamin-lieser i had similar problem, so it helps me:

https://github.com/statrs-dev/statrs/pull/364/changes

Whatsonyourmind:
The Newton iteration for the inverse CDF (quantile function) of the Inverse Gamma distribution fails because the CDF is extremely flat near zero for small shape parameters like \`alpha = 0.18\`. At \`x = 0.0016\`, the CDF gradient is essentially zero in float64, so Newton's step \`x_{n+1} = x_n - (CDF(x_n) - p) / PDF(x_n)\` overshoots wildly.

Two approaches that work:

**1. Bisection fallback**: Newton converges quadratically when it works, but bisection always converges. A hybrid strategy (Newton with bisection bounds) is robust:

\`\`\`rust
// Start with bracket [lo, hi] from domain knowledge
let mut lo = 0.0_f64;
let mut hi = inverse_gamma_rough_upper_bound(alpha, beta, p);
for _ in 0..100 {
    let mid = (lo + hi) / 2.0;
    if cdf(mid) < p { lo = mid; } else { hi = mid; }
    // Switch to Newton when close enough
    if (hi - lo) < 1e-10 { break; }
}
\`\`\`

**2. Exploit the Gamma relationship**: Since \`InvGamma(alpha, beta)\` is the reciprocal of \`Gamma(alpha, 1/beta)\`, you can compute: \`inverse_cdf_invgamma(p) = 1.0 / inverse_cdf_gamma(1.0 - p)\`. If your Gamma quantile function is already stable (scipy's \`gdtri\` is), this is a one-line fix.

[...]

PR #382, opened 2026-07-16
fix: safeguard Gamma::inverse_cdf Newton step to avoid NaN for shape < 1

gaoflow:
## Summary

`Gamma::inverse_cdf(p)` returns `NaN` for any shape < 1 once `p` is small enough. `ChiSquared` (df < 2) and `Erlang` delegate to the same solver and inherit it — this is the NaN reported in #361 and #373.

```rust
Gamma::new(0.5, 1.0).unwrap().inverse_cdf(0.01);    // NaN — scipy: 7.8544e-5
Gamma::new(0.18, 0.18).unwrap().inverse_cdf(0.25);  // NaN — scipy: 1.6167e-3   (#373)
ChiSquared::new(1.0).unwrap().inverse_cdf(0.01);    // NaN — scipy: 1.5709e-4   (#361)
```

## Root cause

`inverse_cdf` builds a valid bracket `[low, high]` by bisection and then refines it with an *unguarded* Newton–Raphson loop. For shape < 1 the true quantile sits extremely close to 0 while the post-bisection iterate is still far from it, so the first Newton step overshoots to `x < 0`, outside the support. There `cdf(x) == 0` and `pdf(x) == 0`, so the update is `(0 - p) / 0 = -inf`; the next step is `inf - inf = NaN`, which propagates because the loop never re-brackets. The bracket invariant from the bisection phase is discarded by the Newton loop.

## Fix

Merge the two phases into one safeguarded loop (Numerical Recipes' `rtsafe`): keep `[low, high]` as an invariant, tighten it by the sign of `cdf(x) - p` each step, and accept a Newton step only when it is finite and strictly inside the bracket, otherwise take a bisection step. The iterate can no longer leave the support, so the result is always finite; shape ≥ 1 results are unchanged.

I diff-tested the whole `inverse_cdf` surface against scipy.stats / mpmath (dps=60) to scope this: the Gamma family (Gamma/ChiSquared/Erlang) was the only one producing NaN. Beta/StudentsT/FisherSnedecor go through the separately-safeguarded AS 64 `inv_beta_reg`, and Chi/InverseGamma use the bracketed default search — none share this failure — so the fix stays in the Gamma solver.

## Tests

Added regression tests in `gamma.rs` (all NaN/panic before the fix): shape < 1 quantiles are finite; values checked against scipy/mpmath constants; round-trip `cdf(inverse_cdf(p)) ≈ p` and monotonicity across the shape/`p` grid; and ChiSquared/Erlang delegation, including the large-df upper-tail case from #361. Full test suite, `cargo fmt --check` and `cargo clippy` are clean.

### DIFF
--- a/src/distribution/gamma.rs
+++ b/src/distribution/gamma.rs
@@ -194,7 +194,7 @@ impl ContinuousCDF<f64, f64> for Gamma {
             return self.max();
         };
 
-        // Bisection search for MAX_ITERS.0 iterations
+        // Bracket the quantile so that `cdf(low) <= p <= cdf(high)`.
         let mut high = 2.0;
         let mut low = 1.0;
         while self.cdf(low) > p {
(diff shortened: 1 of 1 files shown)


YeungOnion:
I like the solution, we might benefit from this kind of search for other inverse cdf steps for float domains. Thanks for this!

PR #384, opened 2026-07-17
Update nalgebra to 0.35 and add opt-in wasm_js support

day01:
## Motivation

I encountered this while using `statrs` in a Leptos application targeting
`wasm32-unknown-unknown`.

Seeded sampling should not require a host entropy source, while browser
applications using APIs such as `rand::rng()` need an explicit JavaScript
entropy backend.

This PR now builds on the `rand 0.10` migration already present in `main` and
focuses on separating those two WebAssembly use cases.

## Changes

- Update `nalgebra` from 0.34 to 0.35.
- Use `nalgebra/rand-no-std` from the `rand` feature instead of enabling
  nalgebra's full `rand` feature bundle.
- Add the opt-in `wasm_js` feature, enabling:
  - `rand/thread_rng`
  - `getrandom/wasm_js`
- Add CI checks for:
  - `wasm32-unknown-unknown` with seeded RNG support
  - `wasm32-unknown-unknown` with `wasm_js`
  - `wasm32-wasip1`
- Document when applications should enable `wasm_js`.
- Fix the order statistics benchmark so it measures valid inputs.
- Cover the geometric inverse CDF overflow guards.
- Increase MSRV from 1.87 to 1.89, as required by `nalgebra 0.35`.

## Validation

- `cargo fmt -- --check`
- clippy on Rust 1.97 with warnings denied
- default, no-default-features plus `rand`, and all-features tests
- all 11 feature combinations
- `wasm32-unknown-unknown` and `wasm32-wasip1` checks
- MSRV 1.89 with the committed MSRV lockfile
- order statistics benchmark smoke test

## Acknowledgements

Thanks to @RobertJacobsonCDC for the earlier `rand 0.9` migration in #331 and
to @qkniep for the `rand 0.10` work in #374. Those contributions and the
follow-up changes now present in `main` helped shape the final, smaller scope of
this PR.

### DIFF
--- a/.github/workflows/autofix-ci.yml
+++ b/.github/workflows/autofix-ci.yml
@@ -8,7 +8,7 @@ permissions:
   contents: read
 
 env:
-  rust_clippy: 1.87
+  rust_clippy: 1.89
 
 jobs:
   autofix:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,11 +90,5 @@ jobs:
       - name: Check WebAssembly without a host RNG
         run: cargo check --target wasm32-unknown-unknown --no-default-features --features std,nalgebra,rand --lib
 
-      - name: Check WebAssembly with the JavaScript RNG backend
-        run: cargo check --target wasm32-unknown-unknown --all-features --lib
-
       - name: Check WASI without a host RNG
         run: cargo check --target wasm32-wasip1 --no-default-features --features std,nalgebra,rand --lib
-
-      - name: Check WASI with all features
-        run: cargo check --target wasm32-wasip1 --all-features --lib
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,3 +76,25 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - name: Check all possible feature sets
         run: cargo hack check --feature-powerset --no-dev-deps
+
+  wasm:
+    needs: [clippy, fmt]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust stable with WebAssembly targets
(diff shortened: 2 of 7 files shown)


day01:
Previously, both benchmarks almost always measured the fast path returning `NaN` instead of the intended selection algorithms.

day01:
I don't think so coverage on display is necessary, but if you req that @YeungOnion i will add...

day01:
@YeungOnion  Rebased onto current main and aligned with the upstream rand 0.10 changes. The PR now focuses on opt-in wasm_js support

YeungOnion:
Thanks for tidying this up, wanted to get in all the other backlogged work from other contributors. I think enabling more targets is useful, but I'm not as familiar with the perks of rustup and the build system rust offers as I'd like to be - so anything you can clear up would benefit me.

Tell me where my understanding is lacking: this change allows us to specify the backend that rand uses by setting the feature for `getrandom`. We wouldn't need this feature for other targets because... I don't know, but it would create mutually exclusive features which might be unnecessary.

If my understanding is sufficient, I'm wondering why we don't share the approach that [`rand` is taking](https://rust-random.github.io/book/crate-platforms.html#webassembly), which reads like, "our code is backend agnostic because a choice of dependency; at times you'll need to specify that backend yourself".

day01:
You’re right, I missed that. I am removing the passthrough feature and move the configuration to the application README example. .... thx @YeungOnion

day01:
From large change to small contribution :)

----
`nalgebra` now uses `and-no-std` so it will work with stanadrad mechanism as you said 

  all MSRV/WASM/WASI checks pass...

YeungOnion:
takes a lot of intention to solve a problem with [no code](https://github.com/kelseyhightower/nocode) I think it's a good one. Obviously, if we get more issues regarding WASM support, then a feature would make it more discoverable.

PR #390, opened 2026-07-18
Converge the default continuous inverse_cdf instead of a fixed 16 iterations

gaoflow:
The `ContinuousCDF::inverse_cdf` default runs a fixed 16-iteration bisection with no
convergence test, so it only ever narrows the initial bracket by 2^16 regardless of how
wide the bracket is or how small the quantile is. Every continuous distribution without a
closed-form quantile inherits this — currently `Chi` and `InverseGamma` — and the result
sits far from the crate's own `DEFAULT_RELATIVE_ACC` (1e-14):

- `Chi::new(1).inverse_cdf(1e-12)` returns `3.05e-5` instead of `1.25e-12`; in fact every
  `p <= ~1e-6` returns the same `3.05e-5` bracket floor. Even the median is only accurate
  to ~1e-5.
- `InverseGamma` shows the same ~1e-5 baseline error, growing into the tails.

This is the continuous mirror of the discrete side, which already converges via
`internal::integral_bisection_search`.

The fix replaces the fixed count with a bisection that runs until the bracket agrees to
`DEFAULT_RELATIVE_ACC` (capped at 100 iterations and by float resolution). Two smaller
changes make the tails accurate:

- Bracket from the distribution's own `min()`/`max()` when finite, only doubling out from
  ±2 when a bound is infinite. `Chi`/`InverseGamma` are supported on `[0, ∞)`, so the old
  `[-2, 2]` seed wasted the lower half.
- In the upper half, invert `sf` rather than `cdf`. As `cdf` saturates to one it can no
  longer place the quantile (many `x` map to a single `f64` cdf value); `sf` stays well
  conditioned. Distributions that don't override `sf` fall back to `1 - cdf`, so this is a
  no-op for them.

Across `p ∈ [1e-12, 1-1e-12]` the worst relative error for `Chi` and `InverseGamma` drops
from ~1e7 (and ~1e-5 at the median) to `<= 1.3e-14`. Reference quantiles in the new tests
come from scipy, cross-checked against mpmath at 60 digits; a round-trip test covers a
wider grid.

```rust
let q = Chi::new(1).unwrap().inverse_cdf(1e-12); // was 3.05e-5, want 1.25e-12
```

### DIFF
--- a/src/distribution/chi.rs
+++ b/src/distribution/chi.rs
@@ -513,4 +513,43 @@ mod tests {
         density_util::check_continuous_distribution(&create_ok(2), 0.0, 10.0);
         density_util::check_continuous_distribution(&create_ok(5), 0.0, 10.0);
     }
+
+    #[test]
+    fn test_inverse_cdf_reference() {
+        // Chi has no closed-form inverse_cdf, so it exercises the ContinuousCDF
+        // default. References are scipy.stats.chi.ppf, verified against mpmath
+        // (dps=60) to rel err < 4e-15; both tails are covered because the old
+        // fixed-iteration search collapsed there (e.g. k=1, p=1e-12 was off by ~2e7).
+        let cases: &[(u64, f64, f64)] = &[
+            (1, 1e-12, 1.253314137315499e-12),  (1, 1e-8, 1.2533141373155e-8),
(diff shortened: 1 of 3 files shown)


YeungOnion:
Thanks!

would you be willing follow this with something similar to the brent-like approach that uses NR steps that you introduced for Gamma in #382?

Issue #366, opened 2025-12-22
[question]: Some functions can panic


nhatcher:
I see some functions can panic. For instance:

https://github.com/statrs-dev/statrs/blob/master/src/distribution/beta.rs#L191

Is that something that might change in the future or that is a design decision?

This makes things a bit difficult for us. See
https://github.com/ironcalc/IronCalc/issues/654

Apologies if this is not the appropriate channel and since I am here, congratulations on a great crate!

YeungOnion:
This is a good channel for this feedback - I've just been fairly absent as of late.

Previously, the DX requiring a newtype as an argument (just an f64 that implements `num::Bounded`) to these functions didn't seem good and after it would have led to breaking change for the inverse_cdf API which seems to be one of the most used parts of the library.

Since you wrap this functionality for multiple distributions, I'd love to get your take on how you'd _like_ to consume the API. Are you binding each callable or are you binding to the `ContinuousCDF` trait?

nhatcher:
> This is a good channel for this feedback - I've just been fairly absent as of late.
No problem!

> I'd love to get your take on how you'd _like_ to consume the API

I don't really have a preference, as long as it cannot panic.

> Since you wrap this functionality for multiple distributions, I'd love to get your take on how you'd like to consume the API. Are you binding each callable or are you binding to the `ContinuousCDF` trait?

This is an example of how we use it presently:
https://github.com/ironcalc/IronCalc/blob/main/base/src/functions/statistical/chisq.rs

soumyasen1809:
> I'd love to get your take on how you'd _like_ to consume the API

I am not related to either of the projects. But  maybe `statrs` can explicitly forbid panic in source code (allow for tests only) using clippy checks. Then all the panics currently in use can be converted to return an Error type.

YeungOnion:
I've still not been making time for statrs but the part I've wondered about
is how to change this API in a way that's non breaking.
Right now it's been more of a panic-first API instead of relying on a
dynamically validated type at the API, but it's clearly not working.

Right now we use f64 to represent probability, but we could use a new type
that provides
- panicking From<f64> no API breakage, would panic in all the same
scenarios where callers try to evaluate with probabilities outside the
interval.
- or TryFrom<f64> and change the existing API (breaking or extending with
_safe)

soumyasen1809:
> Right now it's been more of a panic-first API instead of relying on a

https://github.com/statrs-dev/statrs/pull/377/changes

[...]

PR #391, opened 2026-07-19
feat(distribution): add non-panicking try_inverse_cdf

YeungOnion:
start to expose fallible API for ContinuousCDF responding to #366

cherry-picked from #377


YeungOnion:
@soumyasen1809 I narrowed to use an error enum instead of a trait object on the error.

Issue #376, opened 2026-05-29
Numerical Stability of Variance


genos:
First off, thank you for `statrs`! This looks like a fantastic library that took a lot of effort.

Regarding the numerical stability of the functions in `statrs::statistics`, I think it would be useful to reach for [Welford's algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm) for reasons outlined [here](https://www.johndcook.com/blog/2008/09/26/comparing-three-methods-of-computing-standard-deviation/) by John D. Cook and implemented [here](https://www.johndcook.com/skewness_kurtosis.html). It could also improve performance; the original algorithm requires only a single pass over the data, and it can also be parallelized.

For example, running the following on my laptop

<details>
<summary>Example Code</summary>

```rust
//! Example from <https://www.johndcook.com/blog/2008/09/26/comparing-three-methods-of-computing-standard-deviation/>
use rand::prelude::*;
use statrs::statistics::Statistics;

/// <https://www.johndcook.com/skewness_kurtosis.html>
#[derive(Default)]
struct RunningStats(f64, f64, f64);

impl RunningStats {
    fn push(&mut self, x: f64) {
        let n1 = self.0;
        self.0 += 1.0;
        let delta = x - self.1;
        let delta_n = delta / self.0;
        let term1 = delta * delta_n * n1;
        self.1 += delta_n;
        self.2 += term1;
    }
    fn variance(&self) -> Option<f64> {
        if self.0 < 2.0 {
            None
        } else {
            Some(self.2 / (self.0 - 1.0))
        }
    }
}

impl<T: Iterator<Item = f64>> From<T> for RunningStats {
    fn from(xs: T) -> Self {
        let mut r = Self::default();
        for x in xs {
            r.push(x);
        }
        r
    }
}

fn main() {
    let mut rng = SmallRng::seed_from_u64(30_613_700); // seed from random.org
    let mut xs = vec![0f64; 1_000_000];
    // John D. Cook's second example: samples ~ U([0, 1]), shifted by 10^12
    for x in &mut xs {
        *x = 1e12 + rng.random::<f64>();
    }
    println!(
        "statrs:       {}\nRunningStats: {}",
        Statistics::variance(&xs),
        RunningStats::from(xs.into_iter())
            .variance()
            .expect("nonempty")
    );
}
```
</details>

yields

```
statrs:       0.31179242499195364
RunningStats: 0.08339059221218162
```

Thanks again!

[...]

Issue #324, opened 2025-02-03
Reductor pattern to compute multiple statistics in one iterator pass.


Evian-Zhang:
It is very common when we need multiple statistics of one data (for example, we may need both min and stddev at same time). However, current API design of statrs only allows us to compute those statistics by iterating the same data multiple times, which may be unnecessary for some statistics.

There are already some discussions in the rust forum ([How to design a composable iterator adapter/consumer API?](https://internals.rust-lang.org/t/how-to-design-a-composable-iterator-adapter-consumer-api/20923), [Adding Iterator::extrema, argmin, argmax](https://internals.rust-lang.org/t/adding-iterator-extrema-argmin-argmax/19599)), and some one has proposed the crate [reductor](https://crates.io/crates/reductor), which seems be able to solve this problem. However, the statistics provided by that crate is far less than the statrs crate.

Would you integrate the reductor pattern into statrs to solve this problem?

YeungOnion:
If I understand this correctly, we would only need to define the recursive cases of the statistics we have, then we would match the API needed for a reductor pattern. This would allow some users to write the code using `fold`.

This would help with reducing the repeated computation, but regarding ergonomics, I don't have an idea on how to approach the reductor pattern since there's not a notion of an identity element (the starting element to reduce from) for the moments. And the higher the order the more inputs needed for it to be defined. 

Do you have any insight on how you might want this to work for an example like (mean, variance)?

---
Note:

There's an approach to this sample streaming for all moments, [Wiki Algorithms for Calculating Variance, Parallel Algorithms](https://en.m.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm). Welford's algorithm is the single sample for the first two moments; also on that same article.

YeungOnion:
I don't want to propose an idea instead of your, but I don't wish to forget mine,

<details>
  <summary>An approach that's not affecting your input before you share it</summary>
  
  
Additionally, the update requires the lower order moments, so this could be a const generic case for expressing, "I want the first K moments from an iterator of numbers",
```rs
let count, mean, m2, m3 = reduce_to_moments<4>(data.iter());
```

or alternatively,
```rs
let count, mean, variance, skewness, kurtosis = Moment::standardize(reduce_to_moments<5>(data.iter()));
```

</details>

Evian-Zhang:
Thank you very much for your kind response!

> Do you have any insight on how you might want this to work for an example like (mean, variance)?

[...]

PR #394, opened 2026-07-19
online moments

YeungOnion:
- **feat: add OnlineMoments accumulator using Welford**
- **fix: wrap online stats with iter_statistics for #376**
- **fix: drop excess precision requirements**

### DIFF
--- a/src/statistics/accumulate.rs
+++ b/src/statistics/accumulate.rs
@@ -1,0 +1,62 @@
+/// A type that accumulates `f64` observations one at a time.
+///
+/// Tuples up to arity 5 fan each observation into each accumulator, so
+/// multiple statistics can share a single pass - the pattern is more
+/// useful when the online stats aren't all moments, but example below:
+///
+/// ```
+/// use statrs::statistics::{Accumulate, OnlineMean, OnlineVariance};
+///
+/// let data = [3.0_f64, -1.0, 4.0, 1.0, -5.0];
+/// let (mean, var): (OnlineMean, OnlineVariance) =
+///     data.iter().copied().fold(Default::default(), Accumulate::push);
+///
+/// assert!(mean.mean().is_some());
+/// assert!(var.variance().is_some());
+/// ```
+pub trait Accumulate: Default + Sized {
+    fn push(self, x: f64) -> Self;
+}
+
+impl<A: Accumulate> Accumulate for (A,) {
+    fn push(self, x: f64) -> Self {
+        (self.0.push(x),)
+    }
+}
+
+impl<A: Accumulate, B: Accumulate> Accumulate for (A, B) {
+    fn push(self, x: f64) -> Self {
+        (self.0.push(x), self.1.push(x))
+    }
+}
+
+impl<A: Accumulate, B: Accumulate, C: Accumulate> Accumulate for (A, B, C) {
+    fn push(self, x: f64) -> Self {
+        (self.0.push(x), self.1.push(x), self.2.push(x))
+    }
+}
+
+impl<A: Accumulate, B: Accumulate, C: Accumulate, D: Accumulate> Accumulate for (A, B, C, D) {
+    fn push(self, x: f64) -> Self {
+        (
+            self.0.push(x),
+            self.1.push(x),
+            self.2.push(x),
+            self.3.push(x),
+        )
+    }
+}
+
+impl<A: Accumulate, B: Accumulate, C: Accumulate, D: Accumulate, E: Accumulate> Accumulate
+    for (A, B, C, D, E)
+{
+    fn push(self, x: f64) -> Self {
+        (
+            self.0.push(x),
+            self.1.push(x),
+            self.2.push(x),
+            self.3.push(x),
+            self.4.push(x),
+        )
+    }
+}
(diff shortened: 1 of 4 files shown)


YeungOnion:
forgot to mention that this was likely only on my mind due to #324, so thanks for the feedback there.

Evian-Zhang:
Wow wonderful work! I can't wait to use this feature!

However, I have two questions:

1. Why `OnlineMean` is `OnlineMoments<2>` instead of `OnlineMoments<1>`? I'm not expert in statistics, is this a typo or intentional use?
2. It seems to me that `push` is actually called several times for each element since <https://github.com/statrs-dev/statrs/blob/10cf6d610c1b8febc0b0606a443eb5e9d6ac8b7d/src/statistics/accumulate.rs#L53-L61>. Moreover, if we write

   ```rust
   let (mean, var, skewness): (......) = data.iter().copied().fold(Default::default(), Accumulate::push);
   ```

   Then the mean value can be calculated by `mean.mean()`, `var.mean()` and `skewness.mean()`, which is a little bit redundant to me.

Based on your code, it occurs to me that we can really write declarative code like this:

```rust
    let (OnlineMean(mean)) = data.iter().copied().fold(Default::default(), Accumulate::push).get();
    // or
    let (OnlineMean(mean), OnlineVariance(variance)) = data.iter().copied().fold(Default::default(), Accumulate::push).get();
    // or
    let (OnlineVariance(variance), OnlineSkewness(skewness)) = data.iter().copied().fold(Default::default(), Accumulate::push).get();
    // then mean, variance, and skewness is Option<f64> directly
```

if we use the following type magic:

```rust
pub trait Moment {
    const ORDER: usize;
    // This method is not meant to be called from user, so we can
    // have less type restrictions. In theory, we can even restrict
    // MS only to the tuple types that include Self.
    fn from<MS: Moments>(acc: &Accumulate<MS>) -> Self;
}
pub trait Moments: Sized {
    // This will always be compiled into constants.
    // In the future, this can be directly an associated const.
    fn order() -> usize;
    fn from(acc: &Accumulate<Self>) -> Self;
}

impl<M1: Moment, M2: Moment> Moments for (M1, M2) {
    fn order() -> usize {
        std::cmp::max(M1::ORDER, M2::ORDER)
    }

    fn from(acc: &Accumulate<Self>) -> Self {
        (M1::from(acc), M2::from(acc))
    }
}
// Also implements for (M1), (M1, M2, M3), ...

pub struct Accumulate<MS: Moments> {
    count: u64,
    // Some type magics can be done here to make
    // the array length equal to MS::order(). But since
    // we only need up to three-order moment, a simple
    // [_; 3] is enough.
    m: [f64; 3],
    phantom: core::marker::PhantomData<MS>,
}

impl<MS: Moments> Default for Accumulate<MS> {
    fn default() -> Self {
        Self {
            count: 0,
            m: [0.0; 3],
            phantom: core::marker::PhantomData::default()
        }
    }
}

impl<MS: Moments> Accumulate<MS> {
    pub fn push(mut self, x: f64) -> Self {
        self.count += 1;
        let n = self.count as f64;

        // Welford / Pebay (2008) central moment update. Update order: M3
        // before M2 before mean; each step uses the previous observation's
        // lower-order accumulators.
        let delta = x - self.m[0];
        let delta_n = delta / n;
        let new_mean = self.m[0] + delta_n;
        let delta2 = x - new_mean;

        if MS::order() >=2  {
            let old_m2 = self.m[1];
            if MS::order() >=3 {
                let inc = delta * (delta_n * delta_n) * (n - 1.0) * (n - 2.0) - 3.0 * delta_n * old_m2;
                self.m[2] += inc;
            }
            self.m[1] += delta * delta2;
        }

        self.m[0] = new_mean;
        self
    }

    pub fn get(&self) -> MS {
        MS::from(self)
    }
}

pub struct OnlineMean(pub Option<f64>);
impl Moment for OnlineMean {
    const ORDER: usize = 1;
    fn from<MS: Moments>(acc: &Accumulate<MS>) -> Self {
        if acc.count == 0 {
            Self(None)
        } else {
            Self(Some(acc.m[0]))
        }
    }
}

pub struct OnlineVariance(pub Option<f64>);
impl Moment for OnlineVariance {
    const ORDER: usize = 2;
    fn from<MS: Moments>(acc: &Accumulate<MS>) -> Self {
        if acc.count < 2 {
            Self(None)
        } else {
            Self(Some(acc.m[1] / (acc.count - 1) as f64))
        }
    }
}

// Also implements for OlineSkewness, OnlineStdDev, OnlinePopulationVariance...
```

The code snippet above has been tested by me that is consistent to your implementation. The difference is that the usage code can be more declarative, and `push` is only called one time for each element.

Evian-Zhang:
@YeungOnion What do you think of this idea? I'm happy to send a PR after #408 is merged (otherwise that PR has to be refactored a lot).

YeungOnion:
yeah, something like that could be nice, I do think there's a way to surface that welford provides all lower order moments and doesn't need separate reducer states. I just merged #408 so you can take a stab at it. I do think part of it was my poor choice in names of type aliases for OnlineMoments<N>. I do like the destructure pattern instead of duplicating a call to `mean` method. But the `.get` finalize seems like a great call for this approach instead of delegating finalization to `.mean`

PR #396, opened 2026-07-20
refactor: Categorical now stores normalized values

YeungOnion:
Cherry-pick of 37f8fdf (FreezyLemon) onto current main. The original commit predates the rand 0.9/0.10 migration and the `std`/`core` split, so it conflicted on:

- `::rand::distr::Distribution` (vs. old `rand::distributions::Distribution` alias)
- `RngExt::random` (vs. old `rng.gen`)

`Categorical` now stores `norm_cdf`/`norm_sf` directly instead of normalizing on every `cdf`/`sf`/`inverse_cdf` call, and `inverse_cdf` uses `binary_search_by` instead of a hand-rolled `binary_index`. The now-unused `sample_unchecked`/`prob_mass_to_cdf`/`cdf_to_sf`/`binary_index` helpers are removed; `Multinomial::sample`, which depended on `categorical::prob_mass_to_cdf`/`sample_unchecked`, now computes its cdf inline instead.

## Test plan
- [x] `cargo build --all-features`
- [x] `cargo test --all-features` (192 passed)
- [x] `cargo clippy --all-features`

### DIFF
--- a/src/distribution/categorical.rs
+++ b/src/distribution/categorical.rs
@@ -22,8 +22,8 @@ use core::f64;
 #[derive(Clone, PartialEq, Debug)]
 pub struct Categorical {
     norm_pmf: Vec<f64>,
-    cdf: Vec<f64>,
-    sf: Vec<f64>,
+    norm_cdf: Vec<f64>,
+    norm_sf: Vec<f64>,
 }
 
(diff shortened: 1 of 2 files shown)


YeungOnion:
GoF tests for sampling will still be a good idea

PR #398, opened 2026-07-20
fix: Automate MSRV lockfile updates

day01:
## Motivation
update `Cargo.lock.MSRV` automatically in release-plz pull requests - doesnt work


YeungOnion:
When this works, will it be that the MSRV file will be updated each time this action runs? Or will it be something else?

day01:
Will be updated when action triggered. It regenerates Cargo.lock.MSRV, but commits and pushes it only if the file actually changed
```
if git diff --quiet -- Cargo.lock.MSRV; then
    exit 0
fi
```

PR #399, opened 2026-07-20
Add initial no_std support

day01:
## Motivation

`statrs` already declares `no_std` when the `std` feature is disabled, but the crate did not compile for a real bare-metal target.

The main blockers were:

- `approx` and `num-traits` enabling `std` through their default features
- floating-point methods unavailable through `core` on bare-metal targets
- error implementations depending on `std::error::Error`
- CI testing a hosted `no_std` configuration instead of a real bare-metal build

**It is not complete no_std feature. it is first step in the direction.**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,11 +59,24 @@ jobs:
       - name: Install Rust stable
         uses: dtolnay/rust-toolchain@stable
 
-      - name: Test no_std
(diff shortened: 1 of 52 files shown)


YeungOnion:
Thanks for this, I think when we started on it `error` wasn't out of `std`, the intent to separate `alloc` could be useful, long term might end up with very little `std` if at all.

Issue #400, opened 2026-07-20
no_std migration


day01:
## Motivation

The initial `no_std` work established a working bare-metal baseline for the core crate, but it did not make the entire API available without `std`.

Several modules were still gated behind `std` even though they primarily required allocation rather than operating-system functionality. The `nalgebra` feature also enabled `std` unconditionally.

## Scope

- [x] support core-only bare-metal builds
- [x] support floating-point operations through `libm`
- [x] migrate errors to `core::error::Error`
- [x] support `rand` without `std`
- [x] add bare-metal CI checks
- [x] move allocation-based APIs from `std` to `alloc`
- [x] migrate `Categorical`, `Empirical` and allocation-based statistics
- [x] make `nalgebra` work with `no_std + alloc`
- [x] verify representative calculations in a real `no_std` environment
- [x] document the supported feature combinations
- [ ] migrate KDE to `no_std + alloc` ([statrs#443](https://github.com/statrs-dev/statrs/pull/443), blocked by [mrhooray/kdtree-rs#92](https://github.com/mrhooray/kdtree-rs/pull/92))

### Known limitation

KDE remains `std`-only on `main`. [statrs#443](https://github.com/statrs-dev/statrs/pull/443) adds `no_std + alloc` support and is blocked by the merge and release of [mrhooray/kdtree-rs#92](https://github.com/mrhooray/kdtree-rs/pull/92). The draft currently pins the upstream Git revision and must switch to the released crates.io version before merge.

### PRs

- https://github.com/statrs-dev/statrs/pull/399
- https://github.com/statrs-dev/statrs/pull/403
- https://github.com/statrs-dev/statrs/pull/443

PR #403, opened 2026-07-20
Support alloc-backed APIs in no_std

day01:
## Motivation

This PR continues the work from [#400](https://github.com/statrs-dev/statrs/issues/400) by extending `no_std` support to APIs that require heap allocation.

Types such as `Vec` and `BTreeMap` are provided by Rust's `alloc` crate and do not inherently require `std`. Using `alloc` allows these APIs to work in `no_std`.

## Changes

- migrate statrs-owned heap allocations from `std` to `alloc`
- make `Categorical`, `Empirical`, `log_spaced`, ranks, and statistical tests available in `no_std`
- enable matrix-backed distributions with `nalgebra` using its `alloc` and `libm` support
- document allocator setup for `no_std` applications
- add bare-metal checks for core-only, `rand`, `nalgebra`, and `nalgebra,rand`
- add executable Wasm verification covering representative calculations

KDE cannot be migrated because its `kdtree` dependency requires `std`.

## Verification
executable `no_std` calculations on `wasm32-unknown-unknown`


YeungOnion:
think moving toward libm might be nice, there wasn't much in the way of standardization of special functions, and it's a good target for shared development.

day01:
@YeungOnion  Would you mind opening a follow-up issue outlining which functions you think should move to or be shared with `libm`?

It would be useful to distinguish the existing overlap ;]

day01:
@YeungOnion may we add it to next release?

PR #404, opened 2026-07-20
Fast Newton-Raphson inverse_cdf for Chi and InverseGamma

gaoflow:
Follows #390 / #382, per @YeungOnion's suggestion to extend the brent-like + Newton-Raphson `inverse_cdf` from Gamma (#382) to the distributions that still fall back to the generic default.

`Chi` and `InverseGamma` were the last two continuous distributions using the generic bisection default for `inverse_cdf` (the two named in #390). This gives each a custom solver in the same vein as Gamma.

## Approach

A shared `internal::newton_raphson_quantile(p, cdf, sf, pdf)` brackets the quantile to a factor of two, then refines it with safeguarded Newton steps (Numerical Recipes' `rtsafe`): `[low, high]` is only ever tightened, and a Newton step that is non-finite or leaves the bracket falls back to bisection. `Chi` and `InverseGamma` call it with their own cdf/sf/pdf.

Two things were needed to keep it accurate and fast across the whole range, not just the middle:

- **Relative convergence.** #382's `prec::convergence` folds in an absolute `1e-9` tolerance, which is meaningless when the quantile itself is ~`1e-12`, so the solver tests the relative Newton step instead. It also checks the raw Newton step for convergence *before* tightening the bracket — otherwise a converged step that rounds onto the endpoint just moved to `x` is rejected and the search bisects the whole span back (this showed up as ~30 wasted iterations near the median).
- **`sf` in the upper half.** As `cdf` saturates to one it can no longer resolve a deep upper-tail quantile (`InverseGamma`'s reaches ~`6e23` at `p = 1 − 1e-12`); inverting `sf` there keeps it well conditioned, as in #390.

## Validation

Checked against `scipy.stats`, cross-checked with mpmath at 60 digits, over `p ∈ [1e-12, 1 − 1e-12]` for `Chi(k)`, k ∈ {1,2,3,5,10} and `InverseGamma(a,b)`, (a,b) ∈ {(1,1),(0.5,0.5),(2,1),(3,2),(5,3)}:

- worst-case relative error `3.4e-15`; round-trip `cdf(inverse_cdf(p)) ≈ p` to ~`3e-15`.
- converges in a handful of Newton steps (medians ~5, vs ~48 fixed bisection steps) — roughly 60% fewer cdf/pdf evaluations across the grid, with the bisection fallback preserving the default's worst-case robustness.

The `test_inverse_cdf_reference` / `test_inverse_cdf_round_trip` cases added in #390 (scipy/mpmath values pinned in both tails) now exercise the new solver and still pass; a new `internal::test_newton_raphson_quantile` covers the shared routine directly against the Exponential closed form. `cargo test`, `cargo fmt --check` and `cargo clippy` are clean.

Gamma keeps its own #382 solver here; happy to fold it into this helper too as a follow-up if you'd prefer a single routine.

### DIFF
--- a/src/distribution/chi.rs
+++ b/src/distribution/chi.rs
@@ -148,6 +148,32 @@ impl ContinuousCDF<f64, f64> for Chi {
             gamma::gamma_ur(self.freedom() as f64 / 2.0, x * x / 2.0)
         }
     }
+
+    /// Calculates the inverse cumulative distribution function for the chi
+    /// distribution at `p`, i.e. the `p`-quantile.
+    ///
+    /// # Panics
+    ///
+    /// If `p` is not in `[0, 1]`.
+    fn inverse_cdf(&self, p: f64) -> f64 {
+        if !(0.0..=1.0).contains(&p) {
+            panic!("p must be in [0, 1]")
+        }
+        if p == 0.0 {
+            return self.min();
+        }
+        if p == 1.0 {
+            return self.max();
+        }
+        // The chi cdf has no closed-form inverse; solve it with the shared
+        // safeguarded Newton search instead of the generic bisection default.
+        super::internal::newton_raphson_quantile(
+            p,
+            |x| self.cdf(x),
+            |x| self.sf(x),
+            |x| self.pdf(x),
+        )
+    }
 }
 
 impl Min<f64> for Chi {
--- a/src/distribution/chi.rs
+++ b/src/distribution/chi.rs
@@ -579,4 +579,23 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn test_inverse_cdf_p0_p1() {
+        let d = create_ok(3);
+        assert_eq!(d.inverse_cdf(0.0), d.min());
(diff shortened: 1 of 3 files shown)


binggao1230:
Chasing the uncovered lines turned up a real defect rather than a coverage gap: the safeguarded step rejected a Newton iterate only when it was non-finite or left the bracket, never when it simply was not making progress, so on the inverse gamma's exp(-rate/x) lower tail (each step advances rate/x by one however far the root is) the search ran out of iterations short of the root — InverseGamma(1, 1).inverse_cdf(1e-200) came back 4.3% high and InverseGamma(1, 400).inverse_cdf(1e-300) 9.0% high, both of which the bisection this replaces got right. Restoring rtsafe's second test, take the Newton step only while it also shrinks at least as fast as a bisection, fixes those; nine deep-tail cases across four parameter sets now match mpmath at dps=60 to under 2e-16, and with tests added for the p == 0, p == 1 and out-of-range arms patch coverage is 100%.

YeungOnion:
Hey this is great! Thanks a bunch!

Issue #405, opened 2026-07-21
Implementing Kolmogorov Smirnov Distribution and Test


AshrafIbrahim03:
I was looking for distribution similarity testing in this crate, but noticed there really isn't any. I found issues #288 and #39 which both mention KS testing. It seems like there hasn't been any movement with that in a few years. #39 mentions that the implementation of KS was too slow. I found an implementation in [this paper](https://doi.org/10.18637/jss.v039.i11) that I could translate into Rust and it seems reasonably fast. I'm pretty sure I tracked down this implementation from what they use in scipy.

I really want to use a KS test in a project of mine and was hoping to contribute to this in some way!

YeungOnion:
I'd definitely accept a contribution for something like this, if you can implement the two sided test statistics with two iterators over univariate float data, that would go a long way as a standalone function. The only sharp edge to consider is licensing, please don't study copyleft material.

AshrafIbrahim03:
What do you mean by the last sentence? Would there be something wrong with me translating the algorithm to Rust?

YeungOnion:
It was more just a general notice because licensing can sometimes be a concern as you look for reference implementations. 

Scipy is BSD licensed which is very permissive, so you can reference and modify as needed without issue, and that would include a rust translation. The paper you linked would also be fair game.

AshrafIbrahim03:
Sounds good. The code itself has a GPL license, so I don't think there's any problem there! 

I want to test for parity between my translation and the original version. Given that there's two original files (one in C and one in Java). Is there a preferred method of testing for parity between my translation and the original code? I know that Rust can natively interface with C(which would make for easy test writing), but I'm translating from the Java code because it's more readable for me.

Is there a place to add tests, not to verify statistical test outputs(like it seems in the tests directory), but to verify parity between statrs and original implementations?


Also, when going through the stats_tests directory, I noticed that there's already a ks_test file that references the same paper that I'm translating from AND there's already implementation for one sided and two sided tests. For clarification, in your initial message in this thread, did you mean that contributing a stand alone function would be the only contribution right now because I was going into this thinking I was going to implement the KS distribution and tests for statrs.

YeungOnion:
I forgot that was there. Maybe I was recalling how it wasn't yet in use for gof testing...? But I mentioned the piece around what to build as a first step to compose nicer API. The distribution itself would be net new, but I also think we're lacking discoverability on the tests. Contributions on either are welcome. 

[...]

PR #406, opened 2026-07-22
Refactored ks_twosample to take in two slices instead of a mutable owned Vec<f64>

AshrafIbrahim03:
Through some discussion in #405 , it seemed like refactoring ks_twosample was needed. Basically just changed the function signature from:
```rust
pub fn ks_twosample(
    mut data1: Vec<f64>,
    mut data2: Vec<f64>,
    method: KSTwoSampleAlternativeMethod,
    nan_policy: NaNPolicy,
) -> Result<(f64, f64), KSTestError>
```

to

```rust
pub fn ks_twosample(
    data1: &[f64],
    data2: &[f64],
    method: KSTwoSampleAlternativeMethod,
    nan_policy: NaNPolicy,
) -> Result<(f64, f64), KSTestError>
```

This is more in line with what's in other files in the same folder.

## Summary by CodeRabbit

* **New Features**
  * Added support for creating sorted collections from slices and owned vectors.
  * Added sorted iteration for floating-point vectors with reliable numeric ordering.
  * Added sorted vector support that tracks and preserves sortedness.
  * Added public APIs for accessing, converting, indexing, and iterating over sorted data.
  * Added conversions for common slice and boxed-slice inputs.
  * Added error handling for invalid ordering and sorted-iterator operations.


day01:
@AshrafIbrahim03 did you verify tests with all targets n features?

AshrafIbrahim03:
I did not. I just reread through the contributing section of the README to see if it details how to run those, but I don't see it. How can I do that?

day01:
probably you find out:
`cargo test`

Thanks, the call sites compile now. I checked the latest commit but three KS tests still fail.  currently order is wrong. are you sure it should be like that? may you add some context to Pr ?

AshrafIbrahim03:
Yes, just saw the last three tests fail, pushing a fix for it now!

Some more context: basically we were talking about a contribution I could make in #405 , and it seemed like changing the ks sampling function to take a `&[f64]` instead of a `mut Vec<f64>` would be more in line with the other files, [`anderson_darling`](https://github.com/statrs-dev/statrs/blob/102945824c835692ca54b066e7696341b3312211/src/stats_tests/anderson_darling.rs#L23) and [`chi_square`](https://github.com/statrs-dev/statrs/blob/102945824c835692ca54b066e7696341b3312211/src/stats_tests/chisquare.rs#L68). 

I also noticed the ks one sample function that I should probably bundle in this PR too!

YeungOnion:
I realize I steered you astray. I think it's better to let the caller opt into copying data. Ideally, we could obtain an iterator over the "sorted" data borrowing, but I don't know if that's easy to express, or we just verify order along with nan policy with an iterator argument.

Thoughts as a user?

day01:
If I need to retain the input data, I can clone it explicitly, otherwise the function can consume and sort it without hidden copies

YeungOnion:
@AshrafIbrahim03 that means we should go the other way, convert those that accept slices to instead own as Vec for data we sort in place.

It would be great if there were a way to get an iterator that's element-wise sorted, because then the API expresses we need each value in order. It's incidental that we happen to sort the data ourselves to access that ordering of the values efficiently. Maybe we'll defer that for larger datasets that need to be streamed.

YeungOnion:
@AshrafIbrahim03 need any help here?

AshrafIbrahim03:
> @AshrafIbrahim03 need any help here?

Just coming back to this now.

AshrafIbrahim03:
> It would be great if there were a way to get an iterator that's element-wise sorted, because then the API expresses we need each value in order.

This sounds like it could be a new interface that implements Iterator, but calling next just returns the sorted elements, no?

AshrafIbrahim03:
Accidentally overwrote my prior commits, but it didn't seem like those were needed. I pushed a commit that has some code with a basic sorted iterator. The implementation is not efficient, but is that the type of input you're looking for in the ks_test functions?

day01:
i dont think so, it still will be clone.

YeungOnion:
> The implementation is not efficient, but is that the type of input you're looking for in the ks_test functions?

Think I'd need to see the API for the ks_*sample functions to be certain, but I do agree with @day01 that if you wanted to provide the `Sorted` as is, it would require a clone since it's a data structure type (owns structure and values) since there's not a Rust standard notion of sorted iterator, it also might require a constructor that doesn't sort in case I don't want my data explicitly sorted again (perhaps it was sorted on write to disk, and read protocol will read sorted) and we just validate order on the fly.

But we have some good constraints, we need to be able to have a pull-iterator in a sorted order (the for loop in the algorithm) and we want to avoid a clone of the data in the source structure.

Let me know if you're okay with hints/want something clearer/have argument we're missing for your approach. If you write out a usage example that exhibits both of these:
- calls the `ks_*sample` function that has API expressing, "I have provided an iterator that I assert is sorted"
- while avoiding a struct/type that the user has to handle explicitly (sim to most structs in `core::iter` like `Chain`, `Cloned`, we don't _have_ to think about them)

It could also be a good simplifying point to assume that you start with a Vec<T: Ord> and then generalize from there asking "what would I need to provide to express a similar constraint?"

AshrafIbrahim03:
I think you're right that taking in a new data structure, like `Sorted` shouldn't be the move. I'm wondering if taking in an `impl IntoSortedIterator` would be a better way to go about this, then implementing that type for different types of collections would allow a caller more flexibility with the passed argument. I'm thinking the interface of interacting with sorted data could be similar to parallelizing using rayon's parallelized iterator? Simply calling `IntoSortedIterator::into_sorted_iterator` on an `Iterator<Item=T:Ord>` would return a `SortedIterator`, allowing the ks_*sample to interact with these collections without mutating the vector itself.

I think an Iterator implementation would be the best approach, because only reason for mutability in the functions is for sorting, even calculating the test statistics takes iterators

There's one problem with this approach that I've been looking into during my spare time, that being how to iterate through a collection in a sorted order without mutating the underlying data structure or making the iterator itself expensive. I've been doing some research trying to figure this out with just `Vec<f64>` before generalizing, but if you have any guidance here that would be great!

I'm working on a minimum viable rewrite of the `ks_*sample` functions using `data: impl IntoSortedIterator` just to see if this change would be feasible in the sampling methods, but after evaluating feasibility there I would figure out how to optimize the Sorted to not just clone and sort the input Vector.

Let me know if this is a good approach or not. If this isn't I would like some more guidance on how to approach this problem!

YeungOnion:
I got stuck on this too. The difficult part seems to be that the property we want to express is about the *values yielded by the iterator*, rather than something its type necessarily enforces. A structure like a B-tree can guarantee ordering as an invariant, but an arbitrary iterator generally can't.

There's also a stronger assumption hiding here: the iterator is traversing a fixed dataset. That's more restrictive than just `ExactSizeIterator`, but it still doesn't seem sufficient to statically establish that the yielded values are ordered. Unless the data is already known to be sorted, I think implementing such an API directly would be difficult, so how can we express or mark that the data has already been sorted?

<details>
<summary>Hint if you want it</summary>

What if we narrow the idea to: **"assert that this iterator will uphold an ordering"**?

<details>
<summary>More hint</summary>

Then the implementation needs to know something about adjacent elements -> perhaps `Peekable`.

<details>
<summary>Last hint</summary>

That also gives you a useful building block for something like merging two sorted arrays into a `Left<Float> | Right<Float>`.

</details>
</details>
</details>

AshrafIbrahim03:
I thought about what you said a little bit and realized that you meant making sorting the caller's issue. I made a new struct called `SortedCollection` in sorted_collection.rs that verifies that an input slice is sorted on instantiation and fails if it isn't sorted. The struct itself exposes the `iter` method on its inner slice so that the inner slice has been validated and an iterator over it can be called when needed.

When I was looking at your hints, I thought that they leaned towards Haskell style arrays with compile time bounds on the values. The issue I see with that approach would be that it would mean that the sorted verification could only happen at compile time. If that's not what you meant by it, I would love more clarification so I can try and implement it.

AshrafIbrahim03:
I'm also thinking that the ks_*sample methods could take a borrow to the SortedCollection struct so that any calculations are done with an iterator derived from a verified sorted collection. If this is the implementation we want to go towards, I think that have some helper methods with making a `SortedCollection` would go a long way to help with usability.

YeungOnion:
Yes, I thought to put it on the caller. The compile level I mentioned would only have the value of a marker trait so  would since we can't figure out if data is sorted by its type unless it's from a data structure that has a traversal that guarantees some ordering.

I think SortedCollection is a good start, perhaps named SortedSlice for specificity, and it doesn't seem close to widen the ks two sample API into something like an iterator extension or newtype that SortedSlice provides.

The reason for an iterator API is that I know someone is using statrs for a polars plugin, and polars handles upstream sorts in an execution plan by a marker variable, so there's not actually data yet let alone sorted, but by the time statrs algorithms are executing there will be sorted data and by then I think it will also store the size as well. @FBruzzesi do you have any input on how you'd consume this API for your polars plugin if you were to use it?

FBruzzesi:
Thanks for the ping @YeungOnion

Context for **my** use case first. Today `polars-stats` only uses `statrs::distribution`, and hypothesis tests are out of scope for now. That said, it might be nice to eventually include them.

Disclaimer: I have not read the polars plugin internals, so treat the below as observed behaviour rather than the documented contract. The plugin API is thinly documented, so it is worth double-checking with the polars team (they have a Discord with dedicated plugins and rust channels).

My understanding is that by the time my plugin function runs, polars has already planned the query and executed the sort. What arrives is plain materialized data, and there is no laziness left on the Rust side of the plugin boundary. I have also never found a way to ask, from inside a plugin, whether the column was sorted upstream.

Even if I could check it, polars' notion of sorted is a looser promise than `SortedCollection` makes. A column flagged sorted ascending can have nulls at the front and NaN at the back, and a user can set that flag by hand with no validation at all:

```python
>>> pl.Series([3.0, float("nan"), None, 1.0]).sort().to_list()
[None, 1.0, 3.0, nan]                      # flagged SORTED_ASC

>>> pl.Series([3.0, 1.0, 2.0]).set_sorted().flags
{'SORTED_ASC': True, 'SORTED_DESC': False} # not sorted, no complaint
```

So I would validate on the plugin side regardless, and that is fine (it's one linear scan). The two things that do cost are elsewhere.

1. Lend the data instead of handing it over: The values live in memory polars owns and the plugin borrows. A `Vec<f64>` parameter means copying the whole column before the test even starts. The `&[f64]` in this PR is the biggest win, independent of the sorted question.
2. Do not require the column to be one contiguous block: Polars can hold a column in several chunks, and I only get a flat slice when there is exactly one chunk and no nulls. So a slice-only API makes me glue the chunks back together first, which is the full copy the slice was meant to avoid. This is the real argument for an iterator API: not deferred data, just data in pieces.

Concretely, the difference would look like something along these lines:

```rust
let a = a_series.f64()?.drop_nulls();  // polars nulls are not NaN; NaNPolicy still applies to NaN
let b = b_series.f64()?.drop_nulls();

// slice API: a full copy each whenever the column arrived in more than one chunk
let (a, b) = (a.rechunk(), b.rechunk());
let out = ks_twosample(
    SortedSlice::try_from_slice(a.cont_slice()?)?,
    SortedSlice::try_from_slice(b.cont_slice()?)?,
    method,
    nan_policy,
)?;

// iterator API: no copy, chunks walked in place
let out = ks_twosample(
    a.into_no_null_iter().assert_sorted(),
    b.into_no_null_iter().assert_sorted(),
    method,
    nan_policy,
)?;
```

So to answer your question directly: `SortedSlice` is the one I would reach for first and it is already a clear improvement over the owned `Vec`, but the iterator version is the one that would actually save a copy.

---

I hope this helps 😇

AshrafIbrahim03:
I added a SortedVec implementation where the underlying data is validated as sorted. I'm thinking of implementing a SortedSlice after this so we can have a pair of types that work well together. 

I think having a slice would be best for the current implementation, but I want to make sure I get some feedback on the SortedVec before going for the SortedSlice.

AshrafIbrahim03:
I added a basic `SortedSlice` implementation over the `SortedVec` I made. `SortedSlice` looks like this right now:
```rust
pub struct SortedSlice<'a, T>(&'a SortMarker, &'a [T]);
```
I have a marker in the `SortedVec` and the `SortedSlice` stores a reference to its associated `SortedVec`'s marker. In @FBruzzesi's comment, it looked like a `SortedSlice` should return an Err if it's not already sorted whereas @YeungOnion mentioned marking data as sorted.

I think there's two possible implementations for a `SortedSlice`, one in the way that I've done, and one that fails on instantiation if the underlying data isn't already sorted. I think the error-on-unsorted implementation would be great for users of the API, but the issue is when the underlying data is mutated. A `SortedVec` / `SortedSlice` would have to be read only to guarantee that the data is sorted. I implemented the `SortedVec` / `SortedSlice` with a marker with the idea that the underlying data could be mutated. If the sort marker changes on the `SortedVec`, that gets updated in all the `SortedSlice`s referencing it. I haven't done this yet, but each mutation to a `SortedVec` / `SortedSlice` in this implementation would also require another check to see if the collection is sorted.

I think that the current implementation would work better for a general use case, but the error-on-unsorted implementation would be good for this specific use case this PR is about.

I'm not sure which one is better for the project as a whole, and would like some guidance on this from a maintainer.

YeungOnion:
My view is that computing statistics should eventually support validation
during iteration to support larger than memory datasets and shall later
widen to be a trait bound while admitting implementations that can
circumvent runtime checking of order given some static guarantee that data
has already been dynamically validated.

I think this will likely be a very thin type, so you calling out which is
better suited for this PR (I agree with this) should drive the design.

Overall,
- I'm opposed to my earlier suggestion of a marker indicating sorted state
- I'm for SortedSlice with providing TryInto<SortedSlice> for slices
- I'm opposed to SortedVec, since this PR doesn't have needs related to
ownership. - I'm opposed to providing a coercion to SortedSlice from owned
data, we shall force clients to borrow somewhere.
- I'm for providing the stat API via TryInto<SortedSlice> as I speculate it
provides better API stability, details below.

## What I think is a side note

The reason for a TryInto<SortedSlice> instead of only providing a
parsing/new function is to accommodate a later iterator API that will only
be able to validate via consumption and will keep the same signature.

```rs
// resolves by attempting TryInto<SortedSlice> on datasets and mapping
conversion error before computing statistics
ks_stat_two_sample(&[f64], &[f64], NanPolicy) -> Result<Statistic,
FirstUnordered | SecondUnordered | NanPolicyViolation>

// resolves transparently as no-op
ks_stat_two_sample(SortedSlice<f64>, SortedSlice<f64>, NanPolicy) ->
Result<Statistic, !FirstUnordered | !SecondUnordered | NanPolicyViolation>
// exclamation for unreachable
```

can later be the same code relying on different internal behavior, where
the signature is Into<SortedIterExt> or something similar to iterate atop
SortedIterExt::validate_sorted
```rs
// resolves by iterating values and validating with potentially late failure
ks_stat_two_sample(&[f64], &[f64], NanPolicy) -> Result<Statistic,
FirstUnordered | SecondUnordered | NanPolicyViolation>

// resolves by borrowing iterator with assert style failure (perhaps is a
debug-enabled assertion)
ks_stat_two_sample(SortedSlice<f64>, SortedSlice<f64>, NanPolicy) ->
Result<Statistic, !FirstUnordered | !SecondUnordered | NanPolicyViolation>
```

Thanks for being willing to deliberate with this so much; I have definitely
drawn this out longer than I expected and I apologize that it's
inconvenient.

AshrafIbrahim03:
No worries! I was honestly looking for a reason to contribute to this project. I want to contribute more, but figured something like this would be a good place to start. I'm honestly really interested in this and would like to dive deeper after this PR to try and create a solution for the repo as a whole. I'm down to open up an issue to try and create a crate-wide solution for this! 

For the commits I just pushed, I implemented `TryInto<SortedSlice<T>` for any slices. I think that's what we're getting at right now. Where you could technically just do this:
```rs
let out = ks_twosample(
    a.try_into()?,
    b.try_into()?,
    method,
    nan_policy,
)?;
```
I would need to go through the ks_*sample functions to change the function signature to use this type, but I already wrote up the implementation for the ks_*sample functions using slices so that part should mostly be handled already, I just want to verify that this is the intended usage before bringing the slice implementation in.

YeungOnion:
Thanks, glad to hear it!

I think for now, we implement this within the test_stats module so the data type is not public and later removed or changed (it does mean that users won't be able to write TryInto<SortedCollection>for their own types, but we'll address that later via iterator API). Similarly, I think the code example you have can fail validation later so that we can support extension that will only be able to validate during data iteration, this means that callers can opt to iterate twice to ensure that some error states are unreachable.

Since we'll later design a broader API based on the issue you'll write for expressing (type) and traversing (behavior) sorted data, we'll effectively have tests about the expressiveness of such API to ask,
- what kinds of changes are needed to accommodate the changes from this PR?
- what kinds of changes are needed for a polars extension author to provide data to the iterator API we'd accept?

So for now, I think if we just do the SortedSlice as the only type for now, we can address this PR, all sequential data can be borrowed. I think the work on other types and behavior you have can be distilled into the broader work.

AshrafIbrahim03:
I changed the API of the ks_*sample functions to use `SortedSlice` instead of a mutable owned vector. I also changed the tests to reflect the API change, and all tests pass.

PR #407, opened 2026-07-26
fix: replace hand-rolled quickselect with select_nth_unstable_by

### LINKED ISSUE #163: segfault in `select_inplace` from v0.12
gdb output:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000055dd04d94b20 in statrs::statistics::slice_statistics::select_inplace ()
[Current thread is 1 (Thread 0x7f4ce53fe700 (LWP 2292366))]
(gdb) bt
#0  0x000055dd04d94b20 in statrs::statistics::slice_statistics::select_inplace ()
#1  0x000055dd04d94417 in statrs::statistics::slice_statistics::<impl statrs::statistics::order_statistics::OrderStatistics<f64> for [f64]>::quantile ()
```

I think my input slice to `quantile` had `NaN` values in it. from experimenting, I was able to trigger similar crashes (on v0.12 codebase) if I put a bunch of `NaN` values into the slice.

relevant convo about use of unsafe: https://github.com/statrs-dev/statrs/pull/109

on a local codebase, I swapped in the safer replacement code that was made in the pull request linked above, and was able to get this panic output:

```
---- statistics::slice_statistics::test::check_quantile_on_256k_slice stdout ----
thread 'statistics::slice_statistics::test::check_quantile_on_256k_slice' panicked at 'index out of bounds: the len is 262144 but the index is 262144', src/statistics/slice_statistics.rs:340:20
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

which corresponds to 

```
        // inside select_inplace
        loop {
            loop {
                begin += 1;
                if arr[begin] >= pivot {  // <- line 340 is this one
                    break;
                }
            }
            loop {
                end -= 1;
                if arr[end] <= pivot {
                    break;
                }
            }
            if end < begin {
                break;
            }
            arr.swap(begin, end);
        }
```

I am using v0.12 (in a bunch of places actually) because I have code that uses the old api. would you be open to a pull request to swap in the safer version of `select_inplace` and issue another v0.12 release? 

(here is the test that triggered the above btw)

```
    #[test]
    fn check_quantile_on_256k_slice() {
        for _ in 0..(1024) {
            let mut xs: Vec<f64> = rand::distributions::Standard.sample_iter(&mut thread_rng()).take(1024*256).collect();
            for i in (0..xs.len()).step_by(42) {
                xs[i] = f64::NAN;
            }
            xs.quantile(0.01) ;
            xs.quantile(0.05) ;
            xs.quantile(0.1)  ;
            xs.quantile(0.2)  ;
            xs.quantile(0.25) ;
            xs.quantile(0.3)  ;
            xs.quantile(0.4)  ;
            xs.quantile(0.45) ;
            xs.quantile(0.5)  ;
            xs.quantile(0.55) ;
            xs.quantile(0.6)  ;
            xs.quantile(0.65) ;
            xs.quantile(0.7)  ;
            xs.quantile(0.75) ;
            xs.quantile(0.8)  ;
            xs.quantile(0.9)  ;
            xs.quantile(0.95) ;
            xs.quantile(0.99) ;
        }
    }

```

agene0001:
`select_inplace` implements the Numerical Recipes quickselect by hand. With NaN in the slice, its partition loop walks past the end of the array — the crash reported in #163 (a segfault under the pre-0.13 `unsafe` version, an index-out-of-bounds panic today). Reproducer that panics on `main`:

```rust
let mut v: Vec<f64> = (0..1000).map(|i| (i as f64 * 0.7).sin()).collect();
for i in (0..1000).step_by(7) { v[i] = f64::NAN; }
Data::new(v).quantile(0.5); // panicked at src/statistics/slice_statistics.rs:106:28
```

Replacing it with `<[T]>::select_nth_unstable_by(rank, f64::total_cmp)` fixes it two ways:

- **NaN gets a defined position** in a total order, so the search stays in bounds (previous behaviour with NaN was arbitrary).
- **Worst case is O(n)** (introselect) instead of the NR quickselect's quadratic worst case.

## Performance

The first version of this PR claimed ~4.4x. That was measured on large random input only, and it was wrong as a general statement — thanks @day01 for asking for a reproducible benchmark, which is what surfaced it.

Shape matters as much as size. The NR quickselect took its pivot from a median of three, which on sorted or reverse-sorted input lands on the true median and finishes in a single partition, so it *beat* plain introselect on exactly those shapes. This PR therefore checks for order first, which recovers that and improves on it — the answer becomes an index rather than a partition.

`benches/order_statistics.rs` gains a `selection scaling` group sweeping size against shape, written against only the public `Data` API so it compiles on any branch and `--save-baseline` works across a checkout. Against `main`, medians of the Criterion intervals:

| shape | 1024 | 65536 | 1048576 |
|---|---|---|---|
| random | +12% | −36% | −76% |
| sorted | −3% | −0% | +15% |
| reversed | −37% | −29% | −26% |
| organ_pipe | −46% | −64% | −71% |
| duplicates | −19% | −64% | −79% |

Twelve of fifteen improve. Sorted input is roughly a wash below a million elements and 15% slower at a million, where the `total_cmp` scan costs more per element than the median-of-three pivot it replaces. Small random input is 12% slower, where introselect's setup costs more than the old quickselect's.

Keeping the fast paths still beats dropping them: without them, sorted is +75% to +89% and reversed +13% to +30% against main.

## Tests

The regression test covers NaN scattered through a large slice, all-NaN input, and NaN at the exact positions the old partition tripped on.

Added on review: every order statistic, the median, and the quantile endpoints checked against a sorted reference over eight shapes and nine sizes, plus permutation invariance. These compare **bit patterns**, not values, so a wrong sign of zero is a failure — which is how the `signed_zeros` shape pins @day01's catch that `<=` considers `[-1.0, +0.0, -0.0, 1.0]` sorted and returns the wrong zero.

Existing order-statistic/median/quantile tests (including `test_median_robust_on_infinities`) pass unchanged.

Closes #163

### DIFF
--- a/benches/order_statistics.rs
+++ b/benches/order_statistics.rs
@@ -94,5 +94,66 @@ fn bench_order_statistic(c: &mut Criterion) {
     group.finish();
 }
 
-criterion_group!(benches, bench_order_statistic);
+/// Selection cost across input sizes and shapes.
+///
+/// Kept separate from the fixed-size group above so the two can evolve
+/// independently, and written against only the public `Data` API so the same
+/// file compiles on any branch. To compare two implementations:
+///
+/// ```text
+/// git checkout main    && cargo bench --bench order_statistics -- --save-baseline main
+/// git checkout <branch> && cargo bench --bench order_statistics -- --baseline main
+/// ```
+///
+/// The shapes matter as much as the sizes. A median-of-three pivot is fine on
+/// random data and degrades on input built to defeat it, so a random-only
+/// benchmark would understate the difference between a quickselect and an
+/// introselect with an O(n) fallback.
+fn bench_selection_scaling(c: &mut Criterion) {
+    fn shaped(shape: &str, n: usize) -> Vec<f64> {
+        let mut rng = StdRng::seed_from_u64(0xB0A7);
+        match shape {
+            // shuffled, the ordinary case
+            "random" => {
+                let mut v: Vec<f64> = (0..n).map(|i| i as f64).collect();
+                v.shuffle(&mut rng);
+                v
+            }
+            // already ordered, both directions
+            "sorted" => (0..n).map(|i| i as f64).collect(),
+            "reversed" => (0..n).rev().map(|i| i as f64).collect(),
+            // ascends then descends: the classic median-of-three adversary,
+            // since the first, middle and last elements are unrepresentative
+            "organ_pipe" => (0..n)
+                .map(|i| if i < n / 2 { i } else { n - i } as f64)
+                .collect(),
+            // few distinct values, so partitions are heavily unbalanced
+            "duplicates" => {
+                let distinct = ((n as f64).sqrt() as usize).max(1);
+                let mut v: Vec<f64> = (0..n).map(|i| (i % distinct) as f64).collect();
+                v.shuffle(&mut rng);
+                v
+            }
+            other => panic!("unknown shape {other}"),
+        }
+    }
+
+    let mut group = c.benchmark_group("selection scaling");
+    for &n in &[1_024usize, 65_536, 1_048_576] {
+        for shape in ["random", "sorted", "reversed", "organ_pipe", "duplicates"] {
+            let data = shaped(shape, n);
+            group.throughput(Throughput::Elements(n as u64));
+            group.bench_function(format!("median/{shape}/{n}"), |b| {
+                b.iter_batched(
+                    || Data::new(data.clone()),
+                    |data| black_box(data.median()),
+                    BatchSize::LargeInput,
+                )
+            });
+        }
+    }
+    group.finish();
+}
+
+criterion_group!(benches, bench_order_statistic, bench_selection_scaling);
 criterion_main!(benches);
--- a/benches/order_statistics.rs
+++ b/benches/order_statistics.rs
@@ -96,38 +96,25 @@ fn bench_order_statistic(c: &mut Criterion) {
 
 /// Selection cost across input sizes and shapes.
 ///
-/// Kept separate from the fixed-size group above so the two can evolve
-/// independently, and written against only the public `Data` API so the same
-/// file compiles on any branch. To compare two implementations:
-///
-/// ```text
-/// git checkout main    && cargo bench --bench order_statistics -- --save-baseline main
-/// git checkout <branch> && cargo bench --bench order_statistics -- --baseline main
-/// ```
-///
-/// The shapes matter as much as the sizes. A median-of-three pivot is fine on
+/// Shape matters as much as size here: a median-of-three pivot is fine on
 /// random data and degrades on input built to defeat it, so a random-only
-/// benchmark would understate the difference between a quickselect and an
-/// introselect with an O(n) fallback.
+/// benchmark would not tell the two implementations apart.
 fn bench_selection_scaling(c: &mut Criterion) {
     fn shaped(shape: &str, n: usize) -> Vec<f64> {
         let mut rng = StdRng::seed_from_u64(0xB0A7);
         match shape {
-            // shuffled, the ordinary case
             "random" => {
                 let mut v: Vec<f64> = (0..n).map(|i| i as f64).collect();
                 v.shuffle(&mut rng);
                 v
             }
-            // already ordered, both directions
             "sorted" => (0..n).map(|i| i as f64).collect(),
             "reversed" => (0..n).rev().map(|i| i as f64).collect(),
-            // ascends then descends: the classic median-of-three adversary,
-            // since the first, middle and last elements are unrepresentative
+            // ascends then descends: defeats median-of-three
             "organ_pipe" => (0..n)
                 .map(|i| if i < n / 2 { i } else { n - i } as f64)
                 .collect(),
-            // few distinct values, so partitions are heavily unbalanced
+            // few distinct values: unbalanced partitions
             "duplicates" => {
                 let distinct = ((n as f64).sqrt() as usize).max(1);
                 let mut v: Vec<f64> = (0..n).map(|i| (i % distinct) as f64).collect();
--- a/src/statistics/slice_statistics.rs
+++ b/src/statistics/slice_statistics.rs
@@ -64,19 +64,15 @@ impl<D: AsMut<[f64]> + AsRef<[f64]>> Data<D> {
         self.0.as_ref().iter()
     }
 
-    // Selection via `<[T]>::select_nth_unstable_by`, an introselect with an
-    // O(n) worst case. `total_cmp` also gives NaN a defined position (positive
-    // NaNs sort past +inf) rather than the arbitrary result the previous
-    // Numerical Recipes partition produced, which is the point of the change.
+    // Introselect via `select_nth_unstable_by`, ordered by `total_cmp` so NaN
+    // has a defined position instead of the arbitrary one the old Numerical
+    // Recipes partition produced.
     //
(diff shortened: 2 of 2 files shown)


day01:
Could you extend the existing benches/order_statistics.rs benchmark and include a reproducible Criterion between main and this branch? The PR reports a 4.4x improvement, but currently there is no benchmark setup or i cannot find it.

day01:
additionally, based ontest, is focused on edge case, can you test positive scenario with values?

benches/order_statistics.rs
```diff
@@ -94,5 +94,66 @@ fn bench_order_statistic(c: &mut Criterion) {
...
 
-criterion_group!(benches, bench_order_statistic);
+/// Selection cost across input sizes and shapes.
+///
+/// Kept separate from the fixed-size group above so the two can evolve
+/// independently, and written against only the public `Data` API so the same
+/// file compiles on any branch. To compare two implementations:
+///
+/// ```text
+/// git checkout main    && cargo bench --bench order_statistics -- --save-baseline main
```
day01: it shouldnt be here, i asked only for results of benches. like in one other of your prs I prepared.

src/statistics/slice_statistics.rs
```diff
@@ -74,61 +85,19 @@ impl<D: AsMut<[f64]> + AsRef<[f64]>> Data<D> {
...
-            }
-            if end <= rank {
-                low = begin;
-            }
+        // `<=` rather than `total_cmp` here, which is both cheaper per element
+        // and still correct: any NaN makes some comparison false, so a slice
+        // containing one never takes a fast path and falls through to the
+        // general case, and without NaN the two orderings agree.
+        let slice = self.0.as_mut();
+        if slice.is_sorted_by(|a, b| a <= b) {
```
day01: `<=` does not define the same ordering as total_cmp for signed zero. For `[-1.0, +0.0, -0.0, 1.0]`, `order_statistic(2)` returns `+0.0`, while the `total_cmp` reference is `-0.0`. Please use total_cmp consistently for both fast paths or remove them.

src/statistics/slice_statistics.rs
```diff
@@ -74,61 +85,19 @@ impl<D: AsMut<[f64]> + AsRef<[f64]>> Data<D> {
...
+            return slice[rank];
+        }
+        if slice.is_sorted_by(|a, b| a >= b) {
```
day01: same as up

src/statistics/slice_statistics.rs
```diff
@@ -74,61 +85,19 @@ impl<D: AsMut<[f64]> + AsRef<[f64]>> Data<D> {
...
-            self[low + 1] = self[end];
-            self[end] = pivot;
-
-            if end >= rank {
-                high = end - 1;
-            }
-            if end <= rank {
-                low = begin;
-            }
+        // `<=` rather than `total_cmp` here, which is both cheaper per element
```
day01: claude things :)

src/statistics/slice_statistics.rs
```diff
@@ -550,6 +519,141 @@ mod tests {
 
     // TODO: test codeplex issue 5667 (Math.NET)
 
+    /// Selection agrees with sorting, on ordinary values across a range of
+    /// shapes and sizes.
+    ///
+    /// The counterpart to the NaN and infinity tests below, which pin that
```
day01: may we make it more human desc?

src/statistics/slice_statistics.rs
```diff
@@ -64,8 +64,19 @@ impl<D: AsMut<[f64]> + AsRef<[f64]>> Data<D> {
...
     }
 
-    // Selection algorithm from Numerical Recipes
-    // See: https://en.wikipedia.org/wiki/Selection_algorithm
+    // Selection via `<[T]>::select_nth_unstable_by`, an introselect with an
+    // O(n) worst case. `total_cmp` also gives NaN a defined position (positive
+    // NaNs sort past +inf) rather than the arbitrary result the previous
+    // Numerical Recipes partition produced, which is the point of the change.
+    //
+    // The ordered fast paths are not incidental. The NR quickselect took its
```
day01: this desc is needed r u sure?

src/statistics/slice_statistics.rs
```diff
@@ -550,6 +519,141 @@ mod tests {
...
+    /// covers the ordered fast paths in `select_inplace`: those return an
+    /// index instead of partitioning, so they have to produce exactly what the
+    /// general path would, including for the descending case where the index
+    /// is counted from the far end.
+    #[test]
+    #[cfg(feature = "std")]
+    fn test_order_statistics_match_sorted_reference() {
+        fn shaped(shape: &str, n: usize) -> Vec<f64> {
+            match shape {
+                // deterministic pseudo-random, no rand dependency needed
```
day01: we have rand, i dont know why we have to forget about rand in tests, deps are welcome especially in tests

benches/order_statistics.rs
```diff
@@ -94,5 +94,66 @@ fn bench_order_statistic(c: &mut Criterion) {
...
+        }
+    }
+
+    let mut group = c.benchmark_group("selection scaling");
+    for &n in &[1_024usize, 65_536, 1_048_576] {
+        for shape in ["random", "sorted", "reversed", "organ_pipe", "duplicates"] {
+            let data = shaped(shape, n);
+            group.throughput(Throughput::Elements(n as u64));
+            group.bench_function(format!("median/{shape}/{n}"), |b| {
+                b.iter_batched(
```
day01, line 134: iter_batched excludes data.clone() from the measurement. Are the numbers in the PR description from a different benchmark?

agene0001:
Good catch on the signed zero — that's a real bug and you gave the exact counterexample. `<=` reports `[-1.0, +0.0, -0.0, 1.0]` as sorted because `+0.0 <= -0.0` holds, so the fast path indexed straight in and returned `+0.0` where the `total_cmp` reference gives `-0.0`. My reasoning for `<=` covered NaN and missed signed zero completely.

Both fast paths now use `total_cmp`. The value tests gained a `signed_zeros` shape and now compare bit patterns rather than values, since `-0.0 == +0.0` would have hidden it — restoring `<=` fails with `signed_zeros/3: order_statistic(2) = 0, want -0`.

On `iter_batched`: yes, the clone is excluded, and that's intended — the measurement should be selection, not allocation. The numbers came from this benchmark via `--baseline main`, but the ones in the description were from the `<=` version, so they were stale. `total_cmp` costs more per element; current figures (medians of the Criterion intervals), and I've updated the description to match:

| shape | 1024 | 65536 | 1048576 |
|---|---|---|---|
| random | +12% | −36% | −76% |
| sorted | −3% | −0% | +15% |
| reversed | −37% | −29% | −26% |
| organ_pipe | −46% | −64% | −71% |
| duplicates | −19% | −64% | −79% |

Twelve of fifteen improve. Sorted input is roughly a wash below a million elements and 15% slower at a million, where the `total_cmp` scan costs more than the median-of-three pivot it replaces. Keeping the fast paths still beats dropping them, which would put sorted back at +75% to +89% and reversed at +13% to +30% — but if you'd rather have the simpler code, say so and I'll drop them.

Also removed the run instructions from the bench file and trimmed the commentary, and switched the tests to `rand`.

YeungOnion:
Thanks for the review @day01 especially with things like docs by Claude - easy to gloss over that stuff. Is this good to go?

day01:
it is, lets play with it.

PR #408, opened 2026-07-26
fix: accumulate central moments relative to the first observation

### LINKED ISSUE #376: Numerical Stability of Variance

agene0001:
Follow-up to #394 / cf11836, which moved `Statistics` onto Welford for this issue. That cured the catastrophic case #376 opened with, but Welford is still poorly conditioned when the data carries a large offset: `mean += delta / n` cannot represent a small increment against a large running mean, so the low bits of every update are dropped.

On the dataset from the issue — `1e12 + U(0,1)`, n = 1e6 — measured against a Neumaier-compensated two-pass reference:

| | relative error in `variance` |
|---|---|
| current (Welford) | 2.5e-4 |
| this PR | **5e-15** |

Central moments are invariant under a shift, so `OnlineMoments` now accumulates moments of `x − first_observation`, keeping every magnitude small. Cost is one subtraction per observation: 1107 µs vs 1103 µs over 1e6 elements, i.e. free. (For comparison, I also prototyped a double-double running mean: 2x slower *and* less accurate, 1.9e-8.)

Also adds `OnlineMoments::merge` (Chan–Golub–LeVeque pairwise update, Pébay's form for M3), which is the parallelisability this issue asks for. It reconciles the two accumulators' different offsets, so it is included here rather than separately. Folding into several accumulators and merging is also slightly better conditioned than one long chain.

Tests: exact-data offset sweep at offsets 2^0 … 2^40 (powers of two + a 2^-10 step so every sample is exactly representable and the reference variance is exact), merge-vs-single-chain equality at every split point, and merge with differing offsets.

Closes #376

## Summary by CodeRabbit

* **New Features**
  * Added the ability to merge online statistical accumulators, including datasets with different starting values.
  * Merging correctly handles moments through the third order and treats empty accumulators as neutral.

* **Bug Fixes**
  * Improved numerical precision when calculating statistics from large-magnitude values.
  * Mean calculations now remain accurate after processing offset data.

* **Documentation**
  * Added coverage for precision, merging, and empty-accumulator behavior.

### DIFF
--- a/src/statistics/online.rs
+++ b/src/statistics/online.rs
@@ -15,6 +15,7 @@ use num_traits::Float as _;
 /// - `2`: `1` + variance
 /// - `3`: `2` + skewness
 /// - values above: not presently implemented
+///
 /// Moments are accumulated for `x - offset`, where `offset` is the first value
 /// pushed. Central moments are invariant under that shift, and it is what makes
 /// the accumulator usable on data with a large offset.
--- a/src/statistics/online.rs
+++ b/src/statistics/online.rs
@@ -11,21 +11,17 @@ use num_traits::Float as _;
 /// Single-pass accumulator for central moments via Welford's online algorithm.
 ///
 /// `ORDER` controls which moments are tracked:
-/// - `1` mean
-/// - `2` + variance
-/// - `3` + skewness
-/// - above does not presently implement further moments
-///
+/// - `1`: count + mean
+/// - `2`: `1` + variance
+/// - `3`: `2` + skewness
+/// - values above: not presently implemented
 /// Moments are accumulated for `x - offset`, where `offset` is the first value
 /// pushed. Central moments are invariant under that shift, and it is what makes
-/// the accumulator usable on data with a large offset: Welford's `mean +=
-/// delta / n` update cannot represent a small increment against a large running
-/// mean, so `1e12 + U(0, 1)` came out with `2.5e-4` relative error in the
-/// variance. Referring everything to the first observation keeps the magnitudes
-/// small and brings that to `5e-15` at no cost (statrs-dev/statrs#376).
+/// the accumulator usable on data with a large offset.
 pub struct OnlineMoments<const ORDER: usize> {
     pub count: u64,
-    /// The first value pushed; `m` holds the moments of `x - offset`.
+    // `m` holds the moments of `x - offset`. Welford's update can become
+    // insensitive reducing unscaled data. See statrs-dev/statrs#376.
     offset: f64,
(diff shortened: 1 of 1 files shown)


YeungOnion:
Hey, @agene0001  think this is good to go with the docs I added?

Issue #288, opened 2024-09-12
tests to verify correctness of sampling


YeungOnion:
We could implement a few test statistics for goodness of fit for these distributions, would at least help me for verifying #287 
- $\chi^2$
- Komolgorov-Smirnov
- Anderson-Darling
- Cramér-von Mises

and then we'd also have some features

DanielLacina:
I'll look into it.

PR #409, opened 2026-07-26
perf: sample Binomial via BINV/BTPE instead of n Bernoulli trials

### LINKED ISSUE #183: Multinomial sampling is very slow
Looking into a program I wrote I found the limiting factor was the multinomial sampling:

let mut rng = SmallRng::from_entropy();
let result = Multinomial::new(&weights, 100000).unwrap();
let counts = result.sample(&mut rng);

Specifically the sampling portion of the above code, the SmallRng call is small in comparison.

I rewrote this particular portion of my code in python using numpy.random.multinomial and found an ~400x speed increase. It appears the C code numpy calls on uses an implementation that chains many binomial calls together, whereas the statsrs implementation uses a cdf.

Wonder if there's any plans to change this?

agene0001:
`Binomial::sample` ran one Bernoulli trial per trial, so a single draw was O(n). The case reported in #183 (5 categories, n = 100 000, sampled via `Multinomial`) took 22.4 ms per 100 draws here; numpy was reported ~400x faster.

This replaces it with Kachitvichyanukul & Schmeiser (1988): BINV inversion for `n·min(p,1−p) < 10`, BTPE rejection otherwise, exact shortcuts for p ∈ {0, 1}, and a Poisson-limit path when `1 − p` rounds to 1.0.

| | before | after |
|---|---|---|
| `Binomial(0.4, 1_000).sample()` | 3.6 µs | 56 ns |
| `Binomial(0.4, 100_000).sample()` | 361 µs | 39 ns |
| `Binomial(0.4, 1e9).sample()` | ~3.6 s | 39 ns |
| `Multinomial(5 cats, n=1e5)` ×100 | 22.4 ms | 22.8 µs (**981x**) |

`Multinomial` benefits automatically since it draws conditioned binomials per category.

The implementation is adapted from `rand_distr` (MIT/Apache-2.0, attributed in-source). One detail worth review attention: the final acceptance test uses the Stirling-series signs from GSL, which **differ from the published paper** and were verified by one of the algorithm's original designers (see the in-source comment).

Correctness: chi-square goodness-of-fit against the exact pmf over five parameter sets covering every code path (BINV, BTPE, both flipped variants), with expected-count pooling and a 6σ threshold on fixed seeds — deterministic, so a failure means a real defect, not chance. Plus moment tests for n = 1e9 and the Poisson-limit path, and degenerate-parameter tests. Relates to the sampling-verification wish list in #288.

Closes #183

### DIFF
--- a/src/distribution/binomial.rs
+++ b/src/distribution/binomial.rs
@@ -122,6 +122,13 @@ impl ::rand::distr::Distribution<u64> for Binomial {
     /// Kachitvichyanukul, V. and Schmeiser, B. W. (1988). Binomial random
     /// variate generation. Communications of the ACM 31(2), 216-222.
     /// <https://doi.org/10.1145/42372.42381>
+    ///
+    /// # Remarks
+    ///
+    /// This picks the algorithm for you, which is what you want unless you have
+    /// a specific reason otherwise. To choose explicitly, build a
+    /// [`BinomialSampler`] with [`Binomial::sampler`]; both algorithms are
+    /// selectable, subject to BTPE being defined for the parameters.
     fn sample<R: ::rand::Rng + ?Sized>(&self, rng: &mut R) -> u64 {
         sample_unchecked(rng, self.n, self.p)
     }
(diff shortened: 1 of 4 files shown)


src/distribution/binomial.rs
```diff
@@ -115,11 +115,218 @@ impl core::fmt::Display for Binomial {
...
+
+    // the distribution is symmetric under p -> 1 - p
+    let flipped = p > 0.5;
+    let p = if flipped { 1.0 - p } else { p };
+    let q = 1.0 - p;
+
+    if q == 1.0 {
+        // p is below ~2^-54 (and not flipped, since p <= 0.5): the
+        // distribution is indistinguishable from Poisson(n * p) to O(p)
+        return super::poisson::sample_unchecked(rng, n as f64 * p) as u64;
```
YeungOnion: needs to flip

src/distribution/binomial/sampling.rs
```diff
@@ -567,4 +774,131 @@ mod tests {
...
+            let mut rng = StdRng::seed_from_u64(0x5EED + n);
+
+            let mean = n as f64 * p;
+            let sd = (mean * (1.0 - p)).sqrt();
+            let lo = (mean - 6.0 * sd).floor().max(0.0) as u64;
+            let hi = ((mean + 6.0 * sd).ceil() as u64).min(n);
+
+            let mut counts = vec![0u64; (hi - lo + 1) as usize];
+            let mut outside = 0u64;
+            for _ in 0..SAMPLES {
```
YeungOnion, line 631: would you not be able to build atop stats_tests::chisquare for this?

YeungOnion:
made some comments on the implementation.

On a broader note, I'm not sure how I feel about AI driven development for this project. Do we need to develop faster? I'm not going to be reviewing much faster, since the speed isn't my goal. If you want to keep one or two of your PRs not as a draft, that would be appreciated.

Even broader, there's mentions of GSL here - while my understanding is that it's complex and developing around licensing from the training corpus, I don't really know how to regard if the model is referring to it. Can you post the chat here?

agene0001:
@YeungOnion Sorry about that. I was using it for a project of mine and though I might spend some time updating it. I didn't mean to dump so many pr's at once. The moved most of the pr's to draft so its not as much as a burden and for the gsl question, I was able to get a response from the session to hopefully clear up any misconceptions. 
The provenance has three steps, and none of them is GSL:

1. **The algorithm** is from a 1988 paper in *Communications of the ACM* by Kachitvichyanukul & Schmeiser. Algorithms as such aren't copyrightable — specific *code* is. Implementing a published paper is normal and fine.
2. **The Rust code** was adapted from `rand_distr`, licensed `MIT OR Apache-2.0` — permissive, compatible with statrs's MIT. It's attributed in-source at `binomial.rs:130-134`, including the copyright lines, which is what MIT actually requires.
3. **GSL appears only inside `rand_distr`'s own comments.** Its authors noted that GSL's version uses different signs on some Stirling terms than the published paper does, and that GSL's authors reported those corrected signs were verified by one of the original algorithm designers.

So the GSL mentions in my code are paraphrases of comments in MIT-licensed code, describing a *mathematical fact* — which signs are correct. Anyone can download the `rand_distr` crate and read lines 364-368 of `src/binomial.rs` for themselves.

day01:
@YeungOnion that kind of function shouldnt be like different backend/trait or sth? i prefer share decission with users about decisisoin about sampling etc

src/distribution/binomial/sampling.rs
```diff
@@ -567,4 +783,199 @@ mod tests {
...
+        assert!((total as f64 - 2000.0).abs() < 268.0, "Poisson-limit path total {total}");
+    }
+
+    /// The Poisson-limit branch reflects its result like every other path, but
+    /// that reflection is currently unreachable: entering the branch needs
+    /// `p <= 2^-54`, while reflecting can never produce a `p` below `2^-53`.
+    /// The margin is a single bit, so pin it -- a later change to either the
+    /// guard or the reflection that makes the branch reachable should fail here
+    /// rather than silently return an unreflected variate.
+    #[test]
```
YeungOnion, line 741: I think this one tests float implementation, perhaps we can expose the logic within a private function? See other comment.

src/distribution/binomial.rs
```diff
@@ -115,11 +115,227 @@ impl core::fmt::Display for Binomial {
...
+    if p >= 1.0 {
+        return n;
+    }
+
+    // the distribution is symmetric under p -> 1 - p
+    let flipped = p > 0.5;
+    let p = if flipped { 1.0 - p } else { p };
+    let q = 1.0 - p;
+
+    if q == 1.0 {
```
YeungOnion: maybe something like this would enable testing in that short circuit -> poisson distribution
`if let Some(s) = short_circuit_function(p) { return p };`

YeungOnion:
Overall, good PR, and would like to bring it in, moving out of draft

YeungOnion:
@agene0001 appreciate it! For the licensing, we'll assume it's okay, we've at least some record of investigation. 


@day01 We could allow the user selection of algorithm, but these are much more developed algorithms than what we have currently, which is an unoptimized coin flip counter. But exposing the option to select algorithm regardless of `n*p` could be good for some. I think it would be good to add atop this PR.

As far as defaults go, it's good if the need is expressed by "I want to sample the Binomial distribution" instead of something more granular, and the docs mention it as well.

agene0001:
Added the algorithm selection, atop this PR as suggested.

`Binomial::sampler(algorithm)` returns a `BinomialSampler` bound to a `BinomialAlgorithm` — `Automatic`, `Inversion` (BINV) or `Rejection` (BTPE). It implements `rand`'s `Distribution`, so it composes with `sample_iter` and the rest:

```rust
let dist = Binomial::new(0.3, 1_000).unwrap();
let sampler = dist.sampler(BinomialAlgorithm::Rejection).unwrap();
let draws: Vec<u64> = sampler.sample_iter(&mut rng).take(5).collect();
```

Sampling a `Binomial` directly is untouched and still chooses for you, per your point about the default being expressed as "I want to sample the Binomial distribution". Its docs now point at the explicit route.

## One caveat on "regardless of `n*p`"

BTPE can't be forced for an arbitrary mean, and I think this is worth knowing independently of the API question.

Its triangle region has radius `floor(2.195*sqrt(n*p*q) - 4.6*q) + 0.5`, which goes **negative** once the mean is small, inverting the region boundaries. The sampler still terminates and still returns values within `[0, n]`, so nothing looks wrong from the outside — but the distribution is not binomial. Measured against the exact pmf over 200k draws:

| n | p | n*p | chi-square | df |
|---|---|---|---|---|
| 100 | 0.04 | 4.0 | **11,042,868** | 14 |
| 100 | 0.044 | 4.4 | 11.6 | 15 |
| 100 | 0.05 | 5.0 | 18.3 | 16 |
| 100 | 0.10 | 10.0 | 26.0 | 24 |

The cliff is exactly where `2.195*sqrt(n*p*q) - 4.6*q` crosses zero, at `n*p ≈ 4.39*q`.

Since the failure is silent, `sampler` validates that expression and returns `BinomialAlgorithmError::RejectionMeanTooSmall` rather than handing back a sampler that quietly lies. BINV has no such restriction and can be forced anywhere, including where `Automatic` would pick BTPE. Degenerate parameters and the Poisson limit run neither algorithm, so they ignore the request instead of failing.

## Tests

The chi-square goodness-of-fit now runs for every valid (parameters, algorithm) pair, including inversion forced at `n = 500, p = 0.85` where the default would pick BTPE. The validity predicate is checked against BTPE's own radius expression across 6 values of `n` by 99 of `p`, along with the invariant that `Automatic` never selects BTPE where it's undefined. Making the predicate unconditionally `true` fails both tests.

`sample_unchecked` and `BinomialSampler` share a `sample_plan` helper, so both apply the `p -> 1 - p` reflection through the same code.

src/distribution/binomial/sampling.rs
```diff
@@ -115,14 +115,437 @@ impl core::fmt::Display for Binomial {
 #[cfg(feature = "rand")]
 #[cfg_attr(docsrs, doc(cfg(feature = "rand")))]
 impl ::rand::distr::Distribution<u64> for Binomial {
+    /// Samples in `O(1)` expected time (independent of `n`) via the BINV
```
day01, line 19: lot of code in one file. maybe some module? or sth? currently 1k is hard to read for human, and llm too tbh.

https://arxiv.org/pdf/2601.03731#:~:text=Our%20results%20unveil%20a%20critical%20Aggregation%20Deficit,loss%20of%20consistency%20beyond%20100%20execution%20steps.
https://cbarkinozer.medium.com/agentic-code-generation-papers-part-2-23d6482da032

pls split it to a few smaller

YeungOnion: Seems a good call, moving Binomial algorithms and their tests into module would be a good grouping

YeungOnion:
I like it! Thanks @day01 and @agene0001

PR #411, opened 2026-07-26
fix: correct Geometric::inverse_cdf against its definition

### LINKED ISSUE #342: issue in default inverse_cdf, found with simple case of Geometric

agene0001:
> **Stacked on #410** — contains its commit; review only the last commit until that merges.

`Geometric` overrides `inverse_cdf` with the closed form `ceil(ln1p(-x) / ln1p(-p))`. The quotient is only *approximately* integral at the step boundaries, so `ceil` lands on either side of the true one: over a sweep of 200 values of `p` × 400 of `k`, **7255 of 16200** pairs had `inverse_cdf(cdf(k)) != k`.

Being an override, this was unreachable by the improvements to the default `inverse_cdf` (#390, #391) — which is why #342 stayed open after the explicit implementation it asks for had landed.

The closed form is now corrected against the definition being inverted — the least `k` with `cdf(k) >= x`:

- ordinary rounding puts it within one step, so the fast path is two or three `cdf` evaluations (8 → 23 ns/call);
- once `x` is within a few ulp of 1, `1 − x` has lost every significant bit and the closed form can be off **without bound** (`cdf` has a plateau ~1/p wide there; at p = 1e-9 the correct answer is 36331335444), so it falls back to a bisection that is exact by definition.

After: 0 genuine round-trip failures (the 402 remaining are cases where `cdf(k-1) == cdf(k)` in f64, so no integer is recoverable), and 0 violations of the definition over 30266 inputs including `x` within ulps of 0 and 1.

This also **un-ignores** `test_inverse_cdf_small_p` and `test_inverse_cdf_extreme_tail` (`#[ignore = "Gaps in pathological corner cases and testing"]`). The former was masking two independent bugs: it constructs `p = 1 − 1e-14`, which the old `ulps_eq!` epsilon (#410) snapped to the degenerate `p == 1` branch, returning 1 instead of `u64::MAX`. It needs both fixes — hence the stacking.

Closes #342

### DIFF
--- a/src/distribution/binomial.rs
+++ b/src/distribution/binomial.rs
@@ -237,7 +237,11 @@ impl Distribution<f64> for Binomial {
         } else {
             (0..self.n + 1).fold(0.0, |acc, x| {
                 let p = self.pmf(x);
-                acc - p * p.ln()
+                // A mass that underflows to zero contributes nothing, taking
+                // `0 * ln 0 == 0` as usual for entropy. Evaluating it would give
+                // `0 * -inf == NaN` and poison the whole sum, which made
+                // `entropy()` return `NaN` for any `n` past roughly 1100.
+                if p > 0.0 { acc - p * p.ln() } else { acc }
             })
         };
         Some(entr)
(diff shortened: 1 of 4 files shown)


YeungOnion:
Closing — this is the same Geometric::inverse_cdf fix already merged as #431.

PR #412, opened 2026-07-26
fix: restore full precision to erf's interval constants

### LINKED ISSUE #440: Special functions and `libm` compatibility
statrs should route all float ops through `num_traits` which means we'll rely on libm or std based.
For a brief period, I thought, we should consider using `libm` since that's a growing default for the no_std environment, but then how complete does a math function backend need to be for consistency?
libm implements special functions that aren't available in std and there are other crates that maintain special functions, this exposes cases where statrs and other functions aren't the same.

The question I have on this is,
> What footguns can we mitigate for users who need some statistics functions that wouldn't already be mitigating if they had a crash course in floating point?

I think the community convergence onto a special functions crate could reduce some of this, since the number of places needed to make consistent choices could be reduced. Right now `num_traits` has this role for the overlap between `std` and `libm`. Perhaps a crate would be able to implement sane trait-level defaults while exposing specific algorithms as functions where the backend selection would only control the trait-level would be useful. Then we can contribute upstream to broader numerics needs as well.

[`special`](https://docs.rs/special/0.14.1/special/) exposes its functions
as traits on float and double atop `libm`
[`puruspe`](https://crates.io/crates/puruspe) is also in this space, author is that of `peroxide` (which to me, it feels like Matlab in Rust).

I don't know what direction to take yet, but some input on if this would be an issue someone actually stumbles on would be nice to know. If this is fruitful, would also try to get input from stainless-steel and Axect. 

---

Here's a summary on what we actually use and whether its available elsewhere. Note that no_std environment will typically imply `libm` version of it could be consumed by a user of `statrs`.

| Function statrs needs | Where used | Hand-rolled? | In `libm`? | In `puruspe`? | In `special`? |
|---|---|---|---|---|---|
| `erf` / `erfc` | `function/erf.rs` | Boost-derived rational/continued-fraction (`erf_impl`) | `libm::erf`, `libm::erfc` | `erf`, `erfc` | `Error::error`, `Error::compl_error` |
| `erf_inv` / `erfc_inv` | `function/erf.rs` | Boost-derived (`erf_inv_impl`) | - | `inverf`, `inverfc` | `Error::inv_error` only (no `erfc_inv`) |
| `ln_gamma` | `function/gamma.rs` | Lanczos approx (Pugh, 2004) | `libm::lgamma` / `lgamma_r` | `ln_gamma` | `Gamma::ln_gamma` |
| `gamma` (Γ) | `function/gamma.rs` | Lanczos approx (Pugh, 2004) | `libm::tgamma` | `gamma` | `Gamma::gamma` |
| `digamma` / `inv_digamma` | `function/gamma.rs` | Algorithm AS 103 | - | - | `Gamma::digamma` only (no `inv_digamma`; also adds `trigamma`) |
| incomplete gamma (`gamma_ui/li/ur/lr`) | `function/gamma.rs` | continued-fraction (Cephes-style, uncited) | - | `gammp`, `gammq`, `invgammp` | `Gamma::inc_gamma` |
| `ln_beta`, `beta` | `function/beta.rs` | built from `ln_gamma` | - | `beta` | `Beta::ln_beta` only (no plain `beta`) |
| incomplete/regularized beta (`beta_inc`, `beta_reg`, `inv_beta_reg`) | `function/beta.rs` | Algorithm AS 63/64/109 | - | `betai`, `invbetai` | `Beta::inc_beta`, `Beta::inv_inc_beta` |
| `logistic`, `logit` | `function/logistic.rs` | trivial, built from `exp`/`ln` | - | - | - |
| `factorial`, `ln_factorial`, `binomial`, `multinomial` | `function/factorial.rs` | lookup table + `ln_gamma` | - | - | - |
| `harmonic`, `gen_harmonic` | `function/harmonic.rs` | definition | - | - | - |

agene0001:
`erf_impl` reconstructs `erfc(z)` for `z >= 0.5` as `exp(-z²)/z · (b + r(z))`, where `r` is a minimax rational fit and `b` a per-interval constant. All 13 `b` constants carry only 10 significant decimal digits:

```rust
0.3440242112,   // interval [0.5, 0.75)
0.419990927,    // interval [0.75, 1.25)
```

Solving for the `b` each interval's fit implies (at 60-digit precision) gives values constant to ~19 digits across each interval — so the rational fits are good to ~1e-19 and the truncated constants were the *only* error source. The recovered values are exactly `f32`-representable: Boost declared them as `float` literals and the port that statrs inherits transcribed the printed decimal. This PR restores the full values, e.g. `0.3440242112` → `0.3440242111682891845703125`.

It also compensates the squaring: `z * z` is rounded before `exp` sees it, costing ~z² ulps (~460 by z = 27). A Dekker product recovers the rounding error of the square exactly (Dekker rather than `f64::mul_add`, which is a slow software FMA on targets without the instruction, e.g. baseline x86-64).

Measured against mpmath at 45 dps over a 700-point sweep plus every interval boundary:

| | before | after |
|---|---|---|
| `erf` | ~1e-10 rel | 0.24 median / **1.1 max ulp** |
| `erfc` | ~1e-10 rel | 0.52 median / **2.8 max ulp** |
| `Normal::cdf` | ~5e-11 rel | 0.32 median |

This propagates to everything built on `erfc` — `Normal` cdf/sf, `LogNormal`, `Levy`. The suite masked it with `epsilon = 1e-9`/`1e-11` tolerances, now tightened to 4 ulp of each expected value.

Six reference literals in the tests were themselves fitted to the old output and are replaced with mpmath values: three `LogNormal::sf` expectations (off by up to 5.1e-12) and three `erfc_inv` expectations — one of which, `erfc_inv(1e-10)`, was off by 8.8e-9 and had needed `epsilon = 1e-7` to pass. `erf_inv`/`erfc_inv` themselves were already ≤2 ulp; only the expectations were wrong.

### DIFF
--- a/src/distribution/log_normal.rs
+++ b/src/distribution/log_normal.rs
@@ -750,13 +750,13 @@ mod tests {
         test_absolute(-0.1, 0.1, 1.0, 1e-107, sf(0.1));
 
         // Wolfram Alpha:: SurvivalFunction[ LogNormalDistribution(-0.1, 0.1), 0.8]
-        test_absolute(-0.1, 0.1, 0.890919989231123, 1e-14, sf(0.8));
+        test_absolute(-0.1, 0.1, 0.890919989236242017902656326193, 1e-14, sf(0.8));
 
         // Wolfram Alpha:: SurvivalFunction[LogNormalDistribution[1.5, 1], 0.8]
-        test_absolute(1.5, 1.0, 0.957568715612642, 1e-14, sf(0.8));
+        test_absolute(1.5, 1.0, 0.957568715614411289470712253063, 1e-14, sf(0.8));
 
         // Wolfram Alpha:: SurvivalFunction[ LogNormalDistribution(2.5, 1.5), 0.1]
-        test_absolute(2.5, 1.5, 0.9993169594777358, 1e-14, sf(0.1));
+        test_absolute(2.5, 1.5, 0.999316959477792115067727898746, 1e-14, sf(0.1));
     }
 
     #[test]
--- a/src/function/erf.rs
+++ b/src/function/erf.rs
@@ -594,82 +594,94 @@ fn erf_impl(z: f64, inv: bool) -> f64 {
             (
                 evaluate::polynomial(z - 0.5, ERF_IMPL_BN)
                     / evaluate::polynomial(z - 0.5, ERF_IMPL_BD),
-                0.3440242112,
+                0.3440242111682891845703125,
             )
         } else if z < 1.25 {
             (
                 evaluate::polynomial(z - 0.75, ERF_IMPL_CN)
                     / evaluate::polynomial(z - 0.75, ERF_IMPL_CD),
-                0.419990927,
+                0.4199909269809722900390625,
             )
         } else if z < 2.25 {
             (
                 evaluate::polynomial(z - 1.25, ERF_IMPL_DN)
                     / evaluate::polynomial(z - 1.25, ERF_IMPL_DD),
-                0.4898625016,
+                0.489862501621246337890625,
             )
         } else if z < 3.5 {
             (
                 evaluate::polynomial(z - 2.25, ERF_IMPL_EN)
                     / evaluate::polynomial(z - 2.25, ERF_IMPL_ED),
-                0.5317370892,
+                0.5317370891571044921875,
             )
         } else if z < 5.25 {
             (
                 evaluate::polynomial(z - 3.5, ERF_IMPL_FN)
                     / evaluate::polynomial(z - 3.5, ERF_IMPL_FD),
-                0.5489973426,
+                0.548997342586517333984375,
             )
         } else if z < 8.0 {
             (
                 evaluate::polynomial(z - 5.25, ERF_IMPL_GN)
                     / evaluate::polynomial(z - 5.25, ERF_IMPL_GD),
-                0.5571740866,
+                0.55717408657073974609375,
             )
         } else if z < 11.5 {
             (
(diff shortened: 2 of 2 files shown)


day01:
isnt better keep erf fixes in https://github.com/rust-lang/compiler-builtins/tree/main/libm/src/math ?

agene0001:
Fixing it there wouldn't change anything here, because statrs doesn't use libm's `erf`.

`src/function/erf.rs` is a self-contained implementation — 28 of its own rational-approximation coefficient tables, inherited from the Math.NET Numerics port the crate began as. `libm` *is* in the dependency tree, via `num-traits`' `libm` feature, but only for basic float ops under `no_std`; `erf.rs` never calls into it. Rust's `std` doesn't expose `erf` at all, so a stats crate has to carry its own either way.

The constants in this PR are interval boundaries in that implementation — the cutoffs selecting which rational approximation to use, e.g. `0.3440242112` → `0.3440242111682891845703125`. libm's `erf` is FDLIBM-derived: a different algorithm with different constants, so there's nothing there these correspond to.

Switching statrs to libm's `erf` outright would be a real option, but it's a much bigger change — different accuracy characteristics, and a behavioural change for every distribution built on it — so probably its own discussion rather than part of this fix.

day01 (approved):

YeungOnion:
still deferring the mathlib discussion in #440 I think this is okay. I'm not quite sure if these numbers came from Boost, but it would be fair to assume that if we had 10 digits matching the same, then we would have been following some other reference. Will take these in good faith, but was actually thinking we might start to get more opinionated on this and maybe use criterion to inspect some of these distributions, but it does mean I may need a way to get reference values at runtime.

PR #414, opened 2026-07-26
fix(density): order-independent neighbourhood radius + reusable DensityEstimator

agene0001:
Two things, one file family (`src/density/`):

**1. A latent ordering bug.** `knn_pdf`/`kde_pdf` read the neighbourhood radius as `neighbors.last().unwrap()`, relying on an ordering `KdTree::within` never documented. kdtree 0.7 happened to return `into_sorted_vec()`; **0.8 returns raw `BinaryHeap` order**, so the last element is an arbitrary neighbour and both densities get scaled by the wrong ball volume. (`nearest()` still sorts in 0.8, which is why only the `Some(bandwidth)` path broke.) `nearest_neighbors` now returns `NearestNeighbors { squared_distances, k, radius }` with `radius` computed as an explicit maximum, documented as order-independent. Verified under both 0.7.0 and 0.8.1; the requirement moves to 0.8.1 and `Cargo.lock.MSRV` is regenerated with `cargo hack check --rust-version`.

The Monte Carlo tests couldn't reliably catch this, so it's pinned deterministically: 7 hand-placed samples, query at 0, squared radius 4 ⇒ exactly k = 5, radius = 2, `knn_pdf == 5/28`. Reading an arbitrary neighbour's distance gives 5/14.

**2. `DensityEstimator`: build the tree once.** `nearest_neighbors` rebuilt the k-d tree on every call — ~59% of a single `knn_pdf` at n = 1e5. The estimator owns the tree:

| 200-point grid | free fns | `DensityEstimator` | |
|---|---|---|---|
| `knn_pdf`, n = 10⁴ | 32.4 ms | 4.4 ms | 6.8x |
| `knn_pdf`, n = 10⁵ | 305.8 ms | 44.1 ms | **6.9x** |
| `kde_pdf` (20 pts), n = 10⁵ | 35.1 ms | 7.4 ms | 4.8x |

`knn_pdf`/`kde_pdf` remain as thin wrappers, and a test asserts the estimator is bit-identical to them over 40 grid points × 3 bandwidths. The tree dimension now comes from the samples rather than the query point, so a mismatched query is rejected instead of silently building a wrong-shaped tree. A `density_grid_200` criterion group tracks the rebuild-vs-prepared comparison.

### DIFF
--- a/Cargo.lock.MSRV
+++ b/Cargo.lock.MSRV
@@ -335,12 +335,12 @@ dependencies = [
 
 [[package]]
 name = "kdtree"
-version = "0.7.0"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0f0a0e9f770b65bac9aad00f97a67ab5c5319effed07f6da385da3c2115e47ba"
+checksum = "9cd666c2aae5dde60d2f9bfa4e1f8c17698fce13d599c0f9b76725641fec13d9"
 dependencies = [
  "num-traits",
- "thiserror 1.0.69",
+ "thiserror",
 ]
 
 [[package]]
@@ -706,7 +706,7 @@ dependencies = [
  "nalgebra",
  "num-traits",
  "rand",
- "thiserror 2.0.19",
+ "thiserror",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


YeungOnion:
Closing — the bug this targets (neighbourhood radius taking the last element instead of the maximum) is already fixed on main independently, via #443/#444, with `neighborhood_radius` now using `max_by` and covered by `neighborhood_radius_uses_farthest_distance`. This PR's version also predates the `DensityEstimator` refactor those PRs introduced, so it doesn't apply against current main.

PR #417, opened 2026-07-26
fix: scale beta_reg's continued-fraction bound with its parameters

agene0001:
> **Stacked on #416** (whose accurate `ln_gamma` sets this PR's error floor); review only the last commit until that merges.

The Lentz recurrence in `checked_beta_reg` was capped at a **fixed 140 iterations** and silently returned whatever it had reached. The recurrence is slowest at the centre of the distribution, where the worst case over `x` grows like `5·min(a,b)^(1/3)` (measured over a,b ∈ [1e2, 1e18] × 4001 values of x), so past `min(a,b) ≈ 1.5e4` the answer was simply wrong. Against the exact identity `I_½(a,a) = ½`:

| a = b | value returned | rel. error |
|---|---|---|
| 1e5 | 0.49999969504 | 6.1e-7 |
| 1e6 | 0.49121972700 | **1.8e-2** |
| 1e7 | 0.21285001452 | **5.7e-1** |

`Binomial::new(0.5, 2e6).cdf(1e6)` returned 0.4916 against a true 0.50028 — this caps `StudentsT`, `FisherSnedecor`, and `Beta` cdfs too.

The bound now scales as `8·min(a,b)^(1/3)` with headroom, clamped to bound worst-case work at ~10 ms. The loop still exits on convergence, so ordinary calls are unchanged (`beta_reg(2.5, 2.5, 0.5)`: 150 ns vs 145 ns).

Plus four robustness fixes found by probing the parameter space rather than sweeping accuracy:

- an underflowed prefix short-circuits to the corresponding endpoint — exact, and avoids `0.0 × h`, which is NaN whenever the recurrence overflowed (`beta_reg(1e300, 1e-300, 0.5)` was NaN before, on both sides of the bound change);
- `x == 0` / `x == 1` return 0/1 directly instead of via the symmetry test, which mapped `x == 0` to 1.0 once `a + b` overflowed;
- the symmetry threshold `(a+1)/(a+b+2)` is computed scaled when `a + b` overflows, otherwise it collapses to 0 and sends every `x` down the transformed branch;
- the result stays in `[0, 1]` (it's a probability; `a = b = 1e20` previously returned −2.56), falling back to the concentrated-limit step function if the truncated recurrence went non-finite.

Regression tests use exact identities needing no reference data — `I_½(a,a) = ½` and `I_x(a,b) + I_{1−x}(b,a) = 1` (both fail on the old fixed bound) — plus a 12×12×6 extreme-parameter grid asserting the result stays a finite probability.

### DIFF
--- a/src/function/beta.rs
+++ b/src/function/beta.rs
@@ -137,6 +137,18 @@ pub fn beta_reg(a: f64, b: f64, x: f64) -> f64 {
 /// `b` is the second beta parameter, and `x` is the upper limit of the
 /// integral.
 ///
+/// # Remarks
+///
+/// Relative accuracy degrades as `a + b` grows, because the leading factor is
+/// evaluated as `exp(ln_gamma(a + b) - ln_gamma(a) - ln_gamma(b) + ...)` and the
+/// cancellation in that exponent grows with `ln_gamma(a + b)`. Measured against
+/// the exact identity `I_{1/2}(a, a) == 1/2`, the relative error is about
+/// `6e-11` at `a = b = 1e4` and `3e-9` at `1e6`.
+///
+/// Past `min(a, b) ~ 1e16` the recurrence is truncated by its iteration bound and
+/// the result is unreliable, though still clamped to `[0, 1]`. Evaluation is
+/// bounded at roughly 10 ms in the worst case.
+///
 /// # Errors
 ///
 /// if `a <= 0.0`, `b <= 0.0`, `x < 0.0`, or `x > 1.0`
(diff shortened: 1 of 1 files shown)


YeungOnion:
Closing — #456 has since landed a full rewrite of `checked_beta_reg` (Temme's uniform asymptotic expansion for the large-parameter prefactor, a 100k-iteration continued fraction that returns `ConvergenceFailed` instead of silently truncating, plus the endpoint/overflow handling this PR added). Every problem here is covered, more rigorously and with reference-validated accuracy. No remaining gap for this PR to fill.

PR #420, opened 2026-07-26
perf: mark evaluate::polynomial #[inline]

agene0001:
Every in-crate caller passes a `const` coefficient slice, so with the length known at the call site LLVM fully unrolls the Horner fold — ~4x against the outlined loop when measured in isolation.

Honest scope: LLVM already inlines this within the crate, so in-crate timings don't move. The attribute makes that guaranteed rather than incidental, and extends it across codegen units and to external callers of this public function. Lowest priority of the series — feel free to close if not wanted.

### DIFF
--- a/src/function/evaluate.rs
+++ b/src/function/evaluate.rs
@@ -10,6 +10,10 @@
 /// # Remarks
 ///
 /// Returns 0 for a 0 length coefficient slice
+// `#[inline]` guarantees cross-crate/cross-codegen-unit inlining: callers pass
+// `const` coefficient slices, and with the length known at the call site LLVM
+// fully unrolls the Horner fold (~4x faster than the outlined loop).
+#[inline]
 pub fn polynomial(z: f64, coeff: &[f64]) -> f64 {
     coeff.iter().rev().fold(0_f64, |acc, val| acc * z + val)
 }


day01:
LLVM already inlines all internal calls. Optimizing direct external use of this helper seems too niche, so I don’t think this change is needed. wdyt @YeungOnion?

YeungOnion:
Agreed. And if the compiler for any reason wasn't inlining it then would be interesting to learn why.

day01:
Cross-crate inlining generally requires #[inline] when LTO is disabled. However, this helper is used almost entirely within statrs and is already inlined internally. I don’t think optimizing direct external calls justifies the attribute.

@agene0001 If this matters for your downstream workload, enabling LTO is an available option

src/function/evaluate.rs
```diff
@@ -10,6 +10,10 @@
 /// # Remarks
 ///
 /// Returns 0 for a 0 length coefficient slice
+// `#[inline]` guarantees cross-crate/cross-codegen-unit inlining: callers pass
```
day01, line 13: inline is not guarantee, it suggests for compiler.

agene0001:
Agreed — closing.

@day01 and @YeungOnion are both right that LLVM already inlines this internally, and optimising the cross-crate external call is too niche to justify the attribute. LTO is the right lever if it ever matters downstream.

Thanks for the review.

PR #421, opened 2026-07-26
feat: add ln_cdf and ln_sf with tail-accurate implementations

### LINKED ISSUE #338: ln_cdf & ln_sf
Hi Statrs,

Thanks for a fantastic library!

I do bioinformatics methods development, and in quite a few situations we find ourselves needing to evaluate in the tails of the distribution (we often need to apply quite aggressive multiple testing corrections). When working in R, we typically use the `log.p=TRUE` parameter for the cumulative distributions.

Are there any plans to provide log CDF functions in statrs? Obviously, default implementations that simply return the log of the regular `cdf` or `sf` functions would be a start, but as you probably know, it's not uncommon to use different numerical algorithms to implement `ln_cdf` or `ln_sf` to achieve good precision in the tails of the distribution.

Tom.

agene0001:
> **Stacked on #419** (top of the accuracy chain — this consumes `ln_gamma_prefix`/`ln_beta_prefix` from #418 and the erf machinery from #412); review only the last commit until those merge.

Implements the `log.p = TRUE` equivalent requested in #338. Every tail probability below ~1e-308 is currently irrecoverable — `Normal::sf(39)` is 0, so its log is `-inf` where the true value is −765.08 — and p-values at that scale are routine after multiple-testing correction, which is exactly the reporter's bioinformatics use case.

**API**: `ln_cdf`/`ln_sf` on `ContinuousCDF` and `DiscreteCDF` with defaults of `self.cdf(x).ln()` (mirroring the existing `ln_pdf`/`ln_pmf` convention), plus tail-accurate overrides for twelve distributions. The key observation making this cheap: the log is in most cases *already computed internally and then exponentiated* — the overrides return the exponent instead of destroying it.

- **`erf::ln_erfc`** (new): reuses the erfc interval machinery below 110, asymptotic series past it; finite until `x²` overflows near 1.3e154. **0.34 median / 1.71 max ulp** against mpmath over the full range. Backs `Normal`/`LogNormal`, with the argument-rounding compensation applied additively.
- **`gamma::ln_gamma_lr`/`ln_gamma_ur`** (new, + `checked_` variants): prefix stays in log domain; CF/series shared with the linear versions. Backs `Gamma`, `ChiSquared`, `Erlang`, `Poisson`.
- **`beta::ln_beta_reg`** (new, + `checked_`): likewise. Backs `Beta`, `Binomial`.
- **Exact closed forms**: `Exp` (`−λx`), `Weibull` (`−(x/λ)^k`), `Pareto` (`α·ln(scale/x)`), `Geometric` (`x·ln1p(−p)`).

**Validation**: against mpmath at 60 significant digits across all twelve distributions, the log values carry the probability's *relative* accuracy at ~1e-15 uniformly — including regimes where `cdf`/`sf` are hard zeros:

```text
Normal::ln_sf(39)              −765.083156564377544  (sf is exactly 0)
Normal::ln_sf(1000)            −500007.826694812184
ChiSquared(1)::ln_sf(5000)     −2504.48458784845137
Binomial(0.5, 1e4)::ln_cdf(1000) −3684.84454005928748  (cdf is exactly 0)
LogNormal(0,1)::ln_cdf(1e-100) −26515.8487124167112
```

**Bonus find**: the consistency test (`ln_cdf` vs `cdf().ln()` where both representable) caught a pre-existing bug — `Exp::cdf` used `1 − exp(−λx)`, which cancels catastrophically for small `λx` (~7 significant digits lost at `λx = 1e-10`). Fixed to `−expm1(−λx)`, matching what `Weibull::cdf` already did.

`StudentsT` and `FisherSnedecor` keep the default for now; their piecewise cdfs need dedicated treatment and can follow separately.

Closes #338


YeungOnion:
Heads up: #456 landed a full rewrite of `checked_beta_reg` since this PR was opened — the old `ln_beta_prefix`/Loader-saddle-point approach this PR's `ln_beta_reg` is built against is gone, replaced by Temme's uniform asymptotic expansion (`temme::beta_reg_temme`) plus `large_params::log_prefactor`. The `ln_beta_reg`/`checked_ln_beta_reg` addition here will need to be rebuilt against that new architecture rather than layered on the old prefactor.

Issue #202, opened 2024-02-25
Implement `erfcx` ?


avhz:
Hi :)

Any possibility of adding the scaled complementary error function:

$$
\texttt{erfcx}(z) = e^{x^2} \cdot \texttt{erfc}(z)
$$

such as in the C++ [Faddeeva library](http://ab-initio.mit.edu/wiki/index.php/Faddeeva_Package).

Thanks for the nice crate!

PR #423, opened 2026-07-26
feat: add the inverse Gaussian (Wald) distribution

### LINKED ISSUE #371: Inverse Gaussian distribution
Unless I missed it, or there's some naming confusion, it seems like this package is missing the [inverse Gaussian/Wald](https://en.wikipedia.org/wiki/Inverse_Gaussian_distribution) distribution?

Is there a technical reason why this might be difficult, or has there just not been demand for it so far? It doesn't seem like it would be any less straightforward than some existing implementations.

Assuming all that is true, would you accept a PR? Can't say whether I'll have the bandwidth so it's a bit of a hypothetical question right now, but it seems like it should be in reach by following the patterns elsewhere in this crate.

agene0001:
Closes #371. Also adds `erfcx`/`ln_erfcx`, which addresses #202.

> **Stacked on #421** (`feat/ln-cdf-ln-sf`), because this implements the
> `ln_cdf`/`ln_sf` introduced there. GitHub can't base a PR on another fork's
> branch, so the diff shown includes #421's commit — **review `30cd084` only**.
> Merging #421 first makes this diff reduce to that one commit.

## What's here

The full trait surface — `Continuous`, `ContinuousCDF` (including tail-accurate
`ln_cdf`/`ln_sf`), `Min`, `Max`, `Mode`, `Distribution`'s moments, and `rand`
sampling via Michael-Schucany-Haas.

## Why the cdf isn't written the textbook way

The conventional form is

```text
Phi(sqrt(l/x) (x/m - 1)) + e^(2l/m) Phi(-sqrt(l/x) (x/m + 1))
```

which cannot be evaluated as written: `e^(2l/m)` overflows once `l/m > ~355`,
even though the product it sits in is a probability bounded by 1. Both terms go
through `erfcx` instead, using

```text
2l/m - u_plus^2 == -u_minus^2
```

(verified exactly in mpmath across six parameter sets) so the large factor
cancels symbolically and is never formed:

```text
cdf = e^(-u_minus^2) [erfcx(-u_minus) + erfcx(u_plus)] / 2
sf  = e^(-u_minus^2) [erfcx( u_minus) - erfcx(u_plus)] / 2
```

whichever of the two is the smaller tail; the other is returned as its
complement. The `e^(-u_minus^2)` factor is Dekker-compensated for the rounding
of `u_minus^2`, and the log variants add that exponent exactly rather than
taking `ln` of a product - an earlier draft differenced two logs of magnitude
`5e4` to obtain `2e-5`, turning 0.7 ulp into `4e-7` of absolute error.

`erfcx(x) = e^(x^2) erfc(x)` is also what makes ratios of `erfc` values well
conditioned generally (#202). `ln_erfcx` is finite across the whole `f64`
range, where `erfc` underflows above ~27 and `erfcx` overflows below ~-27.

## Accuracy

Against mpmath at 60 digits over 1800 points - six parameter sets, ten decades
each. Restricted to points where the probability is a representable nonzero
`f64`, since that is the range a caller can observe a value in:

| | median | p99 | max |
|---|---|---|---|
| `pdf` | 2.54 ulp | 424 ulp | 583 ulp |
| `cdf` | 0.11 ulp | 462 ulp | 1179 ulp |
| `sf` | 0.12 ulp | 42196 ulp | 78686 ulp (~9e-12 rel) |
| `ln_cdf` | 1.1e-23 | 6.6e-14 | 2.4e-13 |
| `ln_sf` | 9.4e-20 | 6.5e-12 | 1.0e-11 |

The log rows are absolute error, which for a log *is* the relative error of the
probability it represents - so the implied probabilities are good to 12 and 11
significant digits respectively.

The linear-domain maxima are `exp` and cancellation limits, not defects in
these expressions: a relative error `eps` in an exponent becomes
`|exponent| * eps` in the result, and `sf` differences two `erfcx` values whose
ratio tends to `(x/m - 1)/(x/m + 1)`. Both are documented on the methods with a
pointer to the log variants.

Past the underflow point the logs stay within 1-2 ulp of the returned value.
Worth noting how that reads in absolute terms: the worst absolute error over the
*unrestricted* sweep is `1.2e-7`, but it occurs at `ln cdf = -3.4e8`, where one
ulp is itself `6e-8`. Neither metric is meaningful alone - ulp-of-the-log
explodes where the log approaches zero, and absolute error explodes where the
log is huge - so both are reported.

## Verification

`cargo fmt --check`, `cargo clippy --all-features --all-targets` clean,
`cargo hack check --feature-powerset --no-dev-deps` clean, `cargo hack check
--rust-version --locked` clean against `Cargo.lock.MSRV`, and the full test
suite across `--all-features`, default, `--no-default-features`, and
`--no-default-features --features rand`.


day01:
- SciPy 1.17.1 using `invgauss(mu / lambda, scale=lambda)`
- mpmath 1.4.1 at 100 decimal digits as the reference
- Apple M1 Max, Rust `--release`
- 15 representative cases covering ordinary values, large `lambda / mu`, underflow, and deep tails

This is not a reproduction of the full 1800-point sweep from the PR.

### Representative values

| Case | mpmath reference | This PR | SciPy |
|---|---:|---:|---:|
| `cdf(μ=1, λ=1, x=1)` | 0.6681020012231706064 | 0.668102001223170650 | 0.668102001223170650 |
| `cdf(1, 800, 0.9)` | 0.001517236267531762119 | 0.00151723626753176078 | 0.00151723626753178182 |
| `ln_cdf(1, 1, 0.001)` | -502.681165509344534 | -502.681165509344567 | -502.681165509344680 |
| `ln_sf(1, 1, 1000)` | -509.590912846424176 | -509.590912846424317 | -509.590912846350591 |
| `ln_sf(1, 1, 100000)` | -50016.4952145489501 | -50016.4952145489588 | -50016.4952138565204 |

### Maximum observed error

Relative error is used for `pdf`, `cdf`, and `sf`; absolute error is used for logarithmic results.

| Function | This PR | SciPy | Result |
|---|---:|---:|---|
| `pdf` | 1.19e-14 | 1.19e-14 | Tie |
| `cdf` | 5.41e-14 | 1.46e-13 | PR better |
| `sf` | 1.28e-13 | 7.36e-11 | PR better |
| `ln_cdf` | 4.91e-13 | 4.91e-13 | Tie |
| `ln_sf` | 8.61e-12 | 6.92e-7 | PR substantially better |

### Tail accuracy

| Case | PR error | SciPy error | Difference |
|---|---:|---:|---|
| `cdf(1, 800, 0.9)` | 8.81e-16 | 1.30e-14 | PR ~15x better |
| `ln_cdf(1, 1, 0.001)` | 3.27e-14 | 1.46e-13 | PR ~4.5x better |
| `ln_sf(1, 1, 1000)` | 1.40e-13 | 7.36e-11 | PR ~525x better |
| `ln_sf(1, 1, 100000)` | 8.61e-12 | 6.92e-7 | PR ~80,000x better |

The ordinary-domain results are effectively identical to SciPy. The significant improvement is in the deep right tail, especially `ln_sf`.

### Performance

Times are medians from repeated runs. SciPy vector timings are per element of a large array; scalar SciPy includes Python call overhead. mpmath uses 100-digit arithmetic and is intended only as a reference.

| Function | This PR, scalar Rust | SciPy vector, per element | SciPy scalar | mpmath |
|---|---:|---:|---:|---:|
| `pdf` | 5.34 ns | 18.6 ns | 26.9 µs | 28.8 µs |
| `cdf` | 21.8 ns | 40.7 ns | 28.9 µs | 203 µs |
| `sf` | 22.2 ns | 40.2 ns | 31.0 µs | 203 µs |
| `ln_cdf` | 25.9 ns | 39.7 ns | 30.1 µs | 215 µs |
| `ln_sf` | 26.7 ns | 38.3 ns | 29.0 µs | 218 µs |

Compared with vectorized SciPy, the PR was approximately 1.4x–3.5x faster per element on this machine. Scalar SciPy is not a direct implementation-level comparison because it is dominated by Python API overhead.

src/prec.rs
```diff
@@ -81,6 +92,68 @@ pub(crate) fn convergence(x: &mut f64, x_new: f64) -> bool {
     res
 }
 
+/// Splits the exact rounding error out of the product `a * b`
+/// (Dekker's algorithm): `a * b == p + dekker_product_err(a, b, p)` exactly.
+///
+/// Used instead of `f64::mul_add`, which falls back to a slow software FMA on
+/// targets without the hardware instruction (e.g. baseline x86-64).
+#[inline]
+pub(crate) fn dekker_product_err(a: f64, b: f64, p: f64) -> f64 {
```
day01, line 112: Why is the Dekker path unconditional? On AArch64/M1, `a.mul_add(b, -p)` compiles to a single `fnmsub`, while this implementation expands to roughly 30 FP instructions. The software fallback concern applies to targets without hardware FMA, such as generic x86-64, but this is a regression on FMA-capable targets.

Rust already selects hardware FMA based on [`target-cpu`](https://doc.rust-lang.org/rustc/codegen-options/index.html#target-cpu) / [`target-feature`](https://doc.rust-lang.org/rustc/codegen-options/index.html#target-feature), and the [`mul_add` documentation](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.mul_add) explicitly describes this performance distinction. Could we use `mul_add` where hardware FMA is available and keep Dekker only as the fallback, with benchmarks for both paths?

src/distribution/inverse_gaussian.rs
```diff
@@ -0,0 +1,659 @@
...
+fn exp_neg_square(u: f64) -> f64 {
+    let sq = u * u;
+    let err = crate::prec::dekker_product_err(u, u, sq);
+    (-sq).exp() * (1.0 - err)
+}
+
+/// `-u^2` exactly, as the leading term of a log-domain result.
+fn neg_square_exact(u: f64) -> f64 {
+    let sq = u * u;
+    -sq - crate::prec::dekker_product_err(u, u, sq)
```
day01: This does not preserve the product residual. Since `sq` is already the rounded `u * u`, evaluating `-sq - err` rounds back to `-sq` before the caller adds the logarithmic term. For ordinary finite inputs this function is therefore effectively identical to `-(u * u)`, so the log variants do not actually “add the exponent exactly”. The residual needs to be carried into the final sum, for example as `(-sq + log_term) - err`, or returned as a `(hi, lo)` pair.

src/distribution/inverse_gaussian.rs
```diff
@@ -0,0 +1,673 @@
...
+    /// quadratic it satisfies, then that root or its conjugate `mu^2 / y` is
+    /// selected with the appropriate probability.
+    fn sample<R: ::rand::Rng + ?Sized>(&self, rng: &mut R) -> f64 {
+        let z = crate::distribution::ziggurat::sample_std_normal(rng);
+        let nu = z * z;
+        let mnu = self.mu * nu;
+        // The textbook root `mu + (mu / 2 lambda) (mnu - sqrt(mnu (4 lambda +
+        // mnu)))` cancels catastrophically for `mnu >> lambda`; rationalizing
+        // gives this equivalent, subtraction-free form.
+        let y = self.mu - (2.0 * self.mu * mnu) / (mnu + (mnu * (4.0 * self.lambda + mnu)).sqrt());
```
day01: `z` can be exactly zero because the RNG produces a finite set of `f64` values. Substituting `z = 0` gives `nu = 0`, `mnu = 0`, and:

`y = mu - 0 / (0 + sqrt(0)) = mu - 0 / 0 = NaN`.

src/distribution/inverse_gaussian.rs
```diff
@@ -0,0 +1,673 @@
...
+impl Mode<Option<f64>> for InverseGaussian {
+    /// Returns the mode of the inverse Gaussian distribution
+    ///
+    /// # Formula
+    ///
+    /// ```text
+    /// μ [sqrt(1 + 9μ^2 / (4λ^2)) - 3μ / (2λ)]
+    /// ```
+    fn mode(&self) -> Option<f64> {
+        let r = self.mu / self.lambda;
```
day01:
```suggestion
        let ratio = self.lambda / self.mu;
```

src/distribution/inverse_gaussian.rs
```diff
@@ -0,0 +1,673 @@
...
+        Some(self.mu * ((1.0 + 2.25 * r * r).sqrt() - 1.5 * r))
```
day01:
```suggestion
        Some(self.lambda / ((ratio * ratio + 2.25).sqrt() + 1.5))
```

day01: This is algebraically equivalent to the documented mode formula, but avoids catastrophic cancellation.

example `mu = 1e8` and `lambda = 1`:

current:
`r = mu / lambda = 1e8`
`mode = 1e8 * (sqrt(1 + 2.25 * 1e16) - 1.5e8)`
`     = 1e8 * (1.5e8 - 1.5e8)`
`     = 0`

suggests:
`r = lambda / mu = 1e-8`
`mode = 1 / (sqrt(1e-16 + 2.25) + 1.5)`
`     ≈ 1 / 3`

src/prec.rs
```diff
@@ -81,6 +92,100 @@ pub(crate) fn convergence(x: &mut f64, x_new: f64) -> bool {
...
+/// `a * b == p + dekker_product_err(a, b, p)` exactly, for `p == a * b`.
+///
+/// Dispatches on whether the target has a hardware FMA. Where it does this is a
+/// single instruction; where it does not, `f64::mul_add` falls back to a
+/// software routine far slower than Veltkamp's split, so the split is used
+/// instead. Baseline x86-64 has no FMA -- it needs `-C target-feature=+fma` or
+/// a `-C target-cpu` that implies it -- while AArch64 has it in the base ISA.
+#[inline]
+pub(crate) fn dekker_product_err(a: f64, b: f64, p: f64) -> f64 {
+    #[cfg(any(target_arch = "aarch64", target_feature = "fma"))]
```
day01:
```suggestion
    #[cfg(any(
    target_feature = "fma",
    all(target_arch = "aarch64", target_feature = "neon")
))]
```

aarch as softfloat has software fma  n there dekker is good choice

day01: Shouldn't the FMA path be used on every target where `mul_add` lowers to a hardware instruction? It should generally be faster than the Dekker split there, not only on x86 with `fma` and AArch64.

day01 (changes requested):
if we re adding fma sup, it shouldnt be only for aarch, it is wider

day01:
i have a question @YeungOnion , if @agene0001 is changing erfcx, so maybe better to user erfcx as implementation of Cody? I made it times ago: https://github.com/day01/quant-opts/blob/ed23cd715da46da3b8cd52de09904e95b5b1a967/src/lets_be_rational/cody/optimized.rs#L119

imo it is faster and accuracy is better, but maybe i am biased.

agene0001:
All three addressed. Both numerical findings were real — thanks, these were good catches.

## Mode

Confirmed exactly as you described: `mu = 1e8, lambda = 1` returned `0` instead of `1/3`. Taken your reformulation.

One thing worth adding, which I hit while producing test references: evaluating the textbook form in **mpmath at 50 digits also returns zero** for `mu = 1e150`. It needs roughly 400 digits before it gives an answer at all. So the reference values in the new test come from the conjugate form, not from the documented one. The conjugate version also keeps `mu^2` out of the expression, so `mu = 1e300` now works where it previously overflowed.

## The NaN

Also real, and reachable rather than theoretical. The ziggurat builds `u = 2f - 1` from a 53-bit `f`, and `f == 0.5` is one of the values it can take — so `z == 0` has probability `2^-53` per draw. From there `0/0` propagated through both branches of the conjugate selection and the whole sample came back NaN.

Rather than special-casing it, pulling `sqrt(mnu)` out of the radical cancels it against the numerator:

```text
mu - 2 mu sqrt(mnu) / (sqrt(mnu) + sqrt(4 lambda + mnu))
```

The denominator is now bounded below by `2 sqrt(lambda)`, so it is finite throughout and tends to `mu` as `nu -> 0`, which is the right limit. It also stops `mnu (4 lambda + mnu)` overflowing for large `mnu`. Extracted as `msh_smaller_root` so the boundary is testable directly — no practical number of draws would reach it otherwise.

## FMA detection

You were right on both points, and the second one is the one I got wrong more broadly: I had treated AArch64 as unconditional and considered nothing outside x86.

Detection now lives in `build.rs` and sets `cfg(statrs_hardware_fma)` from the target's actual features:

| arch | condition |
|---|---|
| aarch64, arm64ec | `neon` |
| x86, x86_64 | `fma` |
| arm | `vfp4` |
| riscv32, riscv64 | `d` or `f` |
| powerpc, powerpc64 | `vsx` or `altivec` |
| loongarch32/64 | `d` or `f` |
| mips family | `fp64` |
| wasm32, wasm64 | `relaxed-simd` |

Each arm tests for the instruction rather than assuming the architecture implies it, so softfloat AArch64 correctly takes the split. The equivalence test between the two paths is unchanged and still runs both everywhere.

## On the Cody `erfcx`

Deferring to @YeungOnion on whether to adopt it. For what it is worth I would want to benchmark and measure both before anyone decides — mine is currently 0.24 median / 1.1 max ulp for `erf` and 0.52 / 2.8 for `erfc` against mpmath, and I have not measured yours, so I genuinely do not know which wins. Happy to run that comparison if it would help.

Reverting either numerical fix fails its test.

agene0001:
@day01 I ran the comparison. You were right on both counts, and it found a real bug in mine — thanks for pushing on this.

Setup: 7604 points against mpmath at 60 digits, covering `x` from -26 to 26 linearly, a log-spaced positive tail out to 1e300, and dense sampling either side of the 0.46875 threshold. Errors in ulp of the true value. M1 Max, `--release`.

## Accuracy, as it stood

| range | n | statrs med | statrs p99 | statrs max | cody med | cody p99 | cody max |
|---|---|---|---|---|---|---|---|
| x < -1 | 2885 | **31.04** | **359.05** | **463.08** | 0.46 | 2.11 | 3.23 |
| -1 ≤ x < 0 | 515 | 0.40 | 1.40 | 1.64 | 0.45 | 1.72 | 2.34 |
| 0 ≤ x ≤ 0.46875 | 391 | 0.45 | 1.53 | 1.94 | 0.53 | 2.00 | 2.73 |
| 0.46875 < x ≤ 4 | 673 | 0.34 | 1.32 | 2.08 | 0.59 | 2.93 | 4.21 |
| 4 < x ≤ 26 | 2541 | 0.35 | 1.11 | 1.34 | 0.35 | 1.13 | 1.42 |
| x > 26 | 599 | 0.26 | 0.60 | 0.98 | 0.26 | 0.57 | 0.69 |

Yours was ~140x better on the worst case for `x < -1`. So not bias — a genuine defect on my side.

## Why

My negative branch was `exp(x * x) * erfc(x)`. `x * x` carries up to half an ulp, and exponentiating multiplies that by `x^2`, so the error grows like `x^2 / 2` ulp — about 340 by `x = -26`. The comment even said `exp(x^2)` was "harmless up to `x^2 = 709`", which is true of overflow and false of accuracy.

Your `smoothened_exponential_of_positive_square` avoids this by splitting at a multiple of 1/16 so the leading square is exact. I've fixed it the equivalent way, recovering the residual with the Dekker split already used elsewhere in that file and folding it back as `exp(e) ~ 1 + e`:

| range | before | after | cody |
|---|---|---|---|
| x < -1, median | 31.04 | **0.38** | 0.46 |
| x < -1, max | 463.08 | **1.77** | 3.23 |

Overall, statrs is now 0.36 median / 2.08 max against Cody's 0.40 / 4.21, and better in every bucket.

## Speed

Yours is still clearly faster:

| workload | statrs | cody | |
|---|---|---|---|
| mixed sign | 10.26 ns | 4.01 ns | cody 2.56x |
| positive only | 4.79 ns | 2.17 ns | cody 2.21x |

The compensation cost me some of that (7.64 → 10.26 ns on the mixed path); the positive path is unchanged.

## So

The honest summary is that it's now an accuracy-versus-speed trade rather than a clear win either way: statrs is ~2x more accurate on worst case, Cody is ~2.3x faster. Which matters more is a call for @YeungOnion — for a stats library I'd lean accuracy, but I'm obviously not neutral, and 2x throughput on a hot path is not nothing.

Either way the bug your question exposed is fixed independently of that decision.

One methodological note in case it's useful: my first reference values for the regression test were wrong, because I generated them from decimal literals. `-26.6` differs from the nearest f64 by 1.4e-15, and this very amplification turns that into 687 ulp of `erfcx` — enough to fail a correct implementation. References for anything exponential have to come from the exact f64.

day01:
@agene0001 it looks, good. tests on my sites confirms the same state. about Cody it is not part of it. it is larger change.

YeungOnion:
Same note as #421, which this PR is stacked on: #456 rewrote `checked_beta_reg` (Temme expansion + `large_params`, replacing `ln_beta_prefix`), so the beta.rs content inherited through this stack will need rebasing onto that once #421 is reworked.

PR #430, opened 2026-08-07
fix: compute Hypergeometric pmf in log space for large populations

### LINKED ISSUE #426: Hypergeometric::pmf returns 0.0 or NaN for large populations
`pmf` divides three binomial coefficients held in `f64`. Once one of them passes `f64::MAX` the division happens between infinities, and nothing is signalled. `ln_pmf`, `cdf` and `mean` on the same object are correct.

```rust
let d = Hypergeometric::new(1030, 1, 515).unwrap();
d.pmf(0); // 0.0, exact value 1/2
d.pmf(1); // 0.0, exact value 1/2

let d = Hypergeometric::new(20000, 200, 300).unwrap();
d.pmf(0); // NaN, exact value 0.047931510683835526
d.pmf(3); // NaN, exact value 0.22687643066364876
```

`C(1030, 515)` exceeds `f64::MAX` while `C(1029, 515)` does not, so the first case overflows only the denominator and gives 0.0. Both coefficients overflow in the second, giving NaN.

The expected values above, computed with arbitrary-precision integers:

```
python3 -c "from math import comb; from fractions import Fraction as F; h=lambda N,K,n,x: F(comb(K,x)*comb(N-K,n-x),comb(N,n)); print([float(h(*a)) for a in [(1030,1,515,0),(1030,1,515,1),(20000,200,300,0),(20000,200,300,3)]])"
```

teddytennant:
## Problem

`Hypergeometric::pmf` multiplies and divides three binomial coefficients held in `f64`. Once any coefficient exceeds `f64::MAX`, the result is wrong:

- `Hypergeometric::new(1030, 1, 515)`: `pmf(0)` and `pmf(1)` return `0.0` (exact value `0.5`) because only the denominator overflows.
- `Hypergeometric::new(20000, 200, 300)`: `pmf(0)` and `pmf(3)` return `NaN` because numerator and denominator both overflow (`inf/inf`).

`ln_pmf`, `cdf`, and `mean` on the same objects were already correct.

## Root cause

Direct evaluation of `(K choose x) * (N-K choose n-x) / (N choose n)` via `factorial::binomial`, which returns a rounded `f64` that can be `+inf` for large arguments.

## Fix

Compute the PMF in log space, matching the Multinomial fix style already in this repo:

- `pmf(x)` delegates to `self.ln_pmf(x).exp()`
- `ln_pmf` returns `f64::NEG_INFINITY` for out-of-support `x` (`x > draws`, `x > successes`, or `x < min()`), so `pmf` returns `0.0` consistently
- `cdf`/`sf` already sum exp of log-binomial terms and are left unchanged

## Test plan

- [x] Existing hypergeometric unit tests (pmf cases that used exact equality now use absolute tolerance where log-space rounding differs at ~1e-15)
- [x] Regression tests from #426 for populations 1030 and 20000
- [x] Out-of-support guards for `pmf` / `ln_pmf`
- [x] `cargo test --lib` (774 passed, 2 ignored)

Fixes #426

### DIFF
--- a/src/distribution/hypergeometric.rs
+++ b/src/distribution/hypergeometric.rs
@@ -394,14 +394,11 @@ impl Discrete<u64, f64> for Hypergeometric {
     /// ```
     ///
     /// where `N` is population, `K` is successes, and `n` is draws
+    ///
+    /// Computed in log space via [`Self::ln_pmf`] so large binomial
+    /// coefficients that overflow `f64` still yield a finite probability.
     fn pmf(&self, x: u64) -> f64 {
-        if x > self.draws {
-            0.0
-        } else {
-            factorial::binomial(self.successes, x)
-                * factorial::binomial(self.population - self.successes, self.draws - x)
-                / factorial::binomial(self.population, self.draws)
-        }
+        self.ln_pmf(x).exp()
     }
 
     /// Calculates the log probability mass function for the hypergeometric
@@ -415,6 +412,9 @@ impl Discrete<u64, f64> for Hypergeometric {
     ///
     /// where `N` is population, `K` is successes, and `n` is draws
     fn ln_pmf(&self, x: u64) -> f64 {
+        if x > self.draws || x > self.successes || x < self.min() {
+            return f64::NEG_INFINITY;
+        }
         factorial::ln_binomial(self.successes, x)
             + factorial::ln_binomial(self.population - self.successes, self.draws - x)
             - factorial::ln_binomial(self.population, self.draws)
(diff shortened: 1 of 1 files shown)


teddytennant:
The Coverage (nightly) failure here is essentially unrelated to this change.

Recent nightly (1.99.0-nightly, 2026-08-06) started firing the `deprecated` lint on `f64::NAN` / `INFINITY` / `NEG_INFINITY` paths in the ~25 modules that do `use core::f64;` — the module import shadows the primitive in path resolution, so those resolve to the deprecated `core::f64::*` module constants rather than the associated constants. The Coverage job sets `RUSTFLAGS: -D warnings`, so it becomes a hard build failure.

It reproduces on a clean checkout of main (02fdab6) with nothing applied: 90 errors. This branch gives 91 — the one extra is the `f64::NEG_INFINITY` added in `hypergeometric.rs`, which follows the same convention as the existing code in `geometric.rs` and the rest of the crate. Fixing just that line would not make the job green. Without `-D warnings` everything compiles (warnings only). Last green Coverage run on main was Jul 31.

`cargo test --all-features` passes on this branch: 778 passed, 0 failed.

Happy to send a separate PR swapping those paths over to the associated constants (or dropping the `use core::f64;` imports) if you want Coverage unblocked — and I can rebase this on top of it.

YeungOnion:
good call, thanks for this fix and adjusting the tests for the exact integer versions. If someone finds a gotcha due to this, we may need to branch between log space and an actual rational implementation.

PR #431, opened 2026-08-07
fix: correct Geometric::inverse_cdf against its definition

### LINKED ISSUE #342: issue in default inverse_cdf, found with simple case of Geometric

teddytennant:
## Summary

`Geometric::inverse_cdf` used the closed form `ceil(ln1p(-x) / ln1p(-p))` alone. That quotient is only approximately integral at the step boundaries, so `ceil` often lands on the wrong side of the true quantile and breaks the discrete inverse definition (least `k` with `cdf(k) >= x`).

Over a sweep of 200 values of `p` and 400 of `k`, thousands of pairs failed `inverse_cdf(cdf(k)) == k` even though `cdf` separated adjacent integers.

## Root cause

Floating-point evaluation of the closed form is not exact at bucket edges. Being an override, this path was also unreachable by later improvements to the default discrete `inverse_cdf`.

## Fix

Keep the closed form as a fast candidate, then correct it against the definition:

- check `candidate`, `candidate - 1`, and `candidate + 1` with two or three `cdf` evaluations (the usual path);
- when `x` is within a few ulp of 1, `1 - x` loses significant bits and the closed form can be off without bound (cdf plateaus about `1/p` wide), so fall back to exact integer bisection.

## Test plan

- [x] `cargo test --lib geometric::`
- [x] `cargo test --lib` (775 passed, 2 ignored pre-existing pathological cases still marked `#[ignore]`)
- New regressions:
  - `test_inverse_cdf_round_trips_and_matches_definition`
  - `test_inverse_cdf_is_least_k_with_cdf_at_least_x`
  - `test_inverse_cdf_long_plateau`

Fixes #342

### DIFF
--- a/src/distribution/geometric.rs
+++ b/src/distribution/geometric.rs
@@ -132,6 +132,11 @@ impl DiscreteCDF<u64, f64> for Geometric {
     fn cdf(&self, x: u64) -> f64 {
         if x == 0 {
             0.0
+        } else if x == 1 {
+            // Mathematically cdf(1) = p. Evaluating via expm1/ln1p can
+            // undershoot by a ulp on some platforms (notably MSVC), which then
+            // makes inverse_cdf(p) pick k=2 via the definition check.
+            self.p
         } else {
             // 1 - (1 - p) ^ x = 1 - exp(log(1 - p)*x)
             //                 = -expm1(log1p(-p)*x))
@@ -153,6 +158,9 @@ impl DiscreteCDF<u64, f64> for Geometric {
         //           = exp(log1p(-p) * x)
         if x == 0 {
             1.0
+        } else if x == 1 {
+            // Mathematically sf(1) = 1 - p. Keep exact to match cdf(1) = p.
+            1.0 - self.p
         } else {
             ((-self.p).ln_1p() * (x as f64)).exp()
         }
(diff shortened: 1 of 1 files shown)


teddytennant:
The Coverage (nightly) failure here is unrelated to this change.

Recent nightly (1.99.0-nightly, 2026-08-06) started firing the `deprecated` lint on `f64::NAN` / `INFINITY` / `NEG_INFINITY` paths in the ~25 modules that do `use core::f64;` — the module import shadows the primitive in path resolution, so those resolve to the deprecated `core::f64::*` module constants rather than the associated constants. The Coverage job sets `RUSTFLAGS: -D warnings`, so it becomes a hard build failure.

It reproduces on a clean checkout of main (02fdab6) with nothing applied: 90 errors. This branch also produces exactly 90 — it adds no new instances. Without `-D warnings` it compiles fine (warnings only). Last green Coverage run on main was Jul 31.

`cargo test --all-features` passes on this branch: 780 passed, 0 failed.

Happy to send a separate PR swapping those paths over to the associated constants (or dropping the `use core::f64;` imports) if you want Coverage unblocked.

YeungOnion:
thanks for this fix!

PR #436, opened 2026-08-09
fix: `Beta::sf` returning exactly 1.0 for tiny `x`

FBruzzesi:
Closes #432

## Problem

`Beta::sf` formed the complement before calling `beta_reg`:

```rust
beta::beta_reg(self.shape_b, self.shape_a, 1.0 - x)
```

For `x < ~1.1e-16` the expression `1.0 - x` rounds to exactly `1.0`, so `sf` returned `1.0` regardless of the shapes. When `shape_a < 1` the lower tail carries real mass and the answer was wrong: `Beta::new(0.05, 0.05).unwrap().sf(1e-16)` returned `1.0` where the true value is `0.92045095674306394`.

## Fix

For `x` below the continued fraction split point used inside `beta_reg`, namely `(a + 1) / (a + b + 2)`, compute the survival function as `1.0 - beta_reg(a, b, x)` instead.

Results after this change:

| shapes | x | before | after | mpmath (`mp.dps = 60`) |
| --- | --- | --- | --- | --- |
| (0.05, 0.05) | 1e-16 | 1.0 | 0.92045095674306399 | 0.92045095674306394 |
| (0.05, 0.95) | 1e-20 | 1.0 | 0.90041072647564402 | 0.90041072647564386 |
| (0.05, 0.95) | 1e-100 | 1.0 | 0.99999004107264755 | 0.99999004107264756 |

## Tests

Added dedicated "tiny" `x` tests.

## Out of scope

`FisherSnedecor`, `NegativeBinomial` and `StudentsT` call `beta_reg` before form a complement before calling `beta_reg`  may have analogous tail issues; this PR fixes only `Beta::sf` per the issue.

PR #438, opened 2026-08-11
fix: restore exact distribution boundary handling

### LINKED ISSUE #437: bug(regression): `Beta::ln_pdf` returns `-inf` for every `x` within `1e-9` of `1.0`; 0.18.0 returned the finite value
On 0.19.0, `Beta::ln_pdf(x)` returns `-inf` for every `x` in the band `1 - 1e-9 <= x < 1`, for every shape combination I probed, including shapes where the density diverges at the boundary and the true log-density is large and *positive*.

On 0.18.0 the same calls return finite values (that agree with mpmath to 15 digits).

`pdf(x)` is unaffected on the same points, so `pdf` and `ln_pdf` now disagree by an infinite amount on that band.

**Environment:** statrs `0.18.0` vs `0.19.0` side by side as renamed deps (`default-features = false`, features `["std"]` on 0.19), rustc 1.96.0-nightly, aarch64-apple-darwin, `--release`.

## Reproducer

```rust
use statrs::distribution::{Beta, Continuous};

fn main() {
    let x = 1.0 - 1e-9;
    for (a, b) in [(2.0, 3.0), (0.05, 0.05), (5.0, 0.001)] {
        let d = Beta::new(a, b).unwrap();
        println!("Beta({a}, {b}): ln_pdf({x}) = {:e}   pdf({x}) = {:e}", d.ln_pdf(x), d.pdf(x));
    }
}
```

On 0.19.0:

```
Beta(2, 3): ln_pdf(0.999999999) = -inf   pdf(0.999999999) = 1.1999999309233672e-17
Beta(0.05, 0.05): ln_pdf(0.999999999) = -inf   pdf(0.999999999) = 8.904406482310805e6
Beta(5, 0.001): ln_pdf(0.999999999) = -inf   pdf(0.999999999) = 9.815320421817441e5
```

The same program against 0.18.0:

```
Beta(2, 3): ln_pdf(0.999999999) = -3.8961625081668686e1   pdf(0.999999999) = 1.1999999309233672e-17
Beta(0.05, 0.05): ln_pdf(0.999999999) = 1.600205682254767e1   pdf(0.999999999) = 8.904406482310805e6
Beta(5, 0.001): ln_pdf(0.999999999) = 1.3796869938301022e1   pdf(0.999999999) = 9.815320421817441e5
```

(`pdf` is byte-identical across the two versions on all three points, so the disagreement is `ln_pdf`'s alone.)

The band edge is exact: `x = 1 - 1e-9` is the first affected point, one ulp below it the finite value returns.

## Expected

mpmath at `mp.dps = 50`, evaluating `(a-1) ln x + (b-1) ln(1-x) - ln B(a, b)` at the exact double `x = 1 - 1e-9`:

```
Beta(2, 3)       -> -38.961625081668685701
Beta(0.05, 0.05) ->  16.002056822547670151
Beta(5, 0.001)   ->  13.796869938301040045
```

so the 0.18.0 values are correct to the last digit or two, and `-inf` is wrong by an infinite amount.

<details>
<summary>mpmath (python) script</summary>

```python
from mpmath import mp, mpf, log, gamma

mp.dps = 50
x = mpf(1.0 - 1e-9)  # the exact f64 double
for a, b in [(2, 3), (mpf("0.05"), mpf("0.05")), (5, mpf("0.001"))]:
    ln_beta = log(gamma(a)) + log(gamma(b)) - log(gamma(a + b))
    print((a - 1) * log(x) + (b - 1) * log(1 - x) - ln_beta)
```

</details>

day01:
## Summary

This PR fixes regressions caused by using approximate floating-point equality for conditions that represent exact distribution boundaries or degenerate parameter values.

It includes the `Beta::sf` fix from #436 and extends the correction to the other affected distributions and functions.

## Root cause

`prec::ulps_eq!` uses a default absolute epsilon of `1e-9`. As a result, values close to exact boundaries, such as `p ≈ 1`, `shape ≈ 1`, or `x ≈ 1`, were incorrectly treated as exactly equal.

This caused nearby valid inputs to enter special-case branches intended only for exact mathematical identities or degenerate distributions.

## Regression coverage added

The new tests cover:

- `Beta::sf` for tiny `x` values from `1e-16` down to `1e-100`, using high-precision reference values.
- `Beta::ln_pdf` at `x = 1.0 - 1e-9` for multiple shape combinations.
- Correct `Beta::ln_pdf` infinities at `x = 0` and `x = 1`.
- `Beta` parameters close to, but not exactly, the uniform distribution for CDF, SF and PDF.
- `Binomial` with `p = 1.0 - 5e-10`, including PMF, log-PMF and entropy.
- Finite `Binomial` entropy for `n = 1000` when very small PMF terms underflow to zero.
- `Geometric` with probability close to one, including maximum support, log-PMF, skewness and inverse CDF.
- `Gamma` and `Weibull` densities at zero for shapes immediately above and below one.
- The exact special point of `Gamma` with an infinite rate.
- `InverseGamma` with shape close to one, checked against the general density formula.
- `digamma` close to, but not exactly at, a negative integer pole.
- `Weibull` mode for shapes below one.

## Credits

Thanks to @FBruzzesi for reporting the `Beta::ln_pdf` regression in #437 and for preparing the `Beta::sf` fix in #436, which is included in this PR. (cherry picked)

Closes #437.
Includes and extends #436.

### DIFF
--- a/src/distribution/beta.rs
+++ b/src/distribution/beta.rs
@@ -1,6 +1,5 @@
 use crate::distribution::{Continuous, ContinuousCDF, InverseCdfError};
 use crate::function::{beta, gamma};
-use crate::prec;
 use crate::statistics::*;
 #[cfg(not(feature = "std"))]
 use num_traits::Float as _;
@@ -143,7 +142,7 @@ impl ContinuousCDF<f64, f64> for Beta {
             0.0
         } else if x >= 1.0 {
             1.0
-        } else if prec::ulps_eq!(self.shape_a, 1.0) && prec::ulps_eq!(self.shape_b, 1.0) {
+        } else if self.shape_a == 1.0 && self.shape_b == 1.0 {
             x
         } else {
             beta::beta_reg(self.shape_a, self.shape_b, x)
@@ -165,14 +164,14 @@ impl ContinuousCDF<f64, f64> for Beta {
             1.0
         } else if x >= 1.0 {
             0.0
-        } else if prec::ulps_eq!(self.shape_a, 1.0) && prec::ulps_eq!(self.shape_b, 1.0) {
+        } else if self.shape_a == 1.0 && self.shape_b == 1.0 {
             1. - x
         } else if x < (self.shape_a + 1.0) / (self.shape_a + self.shape_b + 2.0) {
             // Below the continued fraction split point of `beta_reg`,
             // `beta_reg(b, a, 1 - x)` reduces to `1 - beta_reg(a, b, x)`;
             // computing the complement here instead avoids `1.0 - x`
             // rounding to 1.0 for tiny x (< ~1.1e-16), which would lose
-            // the lower tail entirely. See #432 
+            // the lower tail entirely. See #432
             1.0 - beta::beta_reg(self.shape_a, self.shape_b, x)
         } else {
             beta::beta_reg(self.shape_b, self.shape_a, 1.0 - x)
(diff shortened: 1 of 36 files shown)


day01:
@YeungOnion regression, can you make a hotfix?

YeungOnion:
@day01 thanks for the regression fixes
@FBruzzesi thanks for the 1-x fixes too

PR #442, opened 2026-08-12
fix: stabilize Gamma PDF for finite in-range parameters

### LINKED ISSUE #422: `Gamma::pdf` returns `NaN` for ordinary, finite, in-range parameters
```rust
use statrs::distribution::{Continuous, Gamma};

fn main() {
    let g = Gamma::new(80.0, 1e-5).unwrap();
    let x = 8e6_f64;
    println!("pdf({}) = {}", x, g.pdf(x));
    println!("ln_pdf({}) = {}", x, g.ln_pdf(x));
    println!("exp(ln_pdf) = {}", g.ln_pdf(x).exp());
}
```

```
pdf(8000000) = NaN
ln_pdf(8000000) = -14.623918976753487
exp(ln_pdf) = 0.0000004455666577034725
```

`shape = 80`, `rate = 1e-5`, and `x = 8e6` (the mean of the distribution) are all finite and within the domain `Gamma::new` accepts. `pdf` returns `NaN`, while `ln_pdf` on the same distribution and the same `x` gives a finite value whose exponential is a normal, non-zero density.

Tested on statrs 0.19.0, with both a debug and a release build.

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

day01:
## Summary

- evaluate the non-trivial Gamma PDF through `ln_pdf().exp()` to avoid overflow/underflow in separately evaluated factors
- preserve exact boundary behavior for `x = 0`, `x = +inf`, and `shape = 1`
- use a compensated sum for `ln(rate) + ln(x)` before multiplying by `shape`
- add a regression test for `Gamma(80, 1e-5).pdf(8e6)`

## Root cause

The previous formula evaluated `rate.powf(shape)` independently. For valid finite parameters this factor could underflow to zero while the remaining factors compensated it mathematically, producing `NaN` instead of a finite density.

## TDD and independent numerical references

The regression case is `Gamma(shape=80, rate=1e-5).pdf(8e6)`.

| Implementation | Result | Relative error vs. mpmath | Outcome |
|---|---:|---:|---|
| statrs (this PR) | `4.455666577034977970e-7` | `2.63e-14` | correct |
| Boost.Math 1.90 | `4.455666577035076430e-7` | `4.22e-15` | agrees |
| R 4.2.1 `dgamma` | `4.455666577035096020e-7` | `1.81e-16` | agrees |
| mpmath 1.4.1, 136-bit | `4.45566657703509521452812298721e-7` | reference | reference |
| SciPy 1.18.0 | `4.455666577035271250e-7` | `3.95e-14` | agrees |

Upstream statrs returns `NaN`; the regression test is red on upstream and green with this change. R and Boost.Math are more accurate for this input; this change is more accurate than SciPy.

## Performance

| Case | Upstream | This change | Difference |
|---|---:|---:|---:|
| issue input: `Gamma(80, 1e-5).pdf(8e6)` | 34.58 ns, returns `NaN` | 30.86 ns, finite result | 10.8% faster |
| typical finite input | 37.74 ns | 27.06 ns | 28.3% faster |
| large-shape input | 26.70 ns | 27.32 ns | 2.3% slower |

| Implementation | Execution mode | Median time per call | Time relative |
|---|---|---:|---:|
| statrs (this PR) | Rust release, default `std` features | 26.05 ns | 1.00x |
| Boost.Math 1.90 | C++17, `-O3 -march=native` | 27.54 ns | 1.06x |
| R 4.2.1 `dgamma` | scalar R call | 755 ns | 29.0x |
| mpmath 1.4.1, 136-bit | scalar Python call | 16.95 us | 650.7x |
| SciPy 1.18.0 `gamma.pdf` | scalar Python call | 21.91 us | 841.0x |

## Related issue

Fixes #422 — [`Gamma::pdf` returns `NaN` for ordinary, finite, in-range parameters](https://github.com/statrs-dev/statrs/issues/422).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Gamma distribution probability calculations for edge cases, including zero-valued inputs.
  * Fixed PDF evaluation when rate-related calculations underflow or nearly cancel.
  * Improved handling of infinite shape and rate values.
  * Enhanced numerical stability and consistency between Gamma PDF and log-PDF results across extreme inputs.
  * Improved reliability for very small, very large, and subnormal numeric values.


day01:
## Correctness

| shape | rate | R `dgamma` | SciPy | mpmath (50 dps) | statrs upstream (pre-PR) | statrs PR #442 (before fix) | statrs PR #442 (after fix) |
|---|---|---|---|---|---|---|---|
| ∞ | 1.0 | **0.0** | **0.0** | NaN¹ | NaN | 0.0 ✓ | **0.0 ✓** |
| 0.5 | ∞ | NaN | NaN | ZeroDivisionError² | NaN | inf ✗ | **NaN ✓** |
| 2.0 | ∞ | NaN | NaN | NaN | NaN | 0.0 ✗ | **NaN ✓** |
| 1.0 | ∞ | NaN | NaN | NaN | NaN | NaN ✓ | **NaN ✓** |

## Performance (`pdf(0.0)`, per call)

| shape | rate | R | SciPy | mpmath (50 dps) | statrs upstream (pre-PR) | statrs PR #442 (before fix) | statrs PR #442 (after fix) |
|---|---|---:|---:|---:|---:|---:|---:|
| ∞ | 1.0 | 0.73 µs | 67.7 µs | 11.1 µs | 2.0 ns | 3.3 ns | **2.7 ns** |
| 0.5 | ∞ | 11.9 µs | 36.7 µs | 7.6 µs | 9.9 ns | 6.1 ns | **3.7 ns** |
| 2.0 | ∞ | 11.7 µs | 36.8 µs | 16.8 µs | 19.0 ns | 4.3 ns | **3.7 ns** |
| 1.0 | ∞ | 11.9 µs | 36.6 µs | 16.9 µs | 3.7 ns | 4.4 ns | **3.7 ns** |

day01:
@coderabbitai review

src/distribution/gamma.rs
```diff
@@ -392,10 +389,25 @@ impl Continuous<f64, f64> for Gamma {
...
+            }
         } else if x.is_infinite() {
             f64::NEG_INFINITY
         } else {
-            self.shape * self.rate.ln() + (self.shape - 1.0) * x.ln()
+            let ln_rate = self.rate.ln();
+            let ln_x = x.ln();
+            let ln_product = ln_rate + ln_x;
+            let virtual_ln_x = ln_product - ln_rate;
+            let ln_product_error = (ln_rate - (ln_product - virtual_ln_x)) + (ln_x - virtual_ln_x);
```
YeungOnion: Why is compensation of ln(a) + ln(b) the chosen accuracy mechanism rather than scaling/decomposing the product so that we can preserve multiplication's behavior without sacrificing dynamic range? I made a push as a test case and as a suggested fix.

YeungOnion:
well this does now fail a doctest at 1e-15 absolute error...

day01:
You're right — decomposing the product is the better approach here because it preserves the dynamic range without relying on compensated `ln(rate) + ln(x)`. I kept the `frexp`

PR #444, opened 2026-08-12
Update kdtree and fix KDE radius

day01:
## Motivation

`kdtree` 0.8.1 intentionally changed `within()` to return unordered results in [mrhooray/kdtree-rs#85](https://github.com/mrhooray/kdtree-rs/pull/85). The existing KDE and KNN implementations used the last result as the neighborhood radius, which produces incorrect densities after the dependency update.

## Changes

- update `kdtree` from 0.7.0 to 0.8.1
- compute the neighborhood radius from the farthest returned distance instead of relying on result order
- add a regression test with deliberately unordered distances
- update the MSRV lockfile

Refs:
https://github.com/statrs-dev/statrs/pull/428

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved density estimates by consistently calculating neighborhood radius from the farthest neighboring sample.
  * Added handling for empty neighbor-distance data.

* **Maintenance**
  * Updated the optional spatial-indexing dependency and lockfile entries.

* **Tests**
  * Added coverage for neighborhood radius calculation and farthest-distance selection.


YeungOnion:
Thanks for the fix! How did you catch this regression?

day01:
compare 1-1 heh terrible :D

PR #446, opened 2026-08-12
fix: Unblock release-plz

day01:
Remove the tracked `tests/no_std/Cargo.lock` and stop using `--locked` for the no_std runtime build.

`release-plz` copies the repository using gitignore rules. Since `*.lock` is ignored, the tracked nested lockfile is omitted from the temporary copy and release-plz treats the repository as dirty.

The no_std runtime crate is only used for CI verification, so its dependencies can be resolved during the build.


day01:
@YeungOnion post #403 regression

YeungOnion:
I see how we can resolve deps onto a no_std target without locked, but how'd you catch the behavior with regards to release-plz? Part of the distributed package?

day01:
@YeungOnion lockfile is part of test, we have gitignore in project and it is correct for packages. i commited only test lockfile to achieve repeatable behaviour, but plz has problem with that. it is safe change without impact on plz or project

PR #447, opened 2026-08-13
fix: Stabilize beta CDF and extreme quantiles

### LINKED ISSUE #434: bug: `Beta::cdf` loses accuracy above shapes of ~1e4 and returns values outside [0, 1] by 1e8
Same context and premise as in #432 

## The issue

`Beta(s, s).cdf(0.5)` is exactly `0.5` for every `s`, by the symmetry `I_{1/2}(s, s) = 1/2`, so this case needs no oracle. What I get:

| `s` | `Beta(s, s).cdf(0.5)` | true | rel. err. |
| --- | --- | --- | --- |
| `1e2` | `0.5000000000001004` | `0.5` | `2.0e-13` |
| `1e3` | `0.5000000000000433` | `0.5` | `8.7e-14` |
| `1e4` | `0.5000000000286743` | `0.5` | `5.7e-11` |
| `1e5` | `0.49999969504052444` | `0.5` | `6.1e-7` |
| `1e6` | `0.49121972699639316` | `0.5` | `1.8e-2` |
| `1e7` | `0.21285001452392216` | `0.5` | `5.7e-1` |
| `1e8` | `-1.1473179165955987` | `0.5` | outside `[0, 1]` |

The last row is negative, so `cdf(0.5).ln()` is `NaN` there. `sf(0.5)` returns the same number, so `cdf + sf` is not `1`.

**Environment:** statrs `0.19.0` (`default-features = false, features = ["std"]`), rustc 1.97.1, aarch64-apple-darwin, `--release`.

## Reproducer

```rust
use statrs::distribution::{Beta, ContinuousCDF};

fn main() {
    for s in [1e2f64, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8] {
        let d = Beta::new(s, s).unwrap();
        println!("Beta({s:e}, {s:e}).cdf(0.5) = {:<22} ln = {:?}", d.cdf(0.5), d.cdf(0.5).ln());
    }
}
```

```
Beta(1e2, 1e2).cdf(0.5) = 0.5000000000001004     ln = -0.6931471805597446
Beta(1e3, 1e3).cdf(0.5) = 0.5000000000000433     ln = -0.6931471805598587
Beta(1e4, 1e4).cdf(0.5) = 0.5000000000286743     ln = -0.6931471805025967
Beta(1e5, 1e5).cdf(0.5) = 0.49999969504052444    ln = -0.6931477904790825
Beta(1e6, 1e6).cdf(0.5) = 0.49121972699639316    ln = -0.7108637421323526
Beta(1e7, 1e7).cdf(0.5) = 0.21285001452392216    ln = -1.547167518472497
Beta(1e8, 1e8).cdf(0.5) = -1.1473179165955987    ln = NaN
```

## Expected

`0.5` in every row.

## Possibly related

I only tested `Beta` here. If `Binomial` reaches the incomplete beta for `P(X > k)`, large `n` may land in the same range, but I have not checked that.

### LINKED ISSUE #435: bug: `inv_beta_reg` panics, does not return, and is non-monotone for small `x`
Same context and premise as in #432

---

Three separate issues in the lower tail, so please treat them independently. All three are also reachable from `Beta::inverse_cdf`.

| # | Issue | Trigger |
| --- | --- | --- |
| 1 | panic | `x` below about `1.6e-162` |
| 2 | "never" returns | `x` below about `1e-50` at `(a, b) = (200, 2)` |
| 3 | constant and non-monotone results | any root below about `1e-17` |

**Environment:** statrs `0.19.0` (`default-features = false, features = ["std"]`), rustc 1.97.1, aarch64-apple-darwin, `--release`, so `debug_assert!` is off.

## Issue 1: panic for very small `x`

```rust
use statrs::function::beta::inv_beta_reg;

fn main() {
    println!("{}", inv_beta_reg(200.0, 2.0, 1e-165)); // panics immediately
}
```

```bash
thread 'main' panicked at .../statrs-0.19.0/src/function/beta.rs:132:31:
called `Result::unwrap()` on an `Err` value: XOutOfRange
```

Also reachable from the public API:

```rust
use statrs::distribution::{Beta, ContinuousCDF};
Beta::new(200.0, 2.0).unwrap().inverse_cdf(1e-170); // panics
```

## Issue 2: no termination for a band of `x`

```rust
use statrs::function::beta::inv_beta_reg;

fn main() {
    println!("{}", inv_beta_reg(200.0, 2.0, 1e-40)); // returns in microseconds
    println!("{}", inv_beta_reg(200.0, 2.0, 1e-60)); // does not return
}
```

(I killed the second call after ~5 mins)

## Issue 3: the result stops moving, and is not monotone

```rust
use statrs::function::beta::inv_beta_reg;

fn main() {
    for q in [1e-2, 1e-4, 1e-10, 1e-20, 1e-30] {
        println!("inv_beta_reg(0.1, 500, {q:e}) = {:e}", inv_beta_reg(0.1, 500.0, q));
    }
}
```

Reference from `scipy.stats.beta(0.1, 500).ppf(q)`, cross-checked with mpmath:

| `q` | statrs | true |
| --- | --- | --- |
| `1e-2` | `2.5310279925347366e-18` | `1.2157036049543745e-23` |
| `1e-4` | `5.297481902267595e-17` | `1.2157036049543762e-43` |
| `1e-10` | `5.164813474103027e-17` | `1.215703604954385e-103` |
| `1e-20` | `5.164813342904473e-17` | `1.2157036049543978e-203` |
| `1e-30` | `5.164813342904473e-17` | `1.2157036049544177e-303` |

From `q = 1e-10` down, every call returns `5.1648133429e-17`. The results are also not monotone in `q`: `1e-2` gives `2.53e-18`, which is smaller than the `5.30e-17` returned for `1e-4`, while the quantile function is increasing. Anything that brackets or bisects on the inverse CDF relies on that ordering.

The same floor shows up on better-behaved shapes, less dramatically:

| `q` | `inv_beta_reg(2, 200, q)` | true | rel. err. |
| --- | --- | --- | --- |
| `1e-20` | `7.053459200759641e-13` | `7.053456158916007e-13` | `4.3e-7` |
| `1e-30` | `7.180186771679575e-16` | `7.053456158585999e-18` | `1e2` |
| `1e-40` | `7.179955800324264e-16` | `7.053456158585986e-23` | `1e7` |

---

FYI I would be interested in trying opening a PR for one or more of these issues

### LINKED ISSUE #440: Special functions and `libm` compatibility

day01:
## Summary

- use exact symmetry and Temme's uniform expansion for large-parameter beta CDFs
- replace the inverse-beta floor and unchecked value-domain Newton step with safeguarded log-domain and shape-two solvers
- certify adjacent and subnormal rounding while preserving monotonicity
- split the beta implementation into focused production modules of at most 250 lines and add regression benchmarks

## Root cause

The forward path formed ill-conditioned gamma and power differences in `f64`; cancellation eventually drove a valid CDF outside `[0, 1]`. The inverse path clamped its initial estimate away from zero and iterated on an underflowing value-domain CDF, causing the fixed `~1e-17` floor, non-monotonic results, non-termination, and an unchecked error panic.

The replacement uses [DLMF 8.17](https://dlmf.nist.gov/8.17), [DLMF 8.18.9–12](https://dlmf.nist.gov/8.18), [TOMS Algorithm 708](https://doi.org/10.1145/131766.131776), and Temme, *Special Functions* (1996), section 11.3.3.2. Boost-derived kernels are attributed and distributed under BSL-1.0 in `THIRD_PARTY_NOTICES.md`.

## TDD and independent numerical references

The issue reproducers fail on upstream and pass with this change. Inputs are passed to every reference as their exact binary64 values. The columns below are signed ULP error against 500-digit MPFR/Boost.Multiprecision references; the endpoint uses the exact `I_x(2,b)` identity at 1100 digits. `0` is correctly rounded.

Cases: center `I_0.5(1e8, 1e8)`; asymmetric `I_x(1e8, 2e8)` at `x=0.33330611678068106`; inverse cases `(a,b,p)=(200,2,1e-60)`, `(200,2,1e-165)`, `(0.1,500,1e-30)`, and the endpoint `(2,1e308,0.5)`.

| Implementation | Center | Asymmetric | `p=1e-60` | `p=1e-165` | Tiny shape | Endpoint | Outcome |
|---|---:|---:|---:|---:|---:|---:|---|
| upstream statrs | outside `[0,1]` | `+2,922,987,503` | timeout | panic | `+4,288,298,043,698,456,868` | timeout | bugs reproduced |
| statrs (this PR) | `0` | `-1` | `0` | `0` | `0` | `0` | correct, 0–1 ULP |
| Boost.Math 1.90 | `+15` | `+15,949,984` | `+12` | `0` | `+3` | returns `0` | less accurate on five cases |
| SciPy 1.18.0 | `+15` | `+15,949,984` | `+12` | `0` | `+3` | returns `0` | same Boost-backed results |
| R 4.2.1 | `0` | `+891` | `0` | `0` | `-234` | timeout | mixed |
| mpmath 1.4.1, 500 digits | timeout | timeout | `0` | `0` | `0` | generic inverse failed | independent cross-check |
| MPFR/Boost.Multiprecision, 500/1100 digits | reference | reference | reference | reference | reference | reference | rounding oracle |

SciPy's beta kernels are Boost-backed, so SciPy and Boost are not independent references. mpmath is independent; MPFR/Boost.Multiprecision is used only as the high-precision arithmetic oracle, not as the Boost.Math implementation under test. A separate 1763-case endpoint audit around rounding thresholds found no error above 1 ULP; a theoretically unresolved adjacent-float overlap selects the monotone ties-to-even result without panicking.

## Performance

| Case | Upstream | This change | Difference |
|---|---:|---:|---:|
| #434 center | 1.909 us, outside `[0,1]` | 3.675 ns, 0 ULP | 99.8% faster |
| #434 large asymmetric CDF | 1.899 us, +2,922,987,503 ULP | 485.581 ns, -1 ULP | 74.4% faster |
| #435 non-termination input | timeout | 363.448 ns, 0 ULP | terminates |
| #435 panic input | panic | 372.712 ns, 0 ULP | finite result |
| #435 tiny-shape input | 5.811 us, wrong/non-monotone | 628.094 ns, 0 ULP | 89.2% faster |
| #435 large-`b` endpoint | timeout | 121.324 ns, 0 ULP | terminates |
| #435 exact midpoint fallback | panic | 14.727 us, 0 ULP | finite, correctly rounded |
| typical finite CDF | 153.811 ns, -46 ULP | 223.018 ns, +1 ULP | 45.0% slower |
| typical finite inverse | 507.253 ns, +10 ULP | 602.560 ns, 0 ULP | 18.8% slower |

| Implementation | Execution mode | #434 asymmetric CDF | #435 `p=1e-60` | #435 tiny shape | #435 endpoint |
|---|---|---:|---:|---:|---:|
| statrs (this PR) | Rust release, `target-cpu=native` | 485.581 ns, -1 ULP | 363.448 ns, 0 ULP | 628.094 ns, 0 ULP | 121.324 ns, 0 ULP |
| Boost.Math 1.90 | C++17, `-O3 -march=native` | 1.966 us, +15,949,984 ULP | 321.625 ns, +12 ULP | 205.119 ns, +3 ULP | 411.673 ns, returns `0` |
| R 4.2.1 | scalar R call | 920.000 ns, +891 ULP | 1.573 us, 0 ULP | 1.120 us, -234 ULP | timeout |
| mpmath 1.4.1, 500 digits | scalar Python reference | timeout | 4.095 ms, 0 ULP | 21.022 ms, 0 ULP | generic inverse failed |
| SciPy 1.18.0 | scalar Python call | 2.692 us, +15,949,984 ULP | 1.099 us, +12 ULP | 985.318 ns, +3 ULP | 1.216 us, returns `0` |

Timings are scalar-call latency; a faster time does not imply a correct result—the accuracy table applies. mpmath is an accuracy oracle rather than a production-speed competitor. Batched SciPy/R throughput is measured separately and is not presented as scalar latency.

## Related issues

Fixes #434 — [`Beta::cdf` loses accuracy above shapes of ~1e4 and returns values outside [0, 1] by 1e8](https://github.com/statrs-dev/statrs/issues/434).

Fixes #435 — [`inv_beta_reg` panics, does not return, and is non-monotone for small `x`](https://github.com/statrs-dev/statrs/issues/435).

## Summary by CodeRabbit

- **New Features**
  - Added beta, incomplete beta, regularized beta, inverse beta, and logarithmic beta functions.
  - Added checked variants that report invalid inputs and convergence failures.
  - Improved numerical handling for extreme parameters, tiny probabilities, and distribution tails.

- **Bug Fixes**
  - Improved beta-distribution survival-function accuracy and inverse-CDF behavior in difficult edge cases.

- **Documentation**
  - Updated licensing information to MIT and Boost Software License 1.0 and added third-party notices.

- **Tests**
  - Expanded high-precision regression coverage and added beta-function benchmarks.


day01:
@YeungOnion i take a boost impl and few articles impl. 
Comparison is correct, please take a look especially on licensing things.

YeungOnion:
One thing I'd like us to consider before adding the Boost implementation is what this means for downstream consumers.

My understanding is that BSL-1.0 is permissive and doesn't have copyleft, so ordinary users of the crate likely wouldn't see much practical impact. 

But it does mean we'd be introducing another piece of licensing/provenance that downstream source redistributors and organizations doing license compliance may need to account for.
I'm initially opposed to adding complexity for crate consumers around licensing. In particular, it would be useful to know what we could do atop public domain or other MIT licensed work, like the Temme paper or Cody's erf or we can figure out exactly what we're taking from boost to see if we can find it elsewhere.

I don't think it's feasible to implement this much code at once without copying Boost, so I think we'll need to do this more piecemeal.

There's also another approach: place this burden elsewhere, can we develop `special` instead and use it as a dep?

day01:
extracted https://github.com/statrs-dev/statrs/pull/456

YeungOnion:
Closing for #456

PR #448, opened 2026-08-16
fix: compute FisherSnedecor pdf in log space for large freedoms

teddytennant:
`FisherSnedecor::pdf` evaluates `((d1 x)^d1 * d2^d2) / (d1 x + d2)^(d1+d2)` with `powf`. At ordinary in-range parameters those powers overflow independently, so `pdf` comes back 0 or NaN and `ln_pdf` (which is just `pdf(x).ln()`) cannot recover.

    FisherSnedecor::new(80.0, 80.0).unwrap().pdf(1.0)   // was 0.0
    FisherSnedecor::new(100.0, 100.0).unwrap().pdf(1.0) // was NaN

F(80, 80) at the mode `x = 1` is a textbook density of about 1.78, not a tail.

Compute `ln_pdf` from the log of the documented formula (`ln_beta` already exists) and take `pdf = ln_pdf.exp()`. Same class of fix as the Hypergeometric / Multinomial pmfs.

A few existing small-dof fixtures differ by about 1 ULP from the old product-then-sqrt path, so those tolerances moved one digit (for example `1e-16` to `1e-15`). That is not how the overflow cases are tested; those have their own expected values.

`cargo test --lib` is green (792 passed).

### DIFF
--- a/src/distribution/fisher_snedecor.rs
+++ b/src/distribution/fisher_snedecor.rs
@@ -380,14 +380,7 @@ impl Continuous<f64, f64> for FisherSnedecor {
     /// where `d1` is the first degree of freedom, `d2` is
     /// the second degree of freedom, and `β` is the beta function
     fn pdf(&self, x: f64) -> f64 {
-        if x.is_infinite() || x <= 0.0 {
-            0.0
-        } else {
-            ((self.freedom_1 * x).powf(self.freedom_1) * self.freedom_2.powf(self.freedom_2)
-                / (self.freedom_1 * x + self.freedom_2).powf(self.freedom_1 + self.freedom_2))
-            .sqrt()
-                / (x * beta::beta(self.freedom_1 / 2.0, self.freedom_2 / 2.0))
-        }
+        self.ln_pdf(x).exp()
     }
 
     /// Calculates the log probability density function for the fisher-snedecor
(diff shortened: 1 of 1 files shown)


day01:
The overflow fix looks right, and `pdf = ln_pdf.exp()` matches the Hypergeometric / Multinomial pattern.

One request in the spirit of #447: the new fixtures match SciPy bit-exactly, but SciPy's `f.pdf` is the same f64 log-space grouping (`betaln` / log-sum), so that is a consistency check, not an independent oracle. High-precision mpmath (80 decimal digits, correctly rounded to binary64) and R land on slightly different last bits:

| Implementation | `F(80,80).pdf(1)` | ULP vs MPFR | `F(100,100).pdf(1)` | ULP vs MPFR |
|---|---|---:|---|---:|
| upstream statrs | `0.0` | — | `NaN` | — |
| this PR / SciPy 1.17.0 | `1.7785575754781575` | +59 | `1.9897309346795113` | +191 |
| R 4.2.1 `df(1, d, d)` | `1.7785575754781444` | 0 | `1.9897309346794692` | ~1 |
| mpmath 80d → binary64 | `1.7785575754781444` | 0 | `1.9897309346794689` | 0 |

can we make it better direct with R or mpmath to keep it more accurate? Of course decision to owners, @YeungOnion

teddytennant:
Fair point, swapped the four fixtures to mpmath at 80 digits rounded to binary64. Suite is still green at the existing 1e-12.

One correction to the table though: the branch does not land on the SciPy value. It gives `1.7785575754780059` for `F(80,80).pdf(1)` and `1.9897309346793983` for `F(100,100).pdf(1)`, so it is roughly 625 and 159 ULP off MPFR, not 59 and 191. The fixtures came from SciPy, the arithmetic did not, which is exactly the reason you asked for a real oracle.

I tried to close the gap and could not. Nearly all of it is `ln_beta`: `ln_beta(40.0, 40.0)` returns `-56.027577129732805` against `-56.027577129732904`, so about 1e-13 absolute, and since the answer is O(1) in log space that comes out as several hundred ULP after `exp`. I also tried regrouping the first term as `0.5 * (d1 * ln(u) + d2 * ln_1p(-u))` with `u = d1*x / (d1*x + d2)`, which kills the cancellation between the three logs. It made things very slightly worse (`1.778557575477968`), because that cancellation was never the problem. So getting near R means a more accurate `ln_beta`, which is its own PR rather than something to bolt onto an overflow fix.

The red `no-std` job is a rustup download that got connection-reset, nothing in the diff.

YeungOnion:
Definitely better than float inf intermediate values.

d_1=d_2, x=1 has a cancellation that probably misses pretty badly. 

```rs
    /// 0.5 * (d1 * ln(d1 * x) + d2 * ln(d2) - (d1 + d2) * ln(d1 * x + d2)) - beta_teem
   /// = 0.5 * (d1 * ln(d1) + d1 * ln(d1) - (d1 + d1) * ln(d1 + d1)) - beta_term
   /// algebraically= -d1 ln(2) - beta_term
```
where the first term would be around 0.5(350+350-812) so there's probably 2 or 3 ULP from nearby subtraction. Exact 1's will do a lot better with a regime parameter like z=(d1 x / d2) 

```
function log_fisher_snedecor_pdf(d1, d2, x):
    a = 0.5 * d1
    b = 0.5 * d2
    z = (d1 * x) / d2
    log_beta = math::lbeta(a,b)

    if z > 2.0: // kinda arbitrary, based on eval of ln_1p(0.5)
        inv_z = 1.0 / z
        term_z     = -(a + b) * math::log1p(inv_z)
        term_x     = -math::log(x)
        term_ratio = -b * math::log(z)

    else:
        term_z = -(a + b) * math::log1p(z)
        term_x = (a - 1.0) * math::log(x) // scales with (d1-1)|1-x|
        term_ratio = a * math::log(d1 / d2)

    return term_z + term_x + term_ratio - log_beta
```

Would have to test this to see if actually improves outside d1=d2, x=1

Would accept this as overflow fix without the extra prec, so up to you @teddytenant, thanks for the fix!

YeungOnion:
taking this thumbs up as good to merge now with more precision later.

PR #450, opened 2026-08-22
Fix/exact boundaries only

YeungOnion:
the work in #410 fixes a few things, one being that ULPs comparison wasn't being used, I must have misunderstood the way `approx` did things when I wrote those macros.

In a sense, this removes the work from #410 by making the macros only use ULPs comparison, and then it goes in an does some removal of special cases.

The other thing that is addresses is taking a wide range of input and "snapping" it to exact values when they're close. This change emphasizes not doing that and to not treat cases as special when they work within another regime's branch, perhaps because that branch will handle it as special, but it shoves the potentially duplicated complexity of checking edge cases upstream.

### DIFF
--- a/src/function/gamma.rs
+++ b/src/function/gamma.rs
@@ -2,7 +2,6 @@
 //! related functions
 
 use crate::consts;
-use crate::prec;
 use core::f64::consts as f64_consts;
 #[cfg(not(feature = "std"))]
 use num_traits::Float as _;
@@ -294,13 +293,6 @@ pub fn checked_gamma_lr(a: f64, x: f64) -> Result<f64, GammaFuncError> {
     let big = 4503599627370496.0;
     let big_inv = 2.22044604925031308085e-16;
 
-    if prec::ulps_eq!(a, 0.0, epsilon = 0.0) {
-        return Ok(1.0);
-    }
-    if prec::ulps_eq!(x, 0.0, epsilon = 0.0) {
-        return Ok(0.0);
-    }
-
     let ax = a * x.ln() - x - ln_gamma(a);
     if ax < -709.78271289338399 {
         if a < x {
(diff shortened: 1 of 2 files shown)


YeungOnion:
@day01 I'd appreciate feedback if you can, this is sort of in response to what was surfaced and resolved in #410 but takes a different stance.

day01:
@YeungOnion can you add me somehow as reviewer, I will find it quicker next time :D

src/function/beta.rs
```diff
@@ -153,14 +148,10 @@ pub fn checked_beta_reg(a: f64, b: f64, x: f64) -> Result<f64, BetaFuncError> {
         return Err(BetaFuncError::XOutOfRange);
     }
 
-    let bt = if x == 0.0 || crate::prec::ulps_eq!(x, 1.0, epsilon = MODULE_EPS) {
```
day01, line 156: Removing this code causes `checked_beta_reg` to return `NaN` for valid large parameters because the logarithmic expression develops non-finite intermediate values.

compared with mpmath at `mp.dps = 80`:

| `a` | `b` | `x` | PR result | mpmath |
|---:|---:|---:|---:|---:|
| `1e200` | `1e200` | `0.0` | `NaN` | `0.0` |
| `1e200` | `1e200` | `1.0` | `NaN` | `1.0` |
| `1e308` | `1.0` | `0.0` | `NaN` | `0.0` |
| `1e308` | `1.0` | `1.0` | `NaN` | `1.0` |

YeungOnion:
So the failing behavior is shown on the edges, but the test cases for (a,b) would similarly fail for any x on (0,1), I'm expecting #456 will fix this based on special treatment for large a,b

I will drop the work on beta from this PR.

Issue #453, opened 2026-08-24
pdf loses the tail in several distributions that exponentiate before dividing


teddytennant:
While fixing #452 I checked whether other distributions share the shape that caused it. Several do.

The pattern is computing `pdf` as `exp(...) / normalizer` in value space. The numerator underflows or overflows before the density itself does, and dividing afterwards cannot recover it. Measured on `main` at 128c9ae:

```
Normal(0,1e-300).pdf(5e-299)        = 0        true = 5.40515e-244
Laplace(0,1e-300).pdf(1e-297)       = 0        true = 2.53798e-135
Exp(1e300).pdf(1e-297)              = 0        true = 5.07596e-135
Chi(160).pdf(45.0)                  = 0        true = 2.5387e-318
Levy(0,2e-297).pdf(1e-300)          = NaN
Gamma(150,1e10).pdf(1e-8)           = NaN
InverseGamma(150,1e-10).pdf(1e-12)  = NaN
Pareto(1e-200,5).pdf(1e-100)        = NaN
Gumbel(0,1).pdf(-1000.0)            = NaN
```

The three `true` values above are exact arithmetic at 120 digits. A density is never NaN, so those five rows are wrong whatever the reference value is.

These are not all one fix:

- `Laplace` and `InverseGamma` define `ln_pdf` as `self.pdf(x).ln()`, so there is no correct log-space form to delegate to. Both need one written, the same way #448 did for FisherSnedecor.
- `Chi` and `Gamma` already branch to `ln_pdf(x).exp()` above a shape threshold (160 and 160.0). The threshold guards the shape parameter, but the overflow here comes from `rate.powf(shape)` and `x.powf(freedom - 1.0)`, which blow up independently of it. Those thresholds want re-tuning rather than a new branch.
- `Gumbel` fails through a double exponential overflowing to NaN, which is a cousin of this rather than the same thing.
- `Normal` is probably the one worth doing first. It is the most used distribution here, its `pdf_unchecked` is the same expression as the LogNormal one, its `ln_pdf_unchecked` is already correct, and `StudentsT::pdf` delegates to it for `freedom >= 1e8`.

I did not roll any of these into #454. That would turn a one-file fix into a rewrite of a third of the distribution module. Happy to send them as separate PRs if you want them, and happy to be told some are not worth changing.

PR #454, opened 2026-08-24
fix: keep the left tail of LogNormal::pdf

### LINKED ISSUE #452: bug: `LogNormal::pdf` in the left tail: 0.0 where the density is finite, then values up to 2x off
In the left tail, `LogNormal::pdf(x)` and `ln_pdf(x).exp()` on the same object disagree over two adjacent bands of `x`:

1. a band where `pdf` returns `0.0` and `ln_pdf(x).exp()` returns an ordinary `f64`,
2. immediately above it, a band where `pdf` returns a finite value that differs from the density by up to a factor of two.

Both bands sit well above the point where the density underflows, both widen with `scale`, and both move with `location`. `scipy.stats.lognorm` and mpmath agree with `ln_pdf(x).exp()` throughout.

**Environment:** statrs main on `128c9ae6d52c2be4574b6d8669136e344b87f65c`, rustc 1.97.1, aarch64-apple-darwin.

## Reproducer

```rust
use statrs::distribution::{Continuous, LogNormal};

fn main() {
    // 1. pdf returns 0.0 where the density is an ordinary double
    let d = LogNormal::new(0.0, 15.0).unwrap();
    println!("{:e}\t{:e}", d.pdf(1e-252), d.ln_pdf(1e-252).exp());

    // 2. just above that band, pdf returns a finite value about 1.9x the density
    let d = LogNormal::new(0.0, 10.0).unwrap();
    println!("{:e}\t{:e}", d.pdf(2.24e-168), d.ln_pdf(2.24e-168).exp());

    // 3. the case this came out of
    let d = LogNormal::new(0.0, 5.0).unwrap();
    println!("{:e}\t{:e}", d.pdf(1.38e-87), d.ln_pdf(1.38e-87).exp());
}
```

```
0e0                     3.047968563730997e-75
8.79927122410328e-158   4.605967253151374e-158
0e0                     2.0733810257862521e-262
```

## Expected

| case | `pdf` | `scipy.stats.lognorm(s=sigma).pdf` | mpmath, `mp.dps = 60` |
| --- | --- | --- | --- |
| `LogNormal(0, 15)`, `x = 1e-252` | `0e0` | `3.047969e-75` | `3.047969e-75` |
| `LogNormal(0, 10)`, `x = 2.24e-168` | `8.799271e-158` | `4.605967e-158` | `4.605967e-158` |
| `LogNormal(0, 5)`, `x = 1.38e-87` | `0e0` | `2.073381e-262` | `2.073381e-262` |

---

As those above seems like "lucky" findings, I asked Claude to run a full sweep. What follows is generated by it.

<details><summary>Claude Opus 5.0 analysis</summary>
<p>

## How far the two bands reach

`LogNormal(0, sigma)`, scanning `x` at 200 points per power of 10. "Width" is the span of band 1 in powers of 10: `1e-96` to `1e-84` is 12 of them.

| `sigma` | band 1: `pdf` returns `0.0` | width | largest density it returns `0.0` for | band 2: worst relative error |
| --- | --- | --- | --- | --- |
| `1` | `6.4e-18` to `1.7e-17` | 0.4 | `3.886588e-308` | `0.91` at `x = 1.72e-17` |
| `2` | `5.3e-36` to `2.9e-34` | 1.7 | `1.412756e-291` | `0.91` at `x = 2.95e-34` |
| `3` | `5.2e-55` to `5.0e-51` | 4.0 | `5.904035e-275` | `0.91` at `x = 5.07e-51` |
| `5` | `5.8e-96` to `1.5e-84` | 11.4 | `1.273613e-241` | `0.91` at `x = 1.50e-84` |
| `7` | `4.7e-141` to `4.4e-118` | 23.0 | `3.161836e-208` | `0.91` at `x = 4.42e-118` |
| `10` | `5.4e-217` to `2.2e-168` | 48.6 | `4.449485e-158` | `0.91` at `x = 2.24e-168` |
| `15` | `5e-324` to `3.3e-252` | 71.8 or more | `1.975989e-74` | `0.97` at `x = 3.31e-252` |
| `19` | `5e-324` to `2.9e-319` | 4.8 or more | `1.813587e-07` | `0.96` at `x = 2.88e-319` |

Reading the `sigma = 10` row: `pdf` returns `0.0` over 48.6 powers of 10, the largest density it calls zero is `4.4e-158`, and the first `x` above that band, `2.24e-168`, is the second line of the reproducer, where `pdf` returns `8.80e-158` against a true `4.61e-158`.

Four notes on that scan:

- Band 2 is wide too. Measured out to where pdf comes back within 1e-9 relative, it spans 0.2 powers of 10 at `sigma = 1`, 1.1 at `sigma = 5`, 2.3 at `sigma = 10` and 3.4 at `sigma = 15`, starting at the top of band 1 and easing back to ordinary `1e-14` agreement above it.
- The `sigma = 15` and `sigma = 19` bands run off the bottom of the `f64` range, so those two widths are floors rather than measurements.
- Below band 1 the density really does underflow (`LogNormal(0, 5).pdf(1e-100)` is about `2.4e-362`), so `0.0` is the right answer there.
- Stepping `sigma` by `0.01`, the largest one with a band is `19.28`, and there is none at `19.5` or above. Every `sigma` I sampled from `0.05` upward has one, but below `sigma` around `0.6` the affected densities are all under `1e-313`, so that end looks immaterial in practice.

## `location` moves both bands

| `(location, scale)` | band 1: `pdf` returns `0.0` | width | largest density it returns `0.0` for |
| --- | --- | --- | --- |
| `(0, 5)` | `5.8e-96` to `1.5e-84` | 11.4 | `1.273613e-241` |
| `(-100, 5)` | `8.2e-145` to `5.5e-128` | 16.8 | `3.394508e-198` |
| `(100, 5)` | `9.4e-47` to `4.0e-41` | 5.6 | `4.778570e-285` |

Band 2 moves with it, with worst relative errors of `0.93` and `0.90` on the two shifted rows. With `location = 100` the first `x` that returns `0.0` is `4.0e-41`, much closer to ordinary magnitudes than any `location = 0` case.

## Where the same scan finds nothing

`LogNormal(0, 5).pdf(1e84)` and `LogNormal(0, 10).pdf(1e200)` both return `0.0`, and mpmath puts both densities below the `f64` range, so the right tail is fine. `Normal::pdf` for `std_dev` in `{1e-3, 1, 100}`, over the same grid across both tails, agrees with `ln_pdf().exp()` to `4.9e-14` or better everywhere.

## Two more corners of the same tail, with subnormal `x`

Less likely to matter than the two bands, but they came out of the same scan:

| call | returns | mpmath, `mp.dps = 60` |
| --- | --- | --- |
| `LogNormal::new(0.0, 0.1).unwrap().pdf(5e-324)` | `NaN` | density is `0` (log density `-2.77e7`) |
| `LogNormal::new(0.0, 0.1).unwrap().ln_pdf(1e-323)` | `inf` | `-2.765723e7` |
| `LogNormal::new(0.0, 20.0).unwrap().pdf(5e-324)` | `4.736269e20` | `5.668494e20` |

The `NaN` and the `inf` show up for `scale <= 0.5`, and `scipy.stats.lognorm` returns `inf` for those same two calls, so mpmath is the only reference I have there. The third row is 16% off with no band 1 anywhere at that `scale` (same at `scale = 50`: `9.856067e272` against `1.179601e273`), and `ln_pdf().exp()` matches mpmath to `2e-13` on both.

</p>
</details>

teddytennant:
`LogNormal::pdf` computes `exp(-d^2/2) / (x * sigma * sqrt(2 pi))`. Both ends of that quotient leave the representable range long before the density does. The numerator underflows in the tails, and the denominator overflows to infinity or underflows towards zero for extreme `x`. Either one on its own collapses the result.

On main:

```
LogNormal(0, 15).pdf(1e-252)              got 0            expected 3.047968563730586e-75
LogNormal(0, 10).pdf(2.24e-168)           got 8.79927e-158 expected 4.60597e-158
LogNormal(0, 5).pdf(1.38e-87)             got 0            expected 2.073381025786355e-262
LogNormal(ln(f64::MAX), 1).pdf(f64::MAX)  got 0            expected 2.219190097936194e-309
```

`ln_pdf` has a version of the same problem. It forms `(x * self.scale).ln()`, which is `-inf` once the product underflows and `+inf` once it overflows, so it returns `+inf` for `LogNormal(0, 0.25).ln_pdf(5e-324)` and `-inf` for `LogNormal(0, 1e10).ln_pdf(f64::MAX)`.

So `pdf` keeps the value-space expression, guarded now on the numerator and the denominator, and falls back to `ln_pdf(x).exp()` instead of repeating the formula. `ln_pdf` splits `ln(x * sigma)` into `ln(x) + ln(sigma)`, but only where the product is not a normal float. The split is not unconditional because `ln(x * sigma)` rounds once instead of twice and is the more accurate form while the product is in range; unconditional, it moves four existing `test_ln_pdf` rows by a ulp and fails one of them against its 1e-16 tolerance.

Two rows in `test_pdf` move by 1 ULP, because the denominator is grouped as `x * (sqrt(2 pi) * sigma)` now so that a subnormal `x` is rounded once instead of twice. Those two are `test_absolute` with the expected values untouched. Nothing else in the existing suite changed.

Tests cover the left tail, the right tail, both denominator extremes, the boundaries either side of each guard, and `ln_pdf` at both ends. Expected values are exact arithmetic on the binary64 inputs rather than SciPy, following 10cf6d6. `cargo test` is green at 803 passed, as are `cargo fmt -- --check`, clippy under `-Dwarnings`, and `cargo check --no-default-features --lib`.

The same exponentiate-then-divide shape appears in nine other distributions. I filed that separately as #453 rather than widen this.

Fixes #452

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved log-normal probability density calculations for extreme tail values.
  * Prevented valid finite densities from being reported as infinite or lost due to numerical underflow.
  * Improved handling of boundary conditions and invalid numeric inputs, including NaN values.
* **Tests**
  * Added coverage for left-tail, right-tail, underflow, boundary, and NaN scenarios.

### DIFF
--- a/src/distribution/log_normal.rs
+++ b/src/distribution/log_normal.rs
@@ -640,18 +640,15 @@ mod tests {
             (0.0, 5.0, 1.38e-87, 2.07338102578635509e-262),
         ];
 
-        let mut bad = alloc::vec::Vec::new();
         for (location, scale, x, expected) in cases {
             let d = create_ok(location, scale);
             let got = d.pdf(x);
             let rel = ((got - expected) / expected).abs();
-            if !prec::relative_eq!(got, expected, epsilon = 0.0, max_relative = 1e-12) {
-                bad.push(alloc::format!(
-                    "LogNormal({location}, {scale}).pdf({x:e}): got {got:e}, expected {expected:e}, rel err {rel:e}"
-                ));
-            }
+            assert!(
+                prec::relative_eq!(got, expected, epsilon = 0.0, max_relative = 1e-12),
+                "LogNormal({location}, {scale}).pdf({x:e}): got {got:e}, expected {expected:e}, rel err {rel:e}"
+            );
         }
-        assert!(bad.is_empty(), "{}", bad.join("\n"));
     }
 
     #[test]
--- a/src/distribution/log_normal.rs
+++ b/src/distribution/log_normal.rs
@@ -335,20 +335,20 @@ impl Continuous<f64, f64> for LogNormal {
         if x <= 0.0 || x.is_infinite() {
             0.0
         } else {
-            let ln_x = x.ln();
-            let d = (ln_x - self.location) / self.scale;
-            let exponent = -0.5 * d * d;
-            let num = exponent.exp();
-            if num.is_normal() {
-                num / (x * consts::SQRT_2PI * self.scale)
+            let d = (x.ln() - self.location) / self.scale;
+            let numer = (-0.5 * d * d).exp();
+            // grouped so that a subnormal `x` is rounded once instead of twice
+            let denom = x * (consts::SQRT_2PI * self.scale);
+            if numer.is_normal() && denom.is_normal() {
+                numer / denom
             } else {
-                // `num` underflows to zero, or to a subnormal holding only a few
-                // significant bits, long before the density itself does. Divide in
-                // log space instead and exponentiate once. This is `ln_pdf`'s
-                // formula with `ln(x * σ)` split into `ln(x) + ln(σ)`, because
-                // `(x * self.scale).ln()` is `-inf` once `x * σ` underflows, which
-                // would make `pdf` return `inf` where it should return zero.
-                (exponent - consts::LN_SQRT_2PI - ln_x - self.scale.ln()).exp()
+                // Either end of the quotient has left the range where dividing is
+                // accurate: `numer` underflows to zero, or to a subnormal holding
+                // only a few significant bits, long before the density itself does,
+                // and `denom` overflows to infinity or underflows for extreme `x`.
+                // Both collapse the quotient to zero where the density is still a
+                // representable f64, so divide in log space and exponentiate once.
+                self.ln_pdf(x).exp()
             }
         }
     }
@@ -367,8 +367,20 @@ impl Continuous<f64, f64> for LogNormal {
         if x <= 0.0 || x.is_infinite() {
             f64::NEG_INFINITY
         } else {
-            let d = (x.ln() - self.location) / self.scale;
-            (-0.5 * d * d) - consts::LN_SQRT_2PI - (x * self.scale).ln()
+            let ln_x = x.ln();
+            let d = (ln_x - self.location) / self.scale;
+            let x_scale = x * self.scale;
+            // `x * σ` overflows to infinity or underflows to zero at the extremes
+            // of the support, and `ln` then hands back `±inf`, so `ln(xσ)` is split
+            // into `ln(x) + ln(σ)` there. Everywhere else the product keeps its
+            // full precision and is the more accurate of the two, because it only
+            // rounds once and `ln(x)` and `ln(σ)` can be much larger than the sum.
+            let ln_x_scale = if x_scale.is_normal() {
+                x_scale.ln()
+            } else {
+                ln_x + self.scale.ln()
+            };
+            (-0.5 * d * d) - consts::LN_SQRT_2PI - ln_x_scale
         }
     }
 }
@@ -593,7 +605,7 @@ mod tests {
         test_absolute(-0.1, 1.5, 0.90492497850024368541682348133921492204585092983646, 1e-15, pdf(0.1));
         test_absolute(-0.1, 1.5, 0.49191985207660942803818797602364034466489243416574, 1e-16, pdf(0.5));
         test_exact(-0.1, 1.5, 0.33133347214343229148978298237579567194870525187207, pdf(0.8));
-        test_exact(-0.1, 2.5, 1.0824698632626565182080576574958317806389057196768, pdf(0.1));
+        test_absolute(-0.1, 2.5, 1.0824698632626565182080576574958317806389057196768, 1e-15, pdf(0.1));
         test_absolute(-0.1, 2.5, 0.31029619474753883558901295436486123689563749784867, 1e-16, pdf(0.5));
         test_absolute(-0.1, 2.5, 0.19922929916156673799861939824205622734205083805245, 1e-16, pdf(0.8));
 
@@ -608,7 +620,7 @@ mod tests {
         test_absolute(1.5, 1.5, 0.17185785323404088913982425377565512294017306418953, 1e-16, pdf(0.8));
         test_absolute(1.5, 2.5, 0.50186885259059181992025035649158160252576845315332, 1e-15, pdf(0.1));
         test_absolute(1.5, 2.5, 0.21721369314437986034957451699565540205404697589349, 1e-16, pdf(0.5));
-        test_exact(1.5, 2.5, 0.15729636000661278918949298391170443742675565300598, pdf(0.8));
+        test_absolute(1.5, 2.5, 0.15729636000661278918949298391170443742675565300598, 1e-16, pdf(0.8));
         test_exact(2.5, 0.1, 5.6836826548848916385760779034504046896805825555997e-500, pdf(0.1));
         test_absolute(2.5, 0.1, 3.1225608678589488061206338085285607881363155340377e-221, 1e-233, pdf(0.5));
         test_absolute(2.5, 0.1, 4.6994713794671660918554320071312374073172560048297e-161, 1e-173, pdf(0.8));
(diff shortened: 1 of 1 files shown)


teddytennant:
codecov flagged the failure-reporting arm in `test_pdf_left_tail`, which never runs on a green build. Folded it into a direct `assert!` so the message is on the assert itself. Same cases, same 1e-12 bound, same text on failure.

src/distribution/log_normal.rs
```diff
@@ -335,8 +335,21 @@ impl Continuous<f64, f64> for LogNormal {
...
+            if num.is_normal() {
+                num / (x * consts::SQRT_2PI * self.scale)
+            } else {
+                // `num` underflows to zero, or to a subnormal holding only a few
+                // significant bits, long before the density itself does. Divide in
+                // log space instead and exponentiate once. This is `ln_pdf`'s
+                // formula with `ln(x * σ)` split into `ln(x) + ln(σ)`, because
+                // `(x * self.scale).ln()` is `-inf` once `x * σ` underflows, which
+                // would make `pdf` return `inf` where it should return zero.
+                (exponent - consts::LN_SQRT_2PI - ln_x - self.scale.ln()).exp()
```
FBruzzesi: Hi @teddytennant thanks for taking care of the issue. I hope its ok for me to chime in.

Wouldn't it make sense to move this bit into `ln_pdf` and conditionally call that one in here?

**Mandatory disclaimer**: Not a rust nor a numerical precision expert by any mean

I was going to propose the following change (+ a series of tests), but you beat me to it:

```diff
diff --git a/src/distribution/log_normal.rs b/src/distribution/log_normal.rs
index d94c06a..10379c4 100644
--- a/src/distribution/log_normal.rs
+++ b/src/distribution/log_normal.rs
@@ -336,7 +336,18 @@ impl Continuous<f64, f64> for LogNormal {
             0.0
         } else {
             let d = (x.ln() - self.location) / self.scale;
-            (-0.5 * d * d).exp() / (x * consts::SQRT_2PI * self.scale)
+            let numer = (-0.5 * d * d).exp();
+            // grouped so that a subnormal `x` is rounded once instead of twice
+            let denom = x * (consts::SQRT_2PI * self.scale);
+            // Both factors leave the normal range long before the density does,
+            // making the quotient 0, NaN, or (through gradual underflow) off by up to
+            // a factor of two; in log space only the density need be representable.
+            // See #452
+            if numer >= f64::MIN_POSITIVE && (f64::MIN_POSITIVE..f64::INFINITY).contains(&denom) {
+                numer / denom
+            } else {
+                self.ln_pdf(x).exp()
+            }
         }
     }
 
@@ -354,8 +365,11 @@ impl Continuous<f64, f64> for LogNormal {
         if x <= 0.0 || x.is_infinite() {
             f64::NEG_INFINITY
         } else {
-            let d = (x.ln() - self.location) / self.scale;
-            (-0.5 * d * d) - consts::LN_SQRT_2PI - (x * self.scale).ln()
+            let ln_x = x.ln();
+            let d = (ln_x - self.location) / self.scale;
+            // `ln(x) + ln(σ)` rather than `ln(xσ)`, whose product under- or overflows
+            // for subnormal `x` or extreme `σ`, making the log density `±∞`. See #452
+            (-0.5 * d * d) - consts::LN_SQRT_2PI - ln_x - self.scale.ln()
         }
     }
 }
```

teddytennant: Done, with one change. The split has to be conditional. Done unconditionally it moves four existing `test_ln_pdf` rows by a ulp, and one of them (location -0.1, scale 2.5, `ln_pdf(0.1)`) goes from 1.4e-17 to 2.4e-16 of error against a 1e-16 tolerance, so it fails. `ln(x * sigma)` rounds once instead of twice, so I kept it while the product is in range and split only when the product overflows or underflows.

Your denominator regrouping is in as written. I spelled the guard as `numer.is_normal() && denom.is_normal()`, which is the same set of values, just to match the line it replaces. It costs a ulp on two `test_pdf` rows that were `test_exact`, so those two are `test_absolute` now with the expected values untouched.

While I was in there, `ln_pdf` had the same bug on the overflow side: `LogNormal(0, 1e10).ln_pdf(f64::MAX)` was -inf. Same branch covers it.

YeungOnion:
Looks good, thanks for this fix!

PR #458, opened 2026-08-26
refactor(distribution): move inverse_cdf panic checks to try_inverse_cdf

YeungOnion:
## Summary
- `inverse_cdf` implementations across the legacy `ContinuousCDF`/`DiscreteCDF` distributions duplicated the `[0, 1]` domain check that `try_inverse_cdf` also needs, panicking on invalid input instead of just computing. Strip those checks out of `inverse_cdf` (now a thin, unchecked path per distribution) and let `try_inverse_cdf` own validation before delegating to it — either via a per-distribution override where the valid domain isn't the generic closed `[0, 1]` (e.g. `Categorical`'s open `(0, 1)`), or via the `ContinuousCDF`/`DiscreteCDF` trait defaults otherwise.
- `Beta` had a fully duplicated `try_inverse_cdf` override (identical check + call, copy-pasted from `inverse_cdf`); removed in favor of the trait default.
- `Erlang`/`ChiSquared` forward `inverse_cdf` to an inner `Gamma`; give them the same treatment for `try_inverse_cdf` so they track `Gamma::try_inverse_cdf` instead of re-deriving the `[0, 1]` check themselves and calling `Gamma::inverse_cdf` directly.
- `DiscreteCDF` gained a `try_inverse_cdf` default method (didn't have one before), mirroring `ContinuousCDF`.
- `Geometric`'s two panics for numeric overflow (result exceeds `u64::MAX`, or an intermediate value overflows `f64`) are intentionally left in `inverse_cdf` — those aren't input-domain checks, they're deferred as a separate design question.

## Test plan
- [x] `cargo test --all-features` — 796 unit tests + 195 doctests pass
- [x] `cargo clippy --all-features --all-targets` — clean
- [x] Updated tests that asserted `#[should_panic]` on `inverse_cdf` to instead assert `try_inverse_cdf` returns `Err(InverseCdfError::ArgumentOutOfRange)`

## Summary by CodeRabbit

- **New Features**
  - Added fallible inverse-CDF support for additional discrete distributions.
  - Added dedicated errors for invalid probabilities and non-representable results.
  - Added clearer NaN and range handling for inverse-CDF calculations.

- **Behavior Changes**
  - Direct inverse-CDF methods generally no longer panic for out-of-range probabilities.
  - Invalid inputs can be handled through fallible APIs where available.
  - Out-of-range inputs now return distribution-specific results, including `NaN` or saturated endpoints, instead of being rejected upfront.

### DIFF
--- a/src/distribution/categorical.rs
+++ b/src/distribution/categorical.rs
@@ -1,4 +1,4 @@
-use crate::distribution::{Discrete, DiscreteCDF, InverseCdfError};
+use crate::distribution::{Discrete, DiscreteCDF, DiscreteInverseCdfError};
 use crate::statistics::*;
 use alloc::vec::Vec;
 #[cfg(not(feature = "std"))]
@@ -196,12 +196,16 @@ impl DiscreteCDF<u64, f64> for Categorical {
 
     /// Calculates the inverse cumulative distribution function for the
     /// categorical distribution at `x`, returning an error instead of
-    /// panicking if `x <= 0.0` or `x >= 1.0`.
-    fn try_inverse_cdf(&self, x: f64) -> Result<u64, InverseCdfError> {
-        if x >= 1.0 || x <= 0.0 {
-            Err(InverseCdfError::ArgumentOutOfRange)
-        } else {
+    /// panicking if `x` is not in the open interval `(0.0, 1.0)`, including
+    /// when `x` is NaN.
+    fn try_inverse_cdf(&self, x: f64) -> Result<u64, DiscreteInverseCdfError<f64>> {
+        // Written as a conjunction of `>`/`<`, not `x >= 1.0 || x <= 0.0`, so
+        // that NaN (which compares false against everything) is rejected
+        // rather than falling through to `locate`, which panics on NaN.
+        if x > 0.0 && x < 1.0 {
             Ok(self.inverse_cdf(x))
+        } else {
+            Err(DiscreteInverseCdfError::ArgumentOutOfRange)
         }
     }
 }
@@ -524,7 +528,7 @@ mod tests {
         let dist = create_ok(&[4.0, 2.5, 2.5, 1.0]);
         assert_eq!(
             dist.try_inverse_cdf(0.0),
-            Err(InverseCdfError::ArgumentOutOfRange)
+            Err(DiscreteInverseCdfError::ArgumentOutOfRange)
         );
     }
 
@@ -533,7 +537,16 @@ mod tests {
         let dist = create_ok(&[4.0, 2.5, 2.5, 1.0]);
         assert_eq!(
             dist.try_inverse_cdf(1.0),
-            Err(InverseCdfError::ArgumentOutOfRange)
+            Err(DiscreteInverseCdfError::ArgumentOutOfRange)
+        );
+    }
+
+    #[test]
+    fn test_try_inverse_cdf_nan_does_not_panic() {
+        let dist = create_ok(&[4.0, 2.5, 2.5, 1.0]);
+        assert_eq!(
+            dist.try_inverse_cdf(f64::NAN),
+            Err(DiscreteInverseCdfError::ArgumentOutOfRange)
         );
     }
 
--- a/src/distribution/categorical.rs
+++ b/src/distribution/categorical.rs
@@ -1,4 +1,4 @@
-use crate::distribution::{Discrete, DiscreteCDF};
+use crate::distribution::{Discrete, DiscreteCDF, InverseCdfError};
 use crate::statistics::*;
 use alloc::vec::Vec;
 #[cfg(not(feature = "std"))]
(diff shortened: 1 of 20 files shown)


src/distribution/categorical.rs
```diff
@@ -195,12 +191,23 @@ impl DiscreteCDF<u64, f64> for Categorical {
...
-
         self.locate(x)
     }
+
+    /// Calculates the inverse cumulative distribution function for the
+    /// categorical distribution at `x`, returning an error instead of
+    /// panicking if `x` is not in the open interval `(0.0, 1.0)`, including
+    /// when `x` is NaN.
+    fn try_inverse_cdf(&self, x: f64) -> Result<u64, DiscreteInverseCdfError<f64>> {
+        // Written as a conjunction of `>`/`<`, not `x >= 1.0 || x <= 0.0`, so
```
day01, line 202: it means the same in code below, are u sure we need explanation of code which is longer then code?

src/distribution/gamma.rs
```diff
@@ -185,9 +185,6 @@ impl ContinuousCDF<f64, f64> for Gamma {
     }
 
     fn inverse_cdf(&self, p: f64) -> f64 {
```
day01, line 187: fter removing the domain panic, `while cdf(high) < p { high *= 2 }` never finishes for `p > 1` (and `low /= 2` spins at 0 for `p < 0`). ChiSquared/Erlang inherit this.

Checked against scipy 1.17 / mpmath:
- `gamma(a=2).ppf(1)` = +∞ (this method already special-cases `p == 1`)
- `gamma(a=2).ppf(1+eps)` = **NaN**
- `gamma(a=2).ppf(2)` = **NaN**
- `gamma(a=2).ppf(-1e-16)` = **NaN**

The new path hangs on those inputs. `1+eps` is the realistic `inverse_cdf(cdf(x))` rounding case. Unchecked should mean NaN/Inf like scipy, not a hang.

src/distribution/mod.rs
```diff
@@ -265,15 +302,13 @@ pub trait DiscreteCDF<K: Sized + Num + Ord + Clone + NumAssignOps, T: Float>:
     /// Due to issues with rounding and floating-point accuracy the default implementation may be ill-behaved
     /// Specialized inverse cdfs should be used whenever possible.
     ///
-    /// # Panics
-    /// this default impl panics if provided `p` not on interval [0.0, 1.0]
+    /// Does not check that `p` lies on `[0.0, 1.0]`; use [`try_inverse_cdf`](DiscreteCDF::try_inverse_cdf)
+    /// if `p` is not already known to be valid.
     fn inverse_cdf(&self, p: T) -> K {
```
day01, line 307: Same search, discrete default (Poisson, Binomial, …). `cdf` saturates at 1, so `p > 1` doubles `ub` until overflow (debug panic) or wrap-to-0 (release hang).

scipy 1.17 `geom(p=0.5).ppf`:
- `ppf(2)` / `ppf(1+eps)` / `ppf(NaN)` → **NaN**
- `ppf(1)` → +∞

src/distribution/log_normal.rs
```diff
@@ -198,10 +194,8 @@ impl ContinuousCDF<f64, f64> for LogNormal {
             0.0
         } else if p < 1.0 {
             (self.location - (self.scale * f64_consts::SQRT_2 * erf::erfc_inv(2.0 * p))).exp()
-        } else if p == 1.0 {
-            f64::INFINITY
         } else {
-            panic!("p must be within [0.0, 1.0]");
+            f64::INFINITY
```
day01, line 198: This `else` is `p >= 1` **or NaN**. Compared to scipy `lognorm(s=1).ppf` and mpmath (`exp(μ + σ√2 erfinv(2p−1))`):

| p | PR | scipy / mpmath |
|---|---|---|
| 1 | +∞ | +∞ |
| 1+eps | **+∞** | **NaN** |
| 2 | **+∞** | **NaN** |
| NaN | **+∞** | **NaN** |
| (0,1) | matches, max 19 ulp vs mpmath-f64 | |

src/distribution/laplace.rs
```diff
@@ -167,9 +167,6 @@ impl ContinuousCDF<f64, f64> for Laplace {
     ///
     /// where `μ` is the location, `b` is the scale
     fn inverse_cdf(&self, p: f64) -> f64 {
```
day01: **[suggestion]** The removed `p <= 0 || p >= 1` panic also fired on the **legal** endpoints. After this change, `μ + b ln(2p)` / `μ − b ln(2−2p)` vs mpmath dps=60 and scipy `laplace.ppf`:

| p | old | PR | scipy / mpmath |
|---|---|---|---|
| 0 | panic | −∞ | −∞ |
| 1 | panic | +∞ | +∞ |
| 1e-16 | −36.1482 | same | same (0 ulp on (0,1)) |
| 1+eps / 2 / NaN | panic | NaN | NaN (scipy) |

So the endpoint behavior now matches the references — worth a test (`inverse_cdf(0) == -∞`, `inverse_cdf(1) == +∞`) so it doesn’t look accidental.

**Suggestion:** Add those two assertions next to the existing Laplace quantile tests.

day01 (changes requested):
some cases can break regression and hang the app

Issue #352, opened 2025-09-01
Development accerlation by AI


wangjiawen2013:
Hi,
Can we converte existing statistic package implementation in one language (maybe Python/C) to Rust using AI (such as gemini). Here is an example project: https://github.com/cool-japan/scirs
This will be time saver.

YeungOnion:
Sorry I didn't get back to this one sooner, but largely the value of AI development comes from moving faster, and while I wouldn't say that development speed is a non-goal I don't treat this as a goal. It's also clear to me that the community doesn't want it either based on the feedback for this issue.

I won't explicitly bar AI contributed code, but at the moment, I'm the project's only reviewer and I am that bottleneck.

PR #460, opened 2026-08-28
fix: Triangular::pdf returns NaN when mode equals min

jaideeppyne:
`Triangular::new` accepts `mode == min` (a right triangle) — `test_create` already covers it with `create_ok(1.0, 2.0, 1.0)`. But `pdf` evaluates the rising edge as `2 * (x - min) / ((max - min) * (mode - min))`, and at `x == min == mode` both the numerator and the `(mode - min)` factor are zero, so the result is `0 / 0` = NaN. `ln_pdf` delegates to `pdf` and returns NaN as well.

```rust
let d = Triangular::new(0.0, 1.0, 0.0).unwrap();
d.pdf(0.0);     // NaN, expected 2.0
d.ln_pdf(0.0);  // NaN, expected ln(2)
```

The mirrored `mode == max` case is already correct: `pdf` takes the same first branch and returns `2 / (max - min)`, the height of the triangle. This handles the degenerate rising edge so both endpoints agree.

`x == min == mode` is the only `x` that can reach that branch when `mode == min`, so no other input changes value. Added `pdf`/`ln_pdf` cases for both endpoint modes and extended `test_continuous` to cover them; both new assertions fail on `main` with NaN.

Kept deliberately minimal since you mentioned in #352 that review bandwidth is the bottleneck.

Disclosure: this was AI assisted. An agent ran a differential sweep of the distributions against mpmath at 60 digits, which is what surfaced this, and drafted the patch; I verified the result and the reasoning before submitting.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed triangular distribution calculations when the mode equals the minimum value.
  - Prevented invalid results at the distribution’s rising-edge boundary.
  - Improved reliability for probability density and log-density calculations in this edge case.

- **Tests**
  - Added coverage for minimum-mode endpoint behavior.
  - Verified accurate density and log-density results.
  - Expanded validation of continuous distribution behavior at endpoint modes.

### DIFF
--- a/src/distribution/triangular.rs
+++ b/src/distribution/triangular.rs
@@ -398,10 +398,6 @@ impl Continuous<f64, f64> for Triangular {
         let c = self.mode;
         if a <= x && x <= c {
             if c == a {
-                // The only `x` reaching this branch is `x == min == mode`, where the
-                // rising edge is degenerate and the expression below is 0 / 0. The
-                // density there is the height of the triangle, matching the value the
-                // falling edge already returns for the mirrored `mode == max` case.
                 2.0 / (b - a)
             } else {
                 2.0 * (x - a) / ((b - a) * (c - a))
@@ -559,12 +555,15 @@ mod tests {
         test_exact(-5.0, -3.0, -4.0, 0.5, pdf(-4.5));
         test_exact(-5.0, -3.0, -4.0, 1.0, pdf(-4.0));
         test_exact(-5.0, -3.0, -4.0, 0.5, pdf(-3.5));
-        // mode == min: the rising edge is degenerate at x == min, the density there
-        // is the height of the triangle
+    }
+
+    #[test]
+    fn test_pdf_mode_at_an_endpoint() {
+        let pdf = |arg: f64| move |x: Triangular| x.pdf(arg);
         test_exact(0.0, 1.0, 0.0, 2.0, pdf(0.0));
         test_exact(-2.0, 5.0, -2.0, 2.0 / 7.0, pdf(-2.0));
-        // mirrored case, mode == max
         test_exact(0.0, 1.0, 1.0, 2.0, pdf(1.0));
+        test_exact(-2.0, 5.0, 5.0, 2.0 / 7.0, pdf(5.0));
     }
 
     #[test]
@@ -585,8 +584,14 @@ mod tests {
         test_exact(-5.0, -3.0, -4.0, 0.5f64.ln(), ln_pdf(-4.5));
         test_exact(-5.0, -3.0, -4.0, 0.0, ln_pdf(-4.0));
         test_exact(-5.0, -3.0, -4.0, 0.5f64.ln(), ln_pdf(-3.5));
+    }
+
+    #[test]
+    fn test_ln_pdf_mode_at_an_endpoint() {
+        let ln_pdf = |arg: f64| move |x: Triangular| x.ln_pdf(arg);
         test_exact(0.0, 1.0, 0.0, 2f64.ln(), ln_pdf(0.0));
         test_exact(-2.0, 5.0, -2.0, (2.0f64 / 7.0).ln(), ln_pdf(-2.0));
+        test_exact(0.0, 1.0, 1.0, 2f64.ln(), ln_pdf(1.0));
     }
 
     #[test]
--- a/src/distribution/triangular.rs
+++ b/src/distribution/triangular.rs
@@ -388,12 +388,24 @@ impl Continuous<f64, f64> for Triangular {
     ///     0
     /// }
     /// ```
+    ///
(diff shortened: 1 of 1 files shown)


src/distribution/triangular.rs
```diff
@@ -388,12 +388,24 @@ impl Continuous<f64, f64> for Triangular {
     ///     0
     /// }
     /// ```
+    ///
+    /// When `mode == min` the second branch is degenerate and reduces to `0 / 0`;
```
day01: **[suggestion]** The added doc paragraph restates the 0/0 diagnosis and the patch (`"the second branch is degenerate and reduces to 0 / 0"`), rather than documenting the special-case formula. The `# Formula` block above is now incomplete for a constructor-legal input (`mode == min`), and the prose leaks the implementation story instead of the identity Wikipedia already writes as `f(c) = 2/(b-a)`.

**Suggestion:** Drop the narrative paragraph and fold the endpoint into the formula, e.g. `if x == mode { 2 / (max - min) }` (Wikipedia’s `x = c` case) or a one-line `mode == min` arm. Same for `ln_pdf` if that formula is kept in sync.

src/distribution/triangular.rs
```diff
@@ -388,12 +388,24 @@ impl Continuous<f64, f64> for Triangular {
...
+    /// the density at that single point is the height of the triangle,
+    /// `2 / (max - min)`.
     fn pdf(&self, x: f64) -> f64 {
         let a = self.min;
         let b = self.max;
         let c = self.mode;
         if a <= x && x <= c {
-            2.0 * (x - a) / ((b - a) * (c - a))
+            if c == a {
+                // The only `x` reaching this branch is `x == min == mode`, where the
```
day01: **[suggestion]** The four-line inline comment restates what the `if c == a` arm does, why the expression below is 0/0, and that the mirrored `mode == max` path already works. That is change narration / design rationale, not a non-obvious constraint.

**Suggestion:** Delete the comment. The `c == a` guard plus `2.0 / (b - a)` is self-explanatory once the doc formula mentions the special case.

src/distribution/triangular.rs
```diff
@@ -547,6 +559,12 @@ mod tests {
         test_exact(-5.0, -3.0, -4.0, 0.5, pdf(-4.5));
         test_exact(-5.0, -3.0, -4.0, 1.0, pdf(-4.0));
         test_exact(-5.0, -3.0, -4.0, 0.5, pdf(-3.5));
+        // mode == min: the rising edge is degenerate at x == min, the density there
```
day01: **[nit]** The test comments (`"the rising edge is degenerate at x == min, the density there is the height of the triangle"`) repeat the same rationale as the production comments.

**Suggestion:** Label the cases only, e.g. `// mode == min` and `// mode == max`.

src/distribution/triangular.rs
day01: Delete the comment. The c == a guard plus 2.0 / (b - a) is self-explanatory once the doc formula mentions the special case

src/distribution/triangular.rs
day01: unnecessary comments, if you prefer present different behaviour it is for different test case there should be explanation, it shouldnt be in comment in source code imo

day01 (changes requested):
too many comments which covers readablity of code

src/distribution/triangular.rs
jaideeppyne: Removed. The doc formula above already states the `mode == min` case, so the guard stands on its own.

src/distribution/triangular.rs
jaideeppyne: Fair, moved rather than commented. The degenerate-edge assertions are now in `test_pdf_mode_at_an_endpoint` and `test_ln_pdf_mode_at_an_endpoint`, so the case is named instead of explained, and `test_pdf` goes back to what it was. I also added the `mode == max` mirror to both while moving them.

Both new tests fail on master and pass with the change.

jaideeppyne:
Point taken on the comment volume, that was overdone. Both are gone now.

The in-body comment in `pdf` is deleted, and the endpoint assertions moved into `test_pdf_mode_at_an_endpoint` and `test_ln_pdf_mode_at_an_endpoint` rather than sitting behind a comment in `test_pdf`. While moving them I added the `mode == max` mirror to both. The only prose left is the one line in the doc formula, which I kept since you referred to it as covering the special case, happy to cut that too.

The code change itself is unchanged, and both new tests fail on master.

day01 (approved):
Results correct, algo confirms with mapmath

day01:
@jaideeppyne Yeah, cut the narrative! XD

jaideeppyne:
Ha, fair. Cut that one too, the doc block is back to just the formula.

YeungOnion:
Thanks for the fix!

PR #469, opened 2026-09-06
fix: sample Gumbel via -ln(U), not ln(-U)

### LINKED ISSUE #461: Sampling a `Gumbel` returns NaN for every parameter pair
```rust
use rand::SeedableRng;
use rand::distr::Distribution;
use statrs::distribution::Gumbel;

fn main() {
    let mut rng = rand::rngs::StdRng::seed_from_u64(42);
    let g = Gumbel::new(0.0, 1.0).unwrap();
    let samples: Vec<f64> = (0..5).map(|_| g.sample(&mut rng)).collect();
    println!("{samples:?}");
}
```

Output:

```
[NaN, NaN, NaN, NaN, NaN]
```

Every draw is NaN, for any location and scale and any RNG. `Gumbel::new(0.0, 1.0)` returns `Ok`, and the distribution's `cdf`, `inverse_cdf` and moments are all finite, so only sampling is affected.

Tested on statrs 0.19.1 and current main (`rand` feature).

teddytennant:
`Gumbel::sample` is inverse transform sampling, so it should be the same map as `inverse_cdf`: `μ - β ln(-ln(U))` for `U ~ Uniform(0, 1)`. It was written as `((-x).ln()).ln()` instead of `(-x.ln()).ln()`. `x` is in `[0, 1)`, so `-x` is negative and `ln(-x)` is NaN on every draw.

```
Gumbel::new(0.0, 1.0).unwrap().sample(&mut rng)  // NaN
```

`cdf`, `inverse_cdf`, and the moments were already finite for the same parameters. The parentheses just sat on the wrong operand; `inverse_cdf` already had the right grouping.

`test_sample` is the same shape as Cauchy's: 20 draws from a seeded `StdRng` must be finite. It failed on main with `sample was NaN` and passes with the grouping fixed.

`cargo test` is green (834 lib + 195 doctests), as are `cargo fmt -- --check`, clippy under `-Dwarnings`, and `cargo check --no-default-features --lib`.

Fixes #461

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected Gumbel distribution sampling to produce valid finite values for positive random inputs.
  - Improved reliability of generated samples across repeated sampling operations.

### DIFF
--- a/src/distribution/gumbel.rs
+++ b/src/distribution/gumbel.rs
@@ -118,7 +118,7 @@ impl core::fmt::Display for Gumbel {
 impl ::rand::distr::Distribution<f64> for Gumbel {
     fn sample<R: rand::Rng + ?Sized>(&self, r: &mut R) -> f64 {
         let x = ::rand::RngExt::random::<f64>(r);
-        self.location - self.scale * ((-x).ln()).ln()
+        self.location - self.scale * (-x.ln()).ln()
     }
 }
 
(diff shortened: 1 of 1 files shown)


YeungOnion:
good call to include the test case, thanks!

